PR #670, opened 2018-06-25
print() is a function in Python 3

cclauss:
This is probably intentional but just in case...

These three files contain __print__ statements instead of __print()__ functions.

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./lab/branches.py:16:29: E999 SyntaxError: invalid syntax
        print "In while True"
                            ^
./lab/run_trace.py:17:26: E999 SyntaxError: invalid syntax
    print "%s%s %s %d @%d" % (
                         ^
./lab/show_platform.py:16:20: E999 SyntaxError: invalid syntax
    print "%30s: %r" % (n, v)
                   ^
```

### DIFF
--- a/lab/branches.py
+++ b/lab/branches.py
@@ -13,11 +13,11 @@ def my_function(x):
 
     i = 0
     while True:
-        print "In while True"
+        print("In while True")
         if i > 0:
             break
         i += 1
-    print "Left the True loop"
+    print("Left the True loop")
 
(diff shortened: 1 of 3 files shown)


nedbat:
Hmm, these print statements probably mean that I don't really need these files?  But it couldn't hurt to fix them :)

PR #672, opened 2018-06-25
Define basestring for Python 3

cclauss:


nedbat:
I'm certain there will be more changes needed to make this work under Python 3.  Are you trying to make it work completely?

cclauss:
Yes.  I was under the impression that it already did run under Python 3.
* https://github.com/nedbat/coveragepy/blob/master/setup.py#L28-L32

nedbat:
@cclauss the "lab" directory is not part of the installed package.  They are tossed-off code that I have found useful while diagnosing problems.  Let's close these until we decide whether to make them fully work or not.

cclauss:
Understood.

PR #675, opened 2018-06-28
Implementation of intern table for file_data keys

nedbat:
This is the code from https://github.com/nedbat/coveragepy-bbmirror/pull/32


nedbat:
@DRMacIver I've moved the pull request here.  We have some failing tests inside Hypothesis on Python 2.7.  Any ideas?

DRMacIver:
> We have some failing tests inside Hypothesis on Python 2.7. Any ideas?

Yeah, sorry, I need to revisit this. I'm juggling a few too many things at the moment so I haven't had a chance to make loading this back into my brain and working on it a priority. I will try to do so at some point in the next month.

My suspicion based on the (very weird) errors is that I've somehow managed to mess up module loading. Maybe an incorrect decref somewhere?

nedbat:
I'm happy to reopen this if you get back to it.

PR #676, opened 2018-06-30
annotate: annotate missing branches with "~"

blueyed:


nedbat:
@blueyed Thanks for persisting with this.  I've never valued the annotation feature.  Do you use it?

tests/farm/annotate/gold_branch/white.py,cover
```diff
@@ -0,0 +1,36 @@
+  # Licensed under the Apache License: http://www.apache.org/licenses/LICENSE-2.0
+  # For details: https://bitbucket.org/ned/coveragepy/src/default/NOTICE.txt
```
nedbat, line 2: You'll need to correct the NOTICE links to GitHub to get the tests to pass.

coverage/annotate.py
```diff
@@ -60,6 +60,7 @@ def annotate_file(self, fr, analysis):
         statements = sorted(analysis.statements)
         missing = sorted(analysis.missing)
         excluded = sorted(analysis.excluded)
+        missing_branches = sorted(analysis.missing_branch_arcs().keys())
```
nedbat, line 63: There's no reason to sort this, since we're just using `in` on it.  Making it a set would be better.  Same for `excluded`, though that was my fault! :)   `statements` and `missing` need to be sorted because of how the algorithm walks them in tandem.

blueyed:
> Do you use it?

Yes, but mostly for debugging/developing.

I've just created https://github.com/nedbat/coveragepy/pull/677 - where this is included already (without the tests).

ProsperousHeart:
> > Do you use it?
> 
> Yes, but mostly for debugging/developing.
> 
> I've just created #677 - where this is included already (without the tests).

@nedbat I think this can be closed?

PR #690, opened 2018-08-08
Persist html sort order in localStorage instead of cookie

federicobond:
This allows the sorting to work in localhost, where cookies are not allowed.

Both JSON.parse and localStorage are widely supported in all major browsers.


nedbat:
Thanks, looks great!

PR #710, opened 2018-09-17
CoverageSqliteData.__nonzero__: do not create DB

blueyed:
This makes is more lazy and avoids creating an empty DB unnecessarily.


nedbat:
I can see how this is better in general, but I'm wondering if there's a real-world scenario that drew your attention to it?

blueyed:
Noticed it when porting covimerage for coverage 5, where I had a simple `if` checking for this to be set already (https://github.com/Vimjas/covimerage/pull/54/files#diff-70032961e5469edf5cc5ce54027c52d9L22).

PR #719, opened 2018-10-08
Update version.py to follow semver rules

nbelakovski:
In its current state _make_version would generate version strings that were not compatiable with semver rules. This was causing issues in that pre-release versions of the code were being downloaded.

The code has been updated to always include minor version and to use a hyphen to separate pre-release versions.

### DIFF
--- a/coverage/version.py
+++ b/coverage/version.py
@@ -11,12 +11,10 @@ version_info = (5, 0, 0, 'alpha', 4)
 def _make_version(major, minor, micro, releaselevel, serial):
     """Create a readable version string from version_info tuple components."""
     assert releaselevel in ['alpha', 'beta', 'candidate', 'final']
-    version = "%d.%d" % (major, minor)
-    if micro:
-        version += ".%d" % (micro,)
+    version = "%d.%d.%d" % (major, minor, micro)
     if releaselevel != 'final':
         short = {'alpha': 'a', 'beta': 'b', 'candidate': 'rc'}[releaselevel]
-        version += "%s%d" % (short, serial)
+        version += "-%s%d" % (short, serial)
     return version
 
 


nbelakovski:
The following commands all generate an error:
`python -c "import semver as s; print(s.parse_version_info('5.0a3'))"`
`python -c "import semver as s; print(s.parse_version_info('5.0.0a3'))"`
`python -c "import semver as s; print(s.parse_version_info('5.0-a3'))"`
This one works:
`python -c "import semver as s; print(s.parse_version_info('5.0.0-a3'))"`
For more detail:
`python -c "import semver as s; print(s.parse_version_info('5.0.0-a3').__repr__())"`

nedbat:
I'm curious where are seeing incorrect downloads.  I'm using the same conventions as Python itself, specified in PEP 440, and understood by pip.

nbelakovski:
I'm seeing it with pip-compile. I'm able to reproduce it if the requirements.in file contains the text "coverage>=4.0a6". As it turns out one of my internal dependencies has this exact line within its requirements file. This seems more like a bug in pip-compile, it shouldn't be going to pre-release versions if the --pre flag is not specified. I'll file an issue with pip-compile. Would you like to close this, since apparently the hyphen breaks compatibility with PEP 440?

nbelakovski:
pip-tools issue: https://github.com/jazzband/pip-tools/issues/691

nedbat:
Hmm, I had thought that PEP440 was flexible enough to allow semver-compliant numbers, but now I see that it isn't.  This isn't the first time someone pointed out I don't follow semver.

Thanks for writing the bug against pip-tools.  It does seem very strange.  Even stranger is that someone would pin something at "4.0a6"

Issue #702, opened 2018-09-08
SQLite Crash on Coverage 5.0a2 inside Hypothesis


TylerADavis:
I was running some unit tests using hypothesis ([link](https://github.com/HypothesisWorks/hypothesis)) which relies on coverage, and I ran into a crash. Specifically, I received the following error message:
```coverage.misc.CoverageException: Couldn't use data file '/Users/me/my_project/.coverage': UNIQUE constraint failed: file.path```

I have experienced this issue both on macOS and Linux while running python 3.6, and the issue does not persist with `COVERAGE_STORAGE=json`.

I have included a gist below which contains files to reproduce the error. Specifically, the error occurs when the hypothesis `assume` statement in `test.py` fails more than half the time and when `wrap` is called. 

To reproduce the error, run the `build.sh` script, which will initialize a virtualenv, install required modules, and run a short example which should throw the error.

Additionally, I am including a vagrantfile which will run `build.sh`in a VM, thus providing a clean slate for testing.

https://gist.github.com/TylerADavis/393ee6382b4f645774100ae0c490c5ba

Please let me know if there is any other information I could provide that would be useful.

nedbat:
Thanks, this reproduces well, even without Vagrant in the mix.

nedbat:
It looks like Hypothesis is creating a number of CoverageData objects, each of which is trying to write to the .coverage file.  @DRMacIver thoughts?

DRMacIver:
> It looks like Hypothesis is creating a number of CoverageData objects, each of which is trying to write to the .coverage file. @DRMacIver thoughts?

Hmm. I *thought* the way Hypothesis was using those CoverageData objects they weren't trying to write to the file at all. I don't know whether my belief there was wrong or whether this has changed as part of the new version of Coverage.

I'm currently arguing that we should pull the Coverage support in Hypothesis anyway (https://github.com/HypothesisWorks/hypothesis/issues/1551). This is not due to problems with coverage-the-library, I just don't think coverage information in general has proven useful in the sort of testing we can feasibly do in normal runs.

Given that, my recommended workaround is to put `use_coverage=False` in your Hypothesis settings, as I don't think you're losing much by doing that. I'll try to make sure we fix this (probably by removing the feature altogether) before you ship the final version of 5.0.

DRMacIver:
https://github.com/HypothesisWorks/hypothesis/pull/1564 will fix this.

nedbat:
[...]

PR #723, opened 2018-10-23
CoverageSqliteData._file_id: use "insert or replace"

blueyed:
Fixes https://github.com/nedbat/coveragepy/issues/702

### DIFF
--- a/coverage/sqldata.py
+++ b/coverage/sqldata.py
@@ -181,13 +181,13 @@ class CoverageSqliteData(SimpleReprMixin):
     def _file_id(self, filename, add=False):
         """Get the file id for `filename`.
 
-        If filename is not in the database yet, add if it `add` is True.
+        If filename is not in the database yet, add it if `add` is True.
         If `add` is not True, return None.
         """
         if filename not in self._file_map:
             if add:
                 with self._connect() as con:
-                    cur = con.execute("insert into file (path) values (?)", (filename,))
+                    cur = con.execute("insert or replace into file (path) values (?)", (filename,))
                     self._file_map[filename] = cur.lastrowid
         return self._file_map.get(filename)
 


coverage/sqldata.py
```diff
@@ -182,8 +182,16 @@ def _file_id(self, filename, add=False):
         if filename not in self._file_map:
             if add:
                 with self._connect() as con:
-                    cur = con.execute("insert into file (path) values (?)", (filename,))
-                    self._file_map[filename] = cur.lastrowid
+                    try:
+                        cur = con.execute("insert into file (path) values (?)", (filename,))
```
anarcat: how about skipping the extra query and handling the duplication in SQL?

```suggestion
                        cur = con.execute("insert OR REPLACE into file (path) values (?)", (filename,))
```

i'm not sure how / if this will change the row id however, but it feels more elegant than handling the exception which is a little brittle: what if the error message changes? what if the column name changes?

anarcat (commented):
LGTM although I would favor a SQL-only simpler approach, as described.

coverage/sqldata.py
blueyed: Makes sense.
Can adjust if @nedbat agrees / wants to handle it this way after all.

blueyed: Changed to "insert or replace", thanks!

coverage/sqldata.py
```diff
@@ -181,13 +181,13 @@ def dump(self):                                         # pragma: debugging
...
 
-        If filename is not in the database yet, add if it `add` is True.
+        If filename is not in the database yet, add it if `add` is True.
         If `add` is not True, return None.
         """
         if filename not in self._file_map:
             if add:
                 with self._connect() as con:
-                    cur = con.execute("insert into file (path) values (?)", (filename,))
+                    cur = con.execute("insert or replace into file (path) values (?)", (filename,))
```
nedbat, line 190: How does this fix relate to #760?

blueyed: You've made the link.. I guess it can help, but can only say it for https://github.com/nedbat/coveragepy/issues/702#issuecomment-432220508 really, although I've not tested the "insert or replace" attempt with that one.

PR #724, opened 2018-10-27
Remove use_2to3=False; it is the default

jdufresne:
Slightly simplifies setup.py.


nedbat:
Thanks for cleaning up the old stuff I don't even see any more! :)

PR #746, opened 2018-12-24
Disallow impossible values for fail_under

### LINKED ISSUE #743: Giving it your all, and then some?
I really like this tool - it's saved my butt a fair amount, and has helped me write better code. Thank you for all that you do!

I recently wrote a quick tool, and when I finished writing all the tests and code, achieving that glorious 100% test coverage. ✨ 🎉 

I wanted to ensure that the quality bar is now set high, so I figured I'd use the nifty `--fail-under` argument and set it to 100 - that works as I expected, yay!

Then I figured, I'm a bit of a smart-ass, what happens if I set the bar to something that is impossible, above 100%?

```shellsession
$ coverage report --fail-under=101 ; echo $?
Name            Stmts   Miss  Cover
-----------------------------------
myproject.py       39      0   100%
2
```

The comparison logic works fine - my setting is indeed higher than the reported coverage, but something deep inside of me asks - is that even possible?

I'm guessing that this is a case where my expectation would be that the tool tells me that I've asked for something impossible - to prevent me from accidentally committing a change that will then trip up the process. Is that reasonable?

miketheman:
Since there's no way were likely to achieve greater than 100% code coverage,
disallow usage of any value above 100.

Resolves #743

Signed-off-by: Mike Fiedler <miketheman@gmail.com>

### DIFF
--- a/tests/test_results.py
+++ b/tests/test_results.py
@@ -105,3 +105,8 @@ class NumbersTest(CoverageTest):
 ])
 def test_should_fail_under(total, fail_under, precision, result):
     assert should_fail_under(float(total), float(fail_under), precision) == result
+
+
+def test_should_fail_under_invalid_value():
+    with pytest.raises(ValueError):
+        should_fail_under(100.0, 101, 0)
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -81,6 +81,7 @@ Matthew Desmarais
 Max Linke
 Michał Bultrowicz
 Mickie Betz
+Mike Fiedler
 Nathan Land
 Noel O'Boyle
 Olivier Grisel
--- a/coverage/results.py
+++ b/coverage/results.py
@@ -283,6 +283,10 @@ def should_fail_under(total, fail_under, precision):
     Returns True if the total should fail.
 
     """
+    # We can never achieve higher than 100% coverage
+    if fail_under > 100.0:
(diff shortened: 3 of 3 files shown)


miketheman:
I tried to find a place to implement this other than a method that currently only returns a bool - but I didn't find anywhere that configs are validated, so I hope this is ok.

Very open to feedback here!

nedbat:
Thanks! I took a quick look earlier, and I think I would have put the check in the same place you did.

I wouldn't make the message so flip :)  Feel free to add yourself to the CONTRIBUTORS file. :)

miketheman:
@nedbat I updated the error message to be a little less... opinionated. 😀 

Codecov failure appears once I force-pushed the branch, it should probably re-report at some later time?

tests/test_results.py
```diff
@@ -105,3 +105,8 @@ def test_covered_ratio(self):
 ])
 def test_should_fail_under(total, fail_under, precision, result):
     assert should_fail_under(float(total), float(fail_under), precision) == result
+
+
+def test_should_fail_under_invalid_value():
+    with pytest.raises(ValueError) as excinfo:
```
nedbat: One more tweak: no need for excinfo here (pylint will complain)

miketheman: Removed!

nedbat:
Thanks!

Issue #748, opened 2019-01-04
Coverage html report lineno font-size too small


whusnoopy:
The CSS defined in html report styles.css set line number font-size to 10px, but Chrome have a minimal font-size setting which with 12px as default, so the font-size for lineno will render as 12px, and the line-height will show as 1.6x12 = 19.2, it cause the lineno tall than code, code part is still with font-size 12px (0.75em) and line-height 16px (1.333333em)

https://github.com/nedbat/coveragepy/blob/7a099a2204e6863824f11432c807442f3ad5306c/coverage/htmlfiles/style.css#L208

------

We can suggest users to change their Chrome minimal font-size to 10px or smaller to ignore this problem, but for others doesn't know the reason, maybe we can change the font-size of lineno to 12px for a better way?

nedbat:
Thanks, I had never seen the minimum font size setting before.  The default for me (and a number of people I asked) is 6 point, do you know why it was 12 for you?

I've [fixed this a slightly different way](https://github.com/nedbat/coveragepy/commit/0694ee9cd33a75e28862234eca1ac3005f54882b) than your pull request.  Your way gave everyone 12-point line numbers.  My way leaves the default as 10, but keeps the alignment correct if the size is actually 12 due to the minimum font setting.

whusnoopy:
Thanks for the better way!

As https://bugs.chromium.org/p/chromium/issues/detail?id=36429 said, Chrome set minimal font-size to 10/12px for Simp/Trad Chinese, Japanese, Korean and Thai, and my locate is ZH-cn (Simp Chinese), that's why I got the problem.

PR #749, opened 2019-01-04
a better way to show lineno in html report

whusnoopy:
try to solve the issue https://github.com/nedbat/coveragepy/issues/748


nedbat:
I've [fixed this a different way](https://github.com/nedbat/coveragepy/commit/0694ee9cd33a75e28862234eca1ac3005f54882b), let me know if it doesn't work for you, and we can keep trying.  Thanks!

Issue #761, opened 2019-01-25
Combing SQL Coverage Data unacceptably slow


strichter:
**Describe the bug**
Generating a combined coverage report is about *50-60 times* slower with SQL-based coverage data than with JSON.

**To Reproduce**
coverage master with sql data coverage. You need a sizable set of coverage data.

**Additional context**
We are combining about 100-200 coverage files into one large one. The resulting SQL coverage file is about 390MB. It used to be able to combine those files in about 1 minute. Now it takes somewhere between 50-60 minutes.

The reason for the inefficiency is obvious: It combines files through the regular coverage measurement recording APIs, which are inappropriate to use here, since they cause one transaction per arc. SQL should be used in proper ways with do bulk reading and inserts.

strichter:
This issue should really have label:5.0-pre-release

strichter:
Okay, some more information. It turns out that the new context-based coverage is about 50-60 times the data compared to the context-free coverage. Our arc table now has 9.5M entries, but there are only 172k unique `(file_id, fromno, tono)` entries.

We have been working all morning on optimizing combining and have come up with the following script that runs in 2:30 minutes, so it is somewhere between 20-30 times faster then the current master solution.

```
import os
import sys
import time
from coverage import sqldata


def main(args=sys.argv[1:]):
    combined_fn = args[0]
    part_fns = args[1:]

    arcs = []
    lines = []

    t0 = t1 = time.time()
    for path in part_fns:
        data = sqldata.CoverageSqliteData(path)
        data.read()
        with data._connect() as conn:
            cur = conn.execute(
                    'SELECT file.path, context.context, arc.fromno, arc.tono '
                    'FROM arc '
                    'INNER JOIN file on file.id = arc.file_id '
                    'INNER JOIN context on context.id = arc.context_id ')
            arcs.extend(cur)
            cur.close()

            cur = conn.execute(
                    'SELECT file.path, context.context, line.lineno '
                    'FROM line '
                    'INNER JOIN file on file.id = line.file_id '
                    'INNER JOIN context on context.id = line.context_id ')
            lines.extend(cur)
            cur.close()

            print(f'{path}: {time.time()-t1}')
            t1 = time.time()

    print('Total Loading:', time.time()-t0)

    t2 = time.time()

    files = set()
    contexts = set()
    for row in arcs:
        files.add(row[0])
        contexts.add(row[1])

    for row in lines:
        files.add(row[0])
        contexts.add(row[1])

[...]

PR #765, opened 2019-01-26
Fast CoverageSqliteData Update/Combine

strichter:
Improve update speed by 25-30x.

### DIFF
--- a/coverage/sqldata.py
+++ b/coverage/sqldata.py
@@ -344,21 +344,16 @@ class CoverageSqliteData(SimpleReprMixin):
         self._start_using()
 
         # Collector for all arcs, lines and tracers
-        files = []
-        contexts = []
-        arcs = []
-        lines = []
-        tracers = {}
         other_data.read()
         with other_data._connect() as conn:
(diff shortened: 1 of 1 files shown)


strichter:
Resolves issue https://github.com/nedbat/coveragepy/issues/761

nedbat:
Thanks! Can you give me a quick tl;dr of this change, and why it's so much faster?

strichter:
@nedbat It actually turns out to be only a 2x speedup. This is because we optimize lookup and insert with just one query per action (read/write) and object (file, context, arc, line, tracer).

The other 12x of our slowdown come actually from this: https://github.com/nedbat/coveragepy/issues/768

strichter:
BTW, commit d9ad81f02a23072d94b1b105cc4aadf75e676d9b resulted in a 2x speedup on our AWS Jenkins test servers.

With all perf PRs combined, 50x of data (due to contexts) is now only 5x slower to combine (in our case from 40s -> ~185s). Good enough to call it a day. :-)

PR #767, opened 2019-01-27
Fix canonical filename cache.

strichter:
On our systems with a large `sys.path` this improves performance by 25x.

(Note: It is always a bad idea to reassign the cachekey during the computation.)

### DIFF
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -59,6 +59,7 @@ def canonical_filename(filename):
 
     """
     if filename not in CANONICAL_FILENAME_CACHE:
+        cf = filename
         if not os.path.isabs(filename):
             for path in [os.curdir] + sys.path:
                 if path is None:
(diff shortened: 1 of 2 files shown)


strichter:
Fixes issue https://github.com/nedbat/coveragepy/issues/768

strichter:
I am not sure why appveoy complains, the tests work for me locally on Py2.7.

nedbat:
Thanks, dumb mistake! :)  The tests are failing because of case-sensitivity on Windows.  TBH, I'm not sure why FilesTest.abs_path isn't just using files.py:abs_file...

nedbat:
This will fix the test:
```
diff --git a/tests/test_files.py b/tests/test_files.py
index 87b85431..e271da1b 100644
--- a/tests/test_files.py
+++ b/tests/test_files.py
@@ -59,7 +59,7 @@ def test_filepath_contains_absolute_prefix_twice(self):

     def test_canonical_filename_ensure_cache_hit(self):
         self.make_file("sub/proj1/file1.py")
-        d = os.path.normpath("sub/proj1")
+        d = actual_path(self.abs_path("sub/proj1"))
         self.chdir(d)
         files.set_relative_directory()
         canonical_path = files.canonical_filename('sub/proj1/file1.py')
```

nedbat:
I'll fix the test post-merge.

strichter:
@nedbat thanks for merging.

PR #769, opened 2019-01-27
Distinct list values for  `CoverageSqliteData.arcs()` and `CoverageSqliteData.lines()`

strichter:
Now the return values properly reflect the data from the JSON data implementation.

(It is also a huge performance improvement!)


nedbat:
Thanks!

PR #776, opened 2019-02-21
ci: Travis: use dist=xenial

blueyed:


nedbat:
Can you explain why?

blueyed:
It is more modern (Ubuntu Xenial instead of Trusty), and will allow to easily add py37 there - which I wanted to in the next commit.

PR #782, opened 2019-03-07
Expose switch_context in coverage API

sjustas:
This PR adds `Coverage.switch_context` method.

I think this is a quite efficient way to expose dynamic context to test runners that already use coverage API.

For example, `nose` test runner:  https://nose.readthedocs.io/en/latest/plugins/cover.html
The runner already knows the test case name, so there is no need to check each frame dymamically if it entered a test function.   In addition, this would allow runners to report labels consistently with their test case name format.

Excerpt from `nose` coverage support:
```python
class Coverage(Plugin):
<...>

    def configure(self, options, conf):
        <...>
        if self.enabled:
            self.status['active'] = True
            self.coverInstance = coverage.coverage(auto_data=False,
                branch=self.coverBranches, data_suffix=conf.worker,
                source=self.coverPackages, config_file=self.coverConfigFile)
            self.coverInstance._warn_no_data = False
            self.coverInstance.is_worker = conf.worker
            self.coverInstance.exclude('#pragma[: ]+[nN][oO] [cC][oO][vV][eE][rR]')
        <...>

    def beforeTest(self, *args, **kwargs):
        """
        Begin recording coverage information.
        """

        if self.coverInstance.is_worker:
            self.coverInstance.load()
            self.coverInstance.start()
```

What do you think about this approach?

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -477,6 +477,20 @@ class Coverage(object):
         self._data.erase(parallel=self.config.parallel)
         self._data = None
 
+    def switch_context(self, new_context):
+        """Switch to a new dynamic context.
+
+        `new_context` is a string to use as the context label
+        for collected data.
+
+        Coverage collection must be started already.
+        """
+        if not self._started:
+            raise CoverageException(                    # pragma: only jython
+                "Cannot switch context, coverage is not started"
+                )
+        self._collector.switch_context(new_context)
+
     def clear_exclude(self, which='exclude'):
         """Clear the exclude list."""
         self._init()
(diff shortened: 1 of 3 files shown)


nedbat:
I love this approach, and had always been intending to connect the test runners so that they could tell us the context.  Not only do they know when the tests start, but with parametrized tests, they know things we can't know (the data used for the test).

I'd want to see a real pytest plugin that uses this before I felt like it really worked.

Also, we'll now have three different ways that context can be started (built-in test_function, context plugins, and test runners calling this method).  I can't quite wrap my head around how they will all intereract, or even quite how to explain it! :)

nedbat:
Thanks, this works well, and as usual, thanks especially for the attention to tests.

PR #783, opened 2019-03-07
Plugin support for dynamic context

sjustas:
This PR allows writing simple dynamic context plugins.

```python
class TestMethodPlugin(coverage.CoveragePlugin):

    def dynamic_context(self, frame):
        if frame.f_code.co_name.startswith("test_"):
            return coverage.context.qualname_from_frame(frame)
        return None

def coverage_init(reg, options):
    reg.add_dynamic_context(TestMethodPlugin())
```

I kept the `dynamic_context = test_function` setting for backwards compatibility.

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -335,6 +335,34 @@ class Coverage(object):
         if not should_skip:
             self._data.read()
 
+    def _combine_context_switchers(self, context_switchers):
+        """Create a single context switcher from multiple switchers.
+
+        `context_switchers` is a list of methods that take a frame
+        as an argument and return a string to use as the new context label.
+
+        Returns a method that composits `context_switchers` methods, or None
+        if `context_switchers` is an empty list.
+
+        When invoked, the combined switcher calls `context_switchers` one-by-one
+        until a string is returned.  Combined switcher returns None if all
+        `context_switchers` return None.
+        """
+        if not context_switchers:
+            return None
+
+        if len(context_switchers) == 1:
+            return context_switchers[0]
+
+        def should_start_context(frame):
+            for switcher in context_switchers:
+                new_context = switcher(frame)
+                if new_context is not None:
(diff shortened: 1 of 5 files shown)


nedbat:
@sjustas Thanks for doing this.  I've been meaning to make time to look at it, but I have a conference coming up.  It may take me a few more weeks, sorry.

sjustas:
@nedbat, no worries, I'll be looking forward to your feedback.

nedbat:
@sjustas This is very comprehensive, thanks.  I appreciate the work you put into the tests, which are significant.

Do you have a use case for chaining the context switchers?  It would be easier to just have one in effect at a time, but maybe you had a need to have more than one?

sjustas:
This could be useful for supporting multiple different TestSuite implementations.  Python is pretty flexible there, even bundled `doctest` has `DocTestFinder` which generates [DocTestCase](https://github.com/python/cpython/blob/master/Lib/doctest.py#L2148) instances from .txt files it finds (like "Readme.txt"), so someone might want write a plugin that sets a context for these tests.

Another example could be [RMLRenderingTestCase](https://github.com/zopefoundation/z3c.rml/blob/master/src/z3c/rml/tests/test_rml.py#L38).  It would probably handle contexts in similar way to `DocTestCase`, but in theory people can make all sorts of test suites and test cases.  It would be nice if the context names would be meaningful.

So I thought it would be nice if developers could implement context plugins that work correctly with each library, and users of these libraries (say unittest + doctest + z3c.rml) could just list all the plugins in configuration.

I think another good way to approach this is to have an option for test runners themselves to set correct contexts for each test case they run, hence the PR https://github.com/nedbat/coveragepy/pull/782 :slightly_smiling_face:

nedbat:
Thanks! I'll be doing some clean up after the merge.

PR #790, opened 2019-03-31
[WIP] ci: Travis: test with Python 3.8-dev

blueyed:


blueyed:
It seems like using "-X fix" with codecov makes some function definitions appear to be covered etc.

blueyed:
Those changes to "def" lines are also there without "-X fix" (https://codecov.io/gh/nedbat/coveragepy/compare/820b255f34a0aac8670b0c819153bb8b38c4b2c6...c49543ba1c80bc0ba08b0e2f4422c49b03006d8d/changes)..
Related to the bash uploader then?!

.travis.yml
```diff
@@ -1,6 +1,3 @@
-# Tell Travis what to do
-# https://travis-ci.com/nedbat/coveragepy
-
```
nedbat: Why remove the comment from the top?

.travis.yml
```diff
@@ -24,6 +22,11 @@ install:
...
-  - if [[ $COVERAGE_COVERAGE == 'yes' ]]; then python igor.py combine_html; fi
-  - if [[ $COVERAGE_COVERAGE == 'yes' ]]; then pip install codecov; fi
-  - if [[ $COVERAGE_COVERAGE == 'yes' ]]; then codecov -X gcov --file coverage.xml; fi
+
+after_script:
+  - |
+    if [[ $COVERAGE_COVERAGE == 'yes' ]]; then
+      python igor.py combine_html
+      pip install codecov
+      codecov -X gcov coveragepy search xcode fix -f coverage.xml -n $TOXENV
```
nedbat: What do these arguments do?

.travis.yml
blueyed: I found it not useful, after having read it a few times - but that's because I know what it is for.  I can leave it in.

.travis.yml
blueyed: The disable more checks, e.g. "coveragepy" is not necessary, since there is a coverage.xml already, and the others are mainly for performance.
But I suspect that "fix" might have an influence on "def"s being covered or not.

nedbat:
Is this ready to merge?

blueyed:
The coverage changes (https://codecov.io/gh/nedbat/coveragepy/compare/820b255f34a0aac8670b0c819153bb8b38c4b2c6...b1cd33194e2e51de8e4b7dde2e4cac8e5e9e9e32/changes) are suspicious, but maybe just because of adding py38 after all?!

nedbat:
TBH, I still don't understand how to read codecov.io well.  Doesn't it say the coverage is increased slightly?  In any case, the changes seem good to me.

nedbat:
(I really have to remember to use the squash merge button...)

blueyed:
> TBH, I still don't understand how to read codecov.io well. Doesn't it say the coverage is increased slightly?

Yes, and it seems to have increased because of lines that should have been covered before already.  Maybe a bug in codecov, maybe a fixed bug in py38..?

PR #791, opened 2019-03-31
[postponed] Set f_trace_lines = 0/False on ignored frames

blueyed:
This is meant for optimization, so that the trace handler is not called
for lines in blocks that are not traced.

TODO:

- [ ] needs fix in cpython (https://bugs.python.org/issue36494), or another way to reset/restore it then
- [ ] test ?

### DIFF
--- a/coverage/pytracer.py
+++ b/coverage/pytracer.py
@@ -15,6 +15,9 @@ if env.PY2:
     YIELD_VALUE = chr(YIELD_VALUE)
 
 
+HAS_F_TRACE_LINES = sys.version_info >= (3, 7)
+
+
 class PyTracer(object):
     """Python implementation of the raw data tracer."""
 
@@ -118,7 +121,8 @@ class PyTracer(object):
                     self.data[tracename] = {}
                 self.cur_file_dict = self.data[tracename]
             else:
-                frame.f_trace_lines = False
+                if HAS_F_TRACE_LINES:
+                    frame.f_trace_lines = False
             # The call event is really a "start frame" event, and happens for
             # function calls and re-entering generators.  The f_lasti field is
             # -1 for calls, and a real offset for generators.  Use <0 as the
--- a/coverage/pytracer.py
+++ b/coverage/pytracer.py
@@ -117,6 +117,8 @@ class PyTracer(object):
                 if tracename not in self.data:
                     self.data[tracename] = {}
                 self.cur_file_dict = self.data[tracename]
+            else:
+                frame.f_trace_lines = False
             # The call event is really a "start frame" event, and happens for
             # function calls and re-entering generators.  The f_lasti field is
             # -1 for calls, and a real offset for generators.  Use <0 as the
--- a/coverage/ctracer/tracer.c
+++ b/coverage/ctracer/tracer.c
@@ -535,7 +535,9 @@ CTracer_handle_call(CTracer *self, PyFrameObject *frame)
         self->pcur_entry->file_data = NULL;
         self->pcur_entry->file_tracer = Py_None;
         SHOWLOG(self->pdata_stack->depth, frame->f_lineno, filename, "skipped");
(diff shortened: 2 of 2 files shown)


blueyed:
This is not much of a gain for the CTracer, but for PyTracer.

With this patch `tox -e py37-coverage -- testing/test_asser*` in pytest's repo takes 50.5s, without it 116.7s (using `timid = 1` to use the PyTracer).

CTracer: 20.5s without this patch, 20.0s with it (but just run once, it is likely due to a function call being less of an overhead in C).

blueyed:
This appears to be py37+ only.

And pdb/bdb would need to be fixed to set it to True explicitly (https://bugs.python.org/issue36494).

blueyed:
Added to cpython in https://github.com/python/cpython/commit/5a8516701f5140c8c989c40e261a4f4e20e8af86 (bpo-31344).

There also `f_trace_opcodes` was added (which must be enabled explicitly).

nedbat:
We should be returning None for the trace function if we don't want to trace the frame, and that should be enough.  But I don't see where we're returning None.  Or is this the decade-old bug?

Yep, it is: https://bugs.python.org/issue11992  *sigh*

blueyed:
I've tried returning None also before, but stopped when noticing that several tests were failing due to it, and assumed coverage.py needs to trace everything still internally.  (I've also tried (quickly) to not add to `data_stack` etc, to match the behavior of "return" not being called anymore).

blueyed:
pypy2 failure (https://travis-ci.com/nedbat/coveragepy/jobs/189126179) might be unrelated?!

nedbat:
I tried restarting the pypy2 failure: https://travis-ci.com/nedbat/coveragepy/jobs/189126179 (oops, same URL...)

coverage/pytracer.py
```diff
@@ -15,6 +15,9 @@
     YIELD_VALUE = chr(YIELD_VALUE)
 
 
+HAS_F_TRACE_LINES = sys.version_info >= (3, 7)
+
```
nedbat, line 19: I'd rather add a constant to PYBEHAVIOR in env.py, perhaps using `hasattr`

nedbat:
The bug about pdb: it affects people trying to stop in the debugger when using coverage? I guess that could happen if you are running a test suite, and want to break on failures?

blueyed:
The problem/bug with pdb is "just" that coverage.py relies on its trace function being called, which is not the case anymore after `pbd.set_trace`, yes.
It affects coverage when a debugger is used during test runs interactively, but is also an issue when you're testing with pdb itself (as with [pdb++](https://github.com/antocuni/pdb)).

coverage/pytracer.py
blueyed: Would it require to get a frame object first?  (i.e. `sys._getframe()`) to inspect it?

nedbat: Sure, though that should be easy.  Though come to think of it, why not just do the hasattr where you try to set the attribute?

nedbat:
@blueyed Are you saying we shouldn't merge this until pdb is changed? Or just that they will want to eventually change?

coverage/pytracer.py
blueyed: My idea was to have as less overhead as possible in the tracing function itself.

blueyed:
The discussion above is not related to this PR really, which should just improve performance.  I was replying to the "bug about pdb" - which is more a conceptual problem due to only having a single trace function only.  pdb could restore any previous one (i.e. coverage.py's), instead of setting it to `None` in the end.

blueyed:
Oh, sorry - I forgot about https://github.com/nedbat/coveragepy/pull/791#issuecomment-478346880 / https://github.com/python/cpython/pull/12640.
Yes, it should not get merged yet.

cfbolz:
I haven't dug super deep into this PR, but maybe it's easier to restart this now that coverage only supports 3.7 upwards?

nedbat:
This same change was made in #1381

PR #792, opened 2019-03-31
Revisit tox.ini

blueyed:


nedbat:
You didn't say why you removed `passenv = *`.  I set environment variables that I want passed to the tests.

blueyed:
> This does not seem to be necessary, and should rather be an explicit
list otherwise.

What envs are you passing?

blueyed:
CI also appears to need some at least.

nedbat:
> What envs are you passing?

Are you going to suggest whitelisting them?  Can you tell me why you want to limit the environment variables?

blueyed:
Yes, I think it is better to whitelist them - this setting/feature is there for a reason, e.g. to not re-use a user's `$HOME` etc.

blueyed:
Not sure about https://travis-ci.com/nedbat/coveragepy/jobs/189125324 though - flaky?!

nedbat:
I really have no interest in putting up barriers without a problem being solved.  Why do this?

nedbat:
I removed the unnecessary tox definitions in 2bfca8b7a9061e4fa9569cdcae3c841be4921e1a.  Thanks for pointing them out.

blueyed:
> I really have no interest in putting up barriers without a problem being solved. Why do this?

I think it is good in general to have an isolated env for tests, especially with regard to $HOME etc, where tests otherwise might mess with accidentally.  Necessary/used environment variables should be whitelisted then instead.  Your call though.. :)

PR #798, opened 2019-04-22
Document about incremental report generation in html_report

sylee957:
I see it was in one of the old releases,
https://coverage.readthedocs.io/en/v4.5.x/changes.html#version-3-5b1-2011-06-05
but I think it is not documented well in the api docs.

We had an issue in https://github.com/sympy/sympy/pull/16694 because we had been mutating the files without getting warned.

I have used `.. warning::` directive, and here is the new doc

I think there are some other details about how the incremental reporting works. For example, I find that the `index.html` is regenerated, but other html files are not regenerated after deleting manually.
Also, is there an option to generate the full report from scratch with subsequent calls?

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -784,6 +784,10 @@ class Coverage(object):
 
         Returns a float, the total percentage covered.
 
+        .. warning::
+            HTML report is generated in incremental way. You should be
+            careful about mutating the files in the folder where the
+            HTML report is generated.
         """
         self.config.from_args(
             ignore_errors=ignore_errors, report_omit=omit, report_include=include,


nedbat:
Can you tell me more about how you were changing the HTML files? I haven't heard of this before.

Also, if you can provide more details about "I find that the index.html is regenerated, but other html files are not regenerated after deleting manually," there might be a bug to fix.

As for rebuilding from scratch, the simplest way is to delete the htmlcov directory and then run "coverage html".

sylee957:
We have been manually deleting the html report files, but not the folder totally.
I see the bug also exists for editing the files.

I have opened another PR to address the possible fixes.
https://github.com/nedbat/coveragepy/issues/799

nedbat:
Help me out though: why are you editing the HTML files?

sylee957:
Because we had to make an option to use a single folder to contain the reports, and do cleanups for the some orphaned html files in the folder if different coverage specification is run.

And I still see that the report files which have no intersection with the new reports are orphaned, instead of being cleaned up. Same applies for adding some junk files like `abc.html` in the coverage folder and seeing they are not automatically cleaned up.

Though I cannot give much statement about why the html files should be edited, or manually managed.

But should there be a feature implemented here to detect the corrupt files and amend?

sylee957:
I may provide some minimal reproducible example, if it is possible.

But I guess there would is some hint from the modified tests in #799, as it had actually tested out that html files should not be recovered after deleting.
I wonder if it was the original intent.

PR #799, opened 2019-04-22
Fix mutated html report file not recovered in subsequent run

sylee957:
Closes #798

Okay, I have found that there is a hash validation scheme.
But it is only comparing the hash of the coverage data residing in the memory, with the hash written in `status.json`

I think that the hash scheme should be changed, so that the hash of the actual file should be written in the json, and used in comparison.

This is an initial draft to fix the bug I have encountered.

### DIFF
--- a/coverage/html.py
+++ b/coverage/html.py
@@ -376,7 +376,7 @@ class HtmlStatus(object):
         usable = False
         try:
             status_file = os.path.join(directory, self.STATUS_FILE)
-            with open(status_file, "r") as fstatus:
+            with open(status_file) as fstatus:
                 status = json.load(fstatus)
         except (IOError, ValueError):
             usable = False
--- a/coverage/html.py
+++ b/coverage/html.py
@@ -195,14 +195,18 @@ class HtmlReporter(Reporter):
         source = fr.source()
 
         # Find out if the file on disk is already correct.
-        this_hash = self.file_hash(source.encode('utf-8'), fr)
-        that_hash = self.status.file_hash(rootname)
-        if this_hash == that_hash:
-            # Nothing has changed to require the file to be reported again.
-            self.files.append(self.status.index_info(rootname))
-            return
+        if os.path.isfile(html_path):
+            this_hash = ""
+            with open(html_path, "r", encoding="utf-8") as f:
+                hasher = Hasher()
+                hasher.update(f.read())
+                this_hash = hasher.hexdigest()
+            that_hash = self.status.file_hash(rootname)
 
-        self.status.set_file_hash(rootname, this_hash)
+            if this_hash == that_hash:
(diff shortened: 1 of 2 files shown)


coverage/html.py
```diff
@@ -195,14 +195,18 @@ def html_file(self, fr, analysis):
...
-            return
-
-        self.status.set_file_hash(rootname, this_hash)
+        if os.path.isfile(html_path):
+            this_hash = ""
+            with open(html_path, "rb") as f:
+                hasher = Hasher()
+                hasher.update(f.read().decode("UTF-8"))
+                this_hash = hasher.hexdigest()
+            that_hash = self.status.file_hash(rootname)
```
nedbat, line 203: Maybe I'm misunderstanding something. This seems like it will solve your use case (force a file to be re-written if the HTML file has been edited), but it loses the original goal.  This code will now skip writing the HTML file if the current HTML file is the same as when it was last written.  But the source code, or the results, may have changed, requiring the file to be re-written, and this hash won't detect that.

nedbat (changes requested):

coverage/html.py
nedbat: There are tests for that behavior, but they are flawed, so they all pass on this branch.

nedbat: I've updated the tests, they are on the nedbat/better-delta-html-tests branch if you want to try them out.

nedbat: But I should warn you: i'm not sure I'm willing to incur a noticeable slow-down to include the hash of the previously generated files.  This seems like an unlikely need, and has a simple work-around.

nedbat: Thanks for the work on this, but it isn't worth the cost. In my experiments, this adds significantly to the cost of HTML generation.

As I understand it, you are modifying all the HTML files. You should delete the directory before running the HTML report.

nedbat:
Thanks for the work on this, but it isn't worth the cost. In my experiments, this adds significantly to the cost of HTML generation.

I tested with an HTML directory with 2783 files. It took 59s to generate the report from scratch into an empty directory.  Changing one source file and generating again took 25s.  Reading and hashing all the .html files took between 6s and 33s (it's hard to get repeatable numbers, but I tried hard to use uncached disk reads).  That's too much time to add to 25s.

As I understand it, you are modifying all the HTML files. You should delete the directory before running the HTML report.

sylee957:
Okay, Thanks for the review. But I have some questions

1. What about adding this as an option for `html_report` rather than default, if speed is a concern.
2. If not, this may have to be documented as #798

sylee957:
And one other question I have is that checking whether the file existing or not would not slow down as much as checking the file mutation.
Because it would only have to check the filepath existing.

Would a less strict check more suitable?

PR #802, opened 2019-05-13
Allow user to specify the suffix as bool or string

earonesty:
Already supported in the `Coverage` object itself.


earonesty:
breaks too many command line usage things... needs to be its own option.  but ez enough to do

PR #811, opened 2019-06-11
Update CHANGES.rst

strichter:
Fix my name. :-) You might also add Albertas Agejevas to the credits, since he did a lot of work here too.


nedbat:
I should know the spelling by now! And I've added Albertas to CHANGES and CONTRIBUTORS.  Thanks both!

PR #816, opened 2019-06-24
Twisted concurrency model

exarkun:
Adds `--concurrency=twisted` which causes coverage to be tracked across child processes created with Twisted's `reactor.spawnProcess`.

I'm not entirely happy with the sitecustomize management code.  Also it seems like this could probably easily be a `--concurrency=exec` feature instead, with a the sitecustomize being written and PYTHONPATH updated early.  Then all processes forked & exec'd will get the logic.  It's less clear when to do cleanup but it's probably still possible.

Also this approach comes with the downside that it obscures any existing sitecustomize.  I don't really know what people use sitecustomize for so I'm not sure how big of a deal this is.

Any other solution for fork&exec seems like a challenge because there is no general way to inject some code into a process you've just exec'd.  exec is inherently about giving up control of what's running to some other program.  There might be some platform-specific tricks but that doesn't seem like an appealing avenue.

While working on this, https://pypi.org/project/coverage_enable_subprocess/ was also pointed out to me and it seems to accomplish roughly the same thing.  It uses the pth file approach instead but this means it requires a virtualenv (or for you to splat junk into your probably-OS-managed Python install which is not a great idea).  It also collects coverage from all Python programs instead of all programs in a tree under the one you start but I don't know if that's much of a practical difference.

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -37,6 +37,7 @@ class Opts(object):
     )
     CONCURRENCY_CHOICES = [
         "thread", "gevent", "greenlet", "eventlet", "multiprocessing",
+        "twisted",
     ]
     concurrency = optparse.make_option(
         '', '--concurrency', action='store', metavar="LIB",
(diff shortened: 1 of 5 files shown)


coverage/cmdline.py
```diff
@@ -635,16 +636,16 @@ def do_run(self, options, args):
...
-            # multiprocessing.
+            # multiprocessing or Twisted.
             for opt_name in ['branch', 'include', 'omit', 'pylib', 'source', 'timid']:
                 # As it happens, all of these options have no default, meaning
                 # they will be None if they have not been specified.
                 if getattr(options, opt_name) is not None:
                     show_help(
-                        "Options affecting multiprocessing must only be specified "
-                        "in a configuration file.\n"
+                        "Options affecting multiprocessing and twisted must "
```
Julian, line 647: (Will probably not be very useful reviewing much of this, but it'd be a bit nicer UX-wise if rather than saying "A or B" here you used `.format` and inserted in `options.concurrency`, i.e. the actual one of the two of those that the user really did use, instead of including ones they didn't)

ProsperousHeart:
Is this still valid?

PR #821, opened 2019-07-10
Isolate os in the db module

Bachmann1234:
I ran into this issue playing around with coverage 5 and my diff_cover project 

I fixed the issue in diff cover (https://github.com/Bachmann1234/diff_cover/pull/118) but it would have been a very difficult issue to diagnose had I not already been working from the coverage source. 

I noticed this isolate_module pattern in other places and it seemed to fix this problem as well. I tested it by running a report on a version of diff-cover without my fix


Bachmann1234:
the steps to trigger the bug btw

checkout diff_cover at commit 242a4de016c1c21ac5f25ab5a00df60a896a7a46
checkout coveragepy at commit 9bc6b93805a5f20a87211a315d00503eddab66dc

on my mac 10.14.5
```
git clone git@github.com:Bachmann1234/diff_cover.git
git clone git@github.com:nedbat/coveragepy.git
virtualenv -p python3 venv
source venv/bin/activate
cd diff_cover
git checkout 242a4de016c1c21ac5f25ab5a00df60a896a7a46
pip install -r requirements/test-requirements.txt
pip uninstall coverage
cd ../coveragepy
pip install -e .
cd ../diff_cover
coverage run -m pytest
```

Test suite should run and end with

```
Data file '/Users/bachmann/Desktop/diff_cover/.coverage' doesn't seem to be a coverage data file: Couldn't use data file '/Users/bachmann/Desktop/diff_cover/.coverage': no such table: coverage_schema
```

This is because of a sloppy mock on my part causing coverage to think the db already exist. It then fails to open said db.

Issue #720, opened 2018-10-09
Add a structured "report", probably JSON


nedbat:
There are lots of use cases that want detailed data from coverage.  The only way they have to get it today is from the XML report, which is limited by being a Java-native format.

A new report which included all of the information in a convenient machine-readable form would enable more ater-tools to be built for coverage.  @atodorov started work on a JSON report here: https://bitbucket.org/ned/coveragepy/pull-requests/61/add-json-command/diff

atodorov:
@nedbat it's been a while but can you summarize the status of the code on the BitBucket pull request. 

If there are only a few things left to do (like resolving conflicts and minor issues) I may be able to pick it up where I left it and send a PR here on GitHub.

nedbat:
@atodorov to be honest, I haven't looked at it closely in a while, but thanks.  If you move it here, we can start reviewing it.

Bachmann1234:
I was curious and it looked like this issue was stale so I started futsing with this by branching off the 4x line.  https://github.com/Bachmann1234/coveragepy/tree/json_report_4

my POC basically duplicated the XML report and replaced al the XML talk with dict calls and called `json.dumps`

The result report looks like https://gist.github.com/Bachmann1234/6347d0db7437e71323a2c391e3031ac0 (assuming I did it right. Its a spike right now and would need more tests/verification)


I think doing this the right way though would include a method that would not include so much duplication. Perhaps creating a dictionary then having the json report dump it and for the XML report simply use that dictionary to drive the rendering.

Though ultimately it comes down to the desires for the json report I guess? Is keeping it close to the cobertura xml appropriate? 🤷‍♂ 

_anyways_ ill keep playing with this

nedbat:
@Bachmann1234 thanks, I think people would appreciate a supported structured report format.  It's *probably* ok to be working from the 4.x line, since the code you are looking at hasn't changed much.  I think you are right that a JSON report and the XML report might share a data gathering phase, and then write the data differently.

I recently did something similar to the HTML report in 5.x, which made it easier to test.

BUT: I don't want to follow the Cobertura schema: they focus on "classes", which isn't right for Python.  Let's get a data format that is right for coverage.py, and then assess how well a common data gathering phase would work.  Don't for

[...]

PR #825, opened 2019-07-15
Json report

Bachmann1234:
The goal of this pr is to eventually lead to closing  https://github.com/nedbat/coveragepy/issues/720

Some example reports from a non trivial project
without Contexts: https://gist.github.com/Bachmann1234/418311192ebff1faca1e3fc54453602c
with Contexts: https://gist.github.com/Bachmann1234/bb799bdc5978a81a1e26daeb1881b950

command help:

```
venv ❯ coverage --help
Coverage.py, version 5.0a7 with C extension
Measure, collect, and report on code coverage in Python programs.

usage: coverage <command> [options] [args]

Commands:
    annotate    Annotate source files with execution information.
    combine     Combine a number of data files.
    erase       Erase previously collected coverage data.
    help        Get help on using coverage.py.
    html        Create an HTML report.
    json        Create a JSON report of coverage results.
    report      Report coverage stats on modules.
    run         Run a Python program and measure code execution.
    xml         Create an XML report of coverage results.

Use "coverage help <command>" for detailed help on any command.
Full documentation is at https://coverage.readthedocs.io/en/coverage-5.0a7
```

```
venv ❯ coverage json --help
Usage: coverage json [options] [modules]

Generate a JSON report of coverage results.

Options:
  --fail-under=MIN      Exit with a status of 2 if the total coverage is less
                        than MIN.
  -i, --ignore-errors   Ignore errors while reading source files.
  --include=PAT1,PAT2,...
                        Include only files whose paths match one of these
                        patterns. Accepts shell-style wildcards, which must be
                        quoted.
  --omit=PAT1,PAT2,...  Omit files whose paths match one of these patterns.
                        Accepts shell-style wildcards, which must be quoted.
  -o OUTFILE            Write the JSON report to this file. Defaults to
                        'coverage.json'
  --pretty-print        Print the json formatted for human readers
  --show-contexts       Show contexts for covered lines.
  --contexts=PAT1,PAT2,...
                        Only display data from lines covered in the given
                        contexts. Accepts shell-style wildcards, which must be
                        quoted.
  --debug=OPTS          Debug options, separated by commas. [env:
                        COVERAGE_DEBUG]
  -h, --help            Get help on this command.
  --rcfile=RCFILE       Specify configuration file. By default '.coveragerc',
                        'setup.cfg' and 'tox.ini' are tried. [env:
                        COVERAGE_RCFILE]

Full documentation is at https://coverage.readthedocs.io/en/coverage-5.0a7
```

simple report:
```
{
    "meta": {
        "version": "5.0a7",
        "timestamp": "2019-08-18T13:50:07.353011",
        "branch_coverage": true,
        "show_contexts": false
    },
    "files": {
        "a.py": {
            "executed_lines": [
                1,
                2
            ],
            "summary": {
                "covered_lines": 2,
                "num_statements": 3,
                "percent_covered": 60.0,
                "missing_lines": 1,
                "excluded_lines": 0,
                "num_branches": 2,
                "num_partial_branches": 1
            },
            "missing_lines": [
                3
            ],
            "excluded_lines": []
        }
    },
    "totals": {
        "covered_lines": 2,
        "num_statements": 3,
        "percent_covered": 60.0,
        "missing_lines": 1,
        "excluded_lines": 0,
        "num_branches": 2,
        "num_partial_branches": 1
    }
}
```

TODO: 
-     Iterate on the schema
-     document the report in the formal docs

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -118,6 +118,15 @@ class Opts(object):
         metavar="OUTFILE",
         help="Write the XML report to this file. Defaults to 'coverage.xml'",
     )
+    output_json = optparse.make_option(
+        '-o', '', action='store', dest="outfile",
+        metavar="OUTFILE",
+        help="Write the JSON report to this file. Defaults to 'coverage.json'",
+    )
+    json_pretty_print = optparse.make_option(
+        '', '--pretty-print', action='store_true',
+        help="Print the json formatted for human readers",
+    )
     parallel_mode = optparse.make_option(
         '-p', '--parallel-mode', action='store_true',
         help=(
@@ -402,6 +411,22 @@ CMDS = {
         usage="[options] [modules]",
         description="Generate an XML report of coverage results."
     ),
+
+    'json': CmdOptionParser(
+        "json",
+        [
+            Opts.fail_under,
+            Opts.ignore_errors,
+            Opts.include,
+            Opts.omit,
+            Opts.output_json,
+            Opts.json_pretty_print,
+            Opts.show_contexts,
+            Opts.contexts,
+            ] + GLOBAL_ARGS,
+        usage="[options] [modules]",
+        description="Generate a JSON report of coverage results."
+    ),
 }
 
 
@@ -565,6 +590,14 @@ class CoverageScript(object):
         elif options.action == "xml":
             outfile = options.outfile
             total = self.coverage.xml_report(outfile=outfile, **report_args)
+        elif options.action == "json":
+            outfile = options.outfile
+            total = self.coverage.json_report(
+                outfile=outfile,
+                pretty_print=options.pretty_print,
+                show_contexts=options.show_contexts,
+                **report_args
+            )
 
         if total is not None:
             # Apply the command line fail-under options, and then use the config
@@ -752,6 +785,7 @@ HELP_TOPICS = {
             erase       Erase previously collected coverage data.
             help        Get help on using coverage.py.
             html        Create an HTML report.
+            json        Create a JSON report of coverage results.
             report      Report coverage stats on modules.
             run         Run a Python program and measure code execution.
             xml         Create an XML report of coverage results.
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -22,11 +22,13 @@ from coverage.disposition import disposition_debug_msg
 from coverage.files import PathAliases, set_relative_directory, abs_file
 from coverage.html import HtmlReporter
 from coverage.inorout import InOrOut
+from coverage.jsonreport import JsonReporter
 from coverage.misc import CoverageException, bool_or_none, join_regex
-from coverage.misc import ensure_dir_for_file, file_be_gone, isolate_module
+from coverage.misc import ensure_dir_for_file, isolate_module
 from coverage.plugin import FileReporter
 from coverage.plugin_support import Plugins
 from coverage.python import PythonFileReporter
+from coverage.report import render_report
 from coverage.results import Analysis, Numbers
 from coverage.summary import SummaryReporter
 from coverage.xmlreport import XmlReporter
@@ -862,33 +864,29 @@ class Coverage(object):
             ignore_errors=ignore_errors, report_omit=omit, report_include=include,
             xml_output=outfile, report_contexts=contexts,
             )
-        file_to_close = None
-        delete_file = False
-        if self.config.xml_output:
-            if self.config.xml_output == '-':
-                outfile = sys.stdout
-            else:
-                # Ensure that the output directory is created; done here
-                # because this report pre-opens the output file.
-                # HTMLReport does this using the Report plumbing because
-                # its task is more complex, being multiple files.
-                ensure_dir_for_file(self.config.xml_output)
-                open_kwargs = {}
-                if env.PY3:
-                    open_kwargs['encoding'] = 'utf8'
-                outfile = open(self.config.xml_output, "w", **open_kwargs)
-                file_to_close = outfile
-        try:
-            reporter = XmlReporter(self)
-            return reporter.report(morfs, outfile=outfile)
-        except CoverageException:
-            delete_file = True
-            raise
-        finally:
-            if file_to_close:
-                file_to_close.close()
-                if delete_file:
-                    file_be_gone(self.config.xml_output)
+        return render_report(self.config.xml_output, XmlReporter(self), morfs)
+
+    def json_report(
+        self, morfs=None, outfile=None, ignore_errors=None,
+        omit=None, include=None, contexts=None, pretty_print=None,
+        show_contexts=None
+    ):
+        """Generate a JSON report of coverage results.
+
+        Each module in `morfs` is included in the report.  `outfile` is the
+        path to write the file to, "-" will write to stdout.
+
+        See :meth:`report` for other arguments.
+
+        Returns a float, the total percentage covered.
+
+        """
+        self.config.from_args(
+            ignore_errors=ignore_errors, report_omit=omit, report_include=include,
+            json_output=outfile, report_contexts=contexts, json_pretty_print=pretty_print,
+            json_show_contexts=show_contexts
+        )
+        return render_report(self.config.json_output, JsonReporter(self), morfs)
 
     def sys_info(self):
         """Return a list of (key, value) pairs showing internal information."""
--- a/coverage/jsonreport.py
+++ b/coverage/jsonreport.py
@@ -1,0 +1,100 @@
+# coding: utf-8
+# Licensed under the Apache License: http://www.apache.org/licenses/LICENSE-2.0
+# For details: https://github.com/nedbat/coveragepy/blob/master/NOTICE.txt
+
+"""Json reporting for coverage.py"""
+import datetime
+import json
+import sys
+
+from coverage import __version__
+from coverage.report import get_analysis_to_report
+from coverage.results import Numbers
+
+
+class JsonReporter(object):
+    """A reporter for writing JSON coverage results."""
+
+    def __init__(self, coverage):
+        self.coverage = coverage
+        self.config = self.coverage.config
+        self.total = Numbers()
+        self.report_data = {}
+
+    def report(self, morfs, outfile=None):
+        """Generate a json report for `morfs`.
+
+        `morfs` is a list of modules or file names.
+
+        `outfile` is a file object to write the json to
+
+        """
+        outfile = outfile or sys.stdout
+        coverage_data = self.coverage.get_data()
+        coverage_data.set_query_contexts(self.config.report_contexts)
+        self.report_data["meta"] = {
+            "version": __version__,
+            "timestamp": datetime.datetime.now().isoformat(),
+            "branch_coverage": coverage_data.has_arcs(),
+            "show_contexts": self.config.json_show_contexts,
+        }
+
+        measured_files = {}
+        for file_reporter, analysis in get_analysis_to_report(self.coverage, morfs):
+            measured_files[file_reporter.relative_filename()] = self.report_one_file(
+                coverage_data,
+                file_reporter,
+                analysis
+            )
+
+        self.report_data["files"] = measured_files
+
+        self.report_data["totals"] = {
+            'covered_lines': self.total.n_executed,
+            'num_statements': self.total.n_statements,
+            'percent_covered': self.total.pc_covered,
+            'missing_lines': self.total.n_missing,
+            'excluded_lines': self.total.n_excluded,
+        }
+
+        if coverage_data.has_arcs():
+            self.report_data["totals"].update({
+                'num_branches': self.total.n_branches,
+                'num_partial_branches': self.total.n_partial_branches,
+            })
+
+        json.dump(
+            self.report_data,
+            outfile,
+            indent=4 if self.config.json_pretty_print else None
+        )
+
+        return self.total.n_statements and self.total.pc_covered
+
+    def report_one_file(self, coverage_data, file_reporter, analysis):
+        """Extract the relevant report data for a single file"""
+        nums = analysis.numbers
+        self.total += nums
+        summary = {
+            'covered_lines': nums.n_executed,
+            'num_statements': nums.n_statements,
+            'percent_covered': nums.pc_covered,
+            'missing_lines': nums.n_missing,
+            'excluded_lines': nums.n_excluded,
+        }
+        reported_file = {
+            'executed_lines': sorted(analysis.executed),
+            'summary': summary,
+            'missing_lines': sorted(analysis.missing),
+            'excluded_lines': sorted(analysis.excluded)
+        }
+        if self.config.json_show_contexts:
+            reported_file['contexts'] = analysis.data.contexts_by_lineno(
+                file_reporter.filename
+            )
+        if coverage_data.has_arcs():
+            reported_file['summary'].update({
+                'num_branches': nums.n_branches,
+                'num_partial_branches': nums.n_partial_branches,
+            })
+        return reported_file
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -301,3 +301,22 @@ also apply to XML output, where appropriate.
 identified as packages in the report.  Directories deeper than this depth are
 not reported as packages.  The default is that all directories are reported as
 packages.
+
+.. _config_json:
+
+[json]
+-----
+
+Values particular to json reporting.  The values in the ``[report]`` section
+also apply to JSON output, where appropriate.
+
+``json_output`` (string, default "coverage.json"): where to write the json
+report.
+
+``json_pretty_print`` (boolean, default false): controls if fields in the json
+are outputted with whitespace formatted for human consumption (True) or for
+minimum file size (False)
+
+``json_show_contexts`` (boolean, default false): should the json report include
+an indication on each line of which contexts executed the line.
+See :ref:`dynamic_contexts` for details.
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -46,6 +46,10 @@ class BaseCmdLineTest(CoverageTest):
         ignore_errors=None, include=None, omit=None, morfs=[], outfile=None,
         contexts=None,
     )
+    _defaults.Coverage().json_report(
+        ignore_errors=None, include=None, omit=None, morfs=[], outfile=None,
+        contexts=None, pretty_print=None, show_contexts=None
+    )
     _defaults.Coverage(
         cover_pylib=None, data_suffix=None, timid=None, branch=None,
         config_file=True, source=None, include=None, omit=None, debug=None,
@@ -69,6 +73,7 @@ class BaseCmdLineTest(CoverageTest):
         cov.report.return_value = 50.0
         cov.html_report.return_value = 50.0
         cov.xml_report.return_value = 50.0
+        cov.json_report.return_value = 50.0
 
         return mk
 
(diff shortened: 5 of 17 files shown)


tests/test_json.py
```diff
@@ -0,0 +1,58 @@
...
+                b = 5
+            """)
+
+        cov = coverage.Coverage()
+        a = self.start_import_stop(cov, "a")
+        output_path = os.path.join(self.temp_dir, "a.json")
+        cov.json_report(a, outfile=output_path)
+        with open(output_path) as result_file:
+            parsed_result = json.loads(result_file.read())
+        del(parsed_result['meta']['timestamp'])
```
Bachmann1234: I should also handle this. Probably by leveraging pytests monkeypatch

nedbat:
Thanks for this.  I haven't had time to look at it yet, but will get to it soon!

Bachmann1234:
@nedbat no worries! I'm in no rush. I would love the feedback whenever you have time. However, I am not blocked as I still need to make time to finish up the tests and such.

Bachmann1234:
howdy! Just checking in to say im still alive. I updated my local master and then rebased this branch to keep it from getting stale. Im thinking ill have some time this weekend to beef up the tests a bit and look at the other todos.

tests/test_json.py
Bachmann1234: addressed, saw assert_recent_timestamp and went with that instead of mocking

Bachmann1234:
Ok, think I got decent tests on this the config file is tested so the only thing this needs (I think) is review/confirmation on the schema and then I need to write some docs.

Bachmann1234:
updated from master

tests/test_json.py
```diff
@@ -0,0 +1,137 @@
...
+"""Test json-based summary reporting for coverage.py"""
+from datetime import datetime
+import json
+import os
+
+import coverage
+from coverage.misc import isolate_module
+from tests.coveragetest import UsingModulesMixin, CoverageTest
+
+os = isolate_module(os)
```
nedbat: No need for this in our own test files.

coverage/cmdline.py
```diff
@@ -118,6 +118,15 @@ class Opts(object):
...
         help="Write the XML report to this file. Defaults to 'coverage.xml'",
     )
+    output_json = optparse.make_option(
+        '-o', '', action='store', dest="outfile",
+        metavar="OUTFILE",
+        help="Write the JSON report to this file. Defaults to 'coverage.json'",
+    )
+    json_pretty_print = optparse.make_option(
+        '', '--pretty-print', action='store_true',
+        help="print the json formatted for human readers",
```
nedbat: Capital P.

coverage/cmdline.py
```diff
@@ -402,6 +411,22 @@ def get_prog_name(self):
...
+            Opts.ignore_errors,
+            Opts.include,
+            Opts.omit,
+            Opts.output_json,
+            Opts.json_pretty_print,
+            Opts.show_contexts,
+            Opts.contexts,
+            ] + GLOBAL_ARGS,
+        usage="[options] [modules]",
+        description="Generate an JSON report of coverage results."
```
nedbat: "an JSON"

coverage/cmdline.py
```diff
@@ -755,6 +788,7 @@ def unglob_args(args):
             report      Report coverage stats on modules.
             run         Run a Python program and measure code execution.
             xml         Create an XML report of coverage results.
+            json        Create a JSON report of coverage results.
```
nedbat: This should go between "html" and "report" (alphabetically)

coverage/control.py
```diff
@@ -862,24 +863,49 @@ def xml_report(
...
             xml_output=outfile, report_contexts=contexts,
             )
+        return self._render_report(self.config.xml_output, XmlReporter(self), morfs)
+
+    def json_report(
+        self, morfs=None, outfile=None, ignore_errors=None,
+        omit=None, include=None, contexts=None, pretty_print=None,
+        show_contexts=None
+    ):
+        """Generate an JSON report of coverage results.
```
nedbat: "an JSON"

coverage/control.py
```diff
@@ -862,24 +863,49 @@ def xml_report(
...
+
+        """
+        self.config.from_args(
+            ignore_errors=ignore_errors, report_omit=omit, report_include=include,
+            json_output=outfile, report_contexts=contexts, json_pretty_print=pretty_print,
+            json_show_contexts=show_contexts
+        )
+        return self._render_report(self.config.json_output, JsonReporter(self), morfs)
+
+    @staticmethod
```
nedbat: I'm not a fan of staticmethod here.  Perhaps this method should be a function in report.py?

coverage/control.py
```diff
@@ -862,24 +863,49 @@ def xml_report(
...
-                ensure_dir_for_file(self.config.xml_output)
+                ensure_dir_for_file(output_path)
                 open_kwargs = {}
                 if env.PY3:
                     open_kwargs['encoding'] = 'utf8'
-                outfile = open(self.config.xml_output, "w", **open_kwargs)
+                outfile = open(output_path, "w", **open_kwargs)
                 file_to_close = outfile
         try:
-            reporter = XmlReporter(self)
```
nedbat, line 882: This moves the construction of XmlReporter out of the try/except.  I hope there isn't an edge case that try/except was handling for us.  The XmlReporter constructor doesn't do much, but it doesn't do nothing.

coverage/jsonreport.py
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        if self.data.has_arcs():
+            self.report_data["totals"].update({
+                'num_branches': self.total.n_branches,
+                'num_partial_branches': self.total.n_partial_branches,
+            })
+
+        outfile.write(
+            json.dumps(
```
nedbat: `json.dump` is better than `dumps` here.

coverage/jsonreport.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        outfile = outfile or sys.stdout
+        self.data.set_query_contexts(self.config.report_contexts)
+        meta = {
+            "version": __version__,
+            "timestamp": str(int(time.time())),
+            "branch_coverage": self.data.has_arcs(),
+            "show_contexts": self.config.json_show_contexts,
+        }
+        self.report_data["meta"] = meta
```
nedbat: Since you don't use meta again later, you can assign the dict to report_data without the intermediate, like you do with "totals" later.

coverage/jsonreport.py
```diff
@@ -0,0 +1,104 @@
...
+        measured_files = {
+            file_reporter.relative_filename():
+            self.report_one_file(file_reporter, analysis)
+            for file_reporter, analysis in get_analysis_to_report(
+                self.coverage,
+                morfs
+            )
+        }
```
nedbat: This dictcomp is a bit too intricate. I missed where the per-file work was being done when I read the function!  I would write this as a full for-statement instead. It's hard to see the structure like this.

coverage/jsonreport.py
```diff
@@ -0,0 +1,104 @@
...
+            "show_contexts": self.config.json_show_contexts,
+        }
+        self.report_data["meta"] = meta
+        measured_files = {
+            file_reporter.relative_filename():
+            self.report_one_file(file_reporter, analysis)
+            for file_reporter, analysis in get_analysis_to_report(
+                self.coverage,
+                morfs
+            )
```
nedbat, line 48: BTW: line length is 100, so this would have fit on one line.

coverage/jsonreport.py
```diff
@@ -0,0 +1,104 @@
...
+        self.total += nums
+        summary = {
+            'covered_lines': nums.n_executed,
+            'num_statements': nums.n_statements,
+            'percent_covered': nums.pc_covered,
+            'missing_lines': nums.n_missing,
+        }
+        reported_file = {
+            'executed_lines': list(sorted(analysis.statements)),
+            'summary': summary,
```
nedbat, line 87: Maybe we don't need "summary" as another level here? Just put its contents directly in the file entry? Or did I miss something?

coverage/jsonreport.py
```diff
@@ -0,0 +1,104 @@
...
+        nums = analysis.numbers
+        self.total += nums
+        summary = {
+            'covered_lines': nums.n_executed,
+            'num_statements': nums.n_statements,
+            'percent_covered': nums.pc_covered,
+            'missing_lines': nums.n_missing,
+        }
+        reported_file = {
+            'executed_lines': list(sorted(analysis.statements)),
```
nedbat: sorted returns a list already, no need to re-list it.

coverage/jsonreport.py
```diff
@@ -0,0 +1,104 @@
...
+            'summary': summary,
+            'missing_lines': list(sorted(analysis.missing))
```
nedbat: ditto on sorted.

coverage/jsonreport.py
```diff
@@ -0,0 +1,104 @@
...
+        }
+        if self.config.json_show_contexts:
+            reported_file['contexts'] = analysis.data.contexts_by_lineno(
+                file_reporter.filename
+            )
+        if self.data.has_arcs():
+            reported_file['summary'].update({
```
nedbat, line 96: I like that the JSON leaves out keys that don't apply.  I wonder if people will feel differently! :)

tests/test_cmdline.py
```diff
@@ -667,6 +672,59 @@ def test_xml(self):
             cov.xml_report(morfs=["mod1", "mod2", "mod3"])
             """)
 
+    def test_json(self):
+        # coverage json [-i] [--omit DIR,...] [FILE1 FILE2 ...]
+        self.cmd_executes("json", """\
+                    cov = Coverage()
```
nedbat: These should all be indented just one stop, as in the rest of the file.

tests/test_cmdline.py
```diff
@@ -871,24 +930,31 @@ def html_report(self, *args_unused, **kwargs_unused):
...
+    ((20, 30, 40, 50), 21, "report", 2),
+    ((20, 30, 40, 50), 29, "html", 0),
+    ((20, 30, 40, 50), 31, "html", 2),
+    ((20, 30, 40, 50), 39, "xml", 0),
+    ((20, 30, 40, 50), 41, "xml", 2),
+    ((20, 30, 40, 50), 49, "json", 0),
+    ((20, 30, 40, 50), 51, "json", 2),
     # Command-line overrides configuration.
-    ((20, 30, 40), 19, "report --fail-under=21", 2),
+    ((20, 30, 40, 50), 19, "report --fail-under=21", 2),
```
nedbat, line 957: Sorry about this bunch of lines.  I should come up with a better way to test this....

tests/test_config.py
```diff
@@ -502,6 +502,17 @@ def test_non_ascii(self):
...
+        self.make_file(".coveragerc", """\
+            [json]
+            pretty_print = True
+            show_contexts = True
+            """)
+        cov = coverage.Coverage()
+
+        self.assertEqual(cov.config.json_show_contexts, True)
+        self.assertEqual(cov.config.json_pretty_print, True)
+
```
nedbat: The other config sections don't have special-purpose test functions. I tend to just add new settings into LOTSA_SETTINGS.  But I could be convinced.

tests/test_json.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class JsonReportTest(UsingModulesMixin, CoverageTest):
+    def _assert_expected_json_report(self, cov, expected_result):
```
nedbat, line 16: If you run "make lint", it will point out where you are missing docstrings.

tests/test_json.py
```diff
@@ -0,0 +1,137 @@
...
+        self.make_file("a.py", """\
+            a = {'b': 1}
+            if a.get('a'):
+                b = 1
+            """)
+        a = self.start_import_stop(cov, "a")
+        output_path = os.path.join(self.temp_dir, "a.json")
+        cov.json_report(a, outfile=output_path)
+        with open(output_path) as result_file:
+            parsed_result = json.loads(result_file.read())
```
nedbat: json.load

coverage/jsonreport.py
```diff
@@ -0,0 +1,104 @@
...
+        `morfs` is a list of modules or file names.
+
+        `outfile` is a file object to write the json to
+
+        """
+        outfile = outfile or sys.stdout
+        self.data.set_query_contexts(self.config.report_contexts)
+        meta = {
+            "version": __version__,
+            "timestamp": str(int(time.time())),
```
nedbat: Would ISO8601 be more friendly?

coverage/jsonreport.py
```diff
@@ -0,0 +1,104 @@
...
+            )
+        if self.data.has_arcs():
+            reported_file['summary'].update({
+                'num_branches': nums.n_branches,
+                'num_partial_branches': nums.n_partial_branches,
+            })
+            reported_file.update({
+                'arcs_missing': analysis.missing_branch_arcs(),
+                'arcs_executed': analysis.arcs_executed(),
+                'arcs_unpredicted': analysis.arcs_unpredicted(),
```
nedbat: This value isn't usually in reports (I think?).  I'd rather leave it out that confuse people at first.

coverage/jsonreport.py
```diff
@@ -0,0 +1,104 @@
...
+                'missing_branch_arcs': analysis.missing_branch_arcs(),
```
nedbat: You've reported this twice (also as 'arcs_missing').  I wonder if we should avoid "arcs" in this report, since it's nerdy and internal.  Reports tend to talk about "branches".

Also, this value is a dict, while "arcs_executed" is a list of pairs.  Let's make them uniform, and both be a list of pairs.

coverage/jsonreport.py
nedbat: `analysis.statements` is not the executed lines, it's the executable lines.  And, awkwardly, it looks like Analysis computed "executed" but does not make it available...?

nedbat: Analysis also has `excluded`, which could be in the report.

nedbat (changes requested):
Matt, thanks so much for taking this on!  My comments run the gamut, I hope you don't mind some pedantry on the English :)

coverage/jsonreport.py
Bachmann1234: I think in my head I was thinking `time.time()` would be easier for machine parsing. However its not like ISO8601 would be hard... and it has the benefit of being much more human readable. So sure thing

coverage/control.py
Bachmann1234: ill take a look

coverage/jsonreport.py
Bachmann1234: all this time and I never actually learned about `load` vs `loads` and `dump` vs `dumps` until today.

coverage/jsonreport.py
Bachmann1234: My thinking was that putting all the counts/stats into a per file summary 'block' made for easier scanning by eye.

Here is a small example. I don't feel super strong about it and can change it. Mostly just walking you though my thinking.

```
"diff_cover/__init__.py": {
            "executed_lines": [
                1,
                3,
                4,
                5
            ],
            "summary": {
                "covered_lines": 4,
                "num_statements": 4,
                "percent_covered": 100.0,
                "missing_lines": 0
            },
            "missing_lines": []
        }
```

coverage/jsonreport.py
Bachmann1234: Think I was overly concerned about python3 generators here...

coverage/jsonreport.py
Bachmann1234: Yeah, the tradeoff is parsers will need to be defensive (or know their settings). However, it's a big win for human readability. The simple case will look simple

You may have noticed by now the whole time writing this my head was going back and forth between the human reader and the machine reader.

Bachmann1234:
> Matt, thanks so much for taking this on! My comments run the gamut, I hope you don't mind some pedantry on the English :)

Not at all. If anything I should be a tad bit embarrassed for "an json" :-P 

Looks like a mixture of small things and at least one or two big errors on my part to correct. I collected all the feedback. Left a few comments and will start working though this. Thanks!

coverage/jsonreport.py
Bachmann1234: Actually looking at this it may have been me doing what xml was doing. Changed to `isoformat`

coverage/control.py
Bachmann1234: Yeah, this change is dangerous. the call to `coverage.get_data()` can throw the exception

I kinda feel like I should just move that out of the constructor. Not just because it makes me not have to rethink this part of the code but under the belief that constructors should exception/side effect free. I also dont think `self.data` is actually doing too much

Open to other thoughts as well of course.

Bachmann1234:
Handled a lot of the smaller feedback. Still have more to look at though

tests/test_json.py
Bachmann1234: addressed

coverage/cmdline.py
Bachmann1234: addressed

coverage/cmdline.py
Bachmann1234: Addressed

coverage/cmdline.py
Bachmann1234: addressed

coverage/control.py
Bachmann1234: addressed

coverage/control.py
Bachmann1234: addressed. Took your advise and moved it into report.py

coverage/control.py
Bachmann1234: I went ahead and moved the dangerous stuff out of the constructor and into the report call

coverage/jsonreport.py
Bachmann1234: addressed

coverage/jsonreport.py
Bachmann1234: addressed

coverage/jsonreport.py
Bachmann1234: addressed

coverage/jsonreport.py
Bachmann1234: addressed

coverage/jsonreport.py
Bachmann1234: addressed

coverage/jsonreport.py
Bachmann1234: addressed

tests/test_cmdline.py
Bachmann1234: addressed

tests/test_config.py
Bachmann1234: addressed, long live LOTSA_SETTINGS

tests/test_json.py
Bachmann1234: addressed

coverage/jsonreport.py
Bachmann1234: Removed the references to arcs and just stuck with the branch percentages in totals

coverage/jsonreport.py
Bachmann1234: done

tests/test_json.py
Bachmann1234: Should be set now

Bachmann1234:
https://github.com/nedbat/coveragepy/pull/825#discussion_r312874321

This piece of feedback is still outstanding. We don't *need* the summary block but I find when looking at the full reports its nice to have these "summary stats" grouped together. Separating these properties from what I have perceived as the more raw data.

So I see it as a tradeoff between making the report a bit easier on the human eye while making it mildly more annoying for the computer to parse since you have to go down one more level to access these properties. 

You see it when just scrolling though a non trivial report
https://gist.github.com/Bachmann1234/bb799bdc5978a81a1e26daeb1881b950

^^ diff cover with contexts enabled.

Bachmann1234:
I think all other feedback has been addressed and I updated the readme to reflect the updated state of the code

Bachmann1234:
updated pr from latest master

tests/test_config.py
```diff
@@ -332,6 +332,10 @@ class ConfigFileTest(UsingModulesMixin, CoverageTest):
         hello = world
         ; comments still work.
         names = Jane/John/Jenny
+
```
nedbat: Lint reports trailing space here.   There's a .editorconfig file in this repo, if you enable it in your editor, it will handle these things automatically.

nedbat:
On the structure of the JSON: OK, files will have a "summary" key.

This is very close to merging!  You'll need to squash the commits.  I can add docs, unless you want to take a pass at it.  I wonder if people will want us to explain the JSON structure? I hope not.. :)

There's also CHANGES.rst and CONTRIBUTORS.txt, which I can tweak up if you like.

nedbat (approved):

Bachmann1234:
added myself to contributors and added a line to changes. Squashed everything down to one commit. I can take a stab at the docs.

doc/config.rst
```diff
@@ -301,3 +301,22 @@ also apply to XML output, where appropriate.
...
+also apply to JSON output, where appropriate.
+
+``json_output`` (string, default "coverage.json"): where to write the json
+report.
+
+``json_pretty_print`` (boolean, default false): controls if fields in the json
+are outputted with whitespace formatted for human consumption (True) or for
+minimum file size (False)
+
+``json_show_contexts`` (boolean, default false): should the json report include
```
Bachmann1234, line 320: I stole the language for this line from the HTML section. I worry its a tad awkward because "on each line" is not quite right in the json context. Its not like the HTML report where every line literally has something I can hover over and see the contexts.

Its more like "Include data in the report showing the contexts recorded for each line"

Bachmann1234:
Ok @nedbat I took a stab at the docs. One place I considered making an update but left alone was the FAQ there is a question that references the XML and HTML reports

```
**Q: I use nose to run my tests, and its coverage plugin doesn't let me create
HTML or XML reports. What should I do?**
```

I could change that to 'HTML, JSON, or XML' reports if you want but I did not think it added much.

the doc changes are in this commit https://github.com/nedbat/coveragepy/pull/825/commits/436eff6a7eaefc39402d77ae8ff827b382d337e8

nedbat:
@Bachmann1234 this looks great.  I am ready to merge it, and will make some minor edits after that.  Squash the commits, and let's get this to master-town! :)

Bachmann1234:
@nedbat Squashed down to one commit and ready to go

Bachmann1234:
I have a watch on the repo so ill keep an eye out for any issues around this (or if something else catches my eye). Im sure something will come up. Also you know how to get ahold of me :-)

PR #850, opened 2019-09-26
Implement the pytest --cov-context option in an "add-on" pytest plugin

ionelmc:
This is a port of https://github.com/pytest-dev/pytest-cov/pull/342

Few notes about this:

- if develop installs are used (they appear to be in the tox configuration) then you get these pesky `Module already imported so cannot be rewritten: __main__` warnings. They are caused by an unfortunate combination of problems: there's a `__main__.py` file in the root of the repo (which shouldn't exist IMO) and the fact that pytest will enumerate all the installed files (importlib_metadata will wrongly enumerate files that aren't normally installed when install is a develop install).
- the `pytest-cov==2.7.1` dep should be replaced to something that is the pytest-cov master
- I've fixed the xdist support (the correct activation point is pytest_sessionstart - otherwise we'd bind to the wrong cov_controller)
- I've renamed `--cov-contexts` to `--cov-context` - I don't get why it was pluralized since there's no comma splitting or something like that in validate_context.

So my suggestion is to take this approach to get more alpha/beta releases of coveragepy out since this whole thing needs overall more work anyway (customization of context, showing something in the reporting about contexts and so on).

### DIFF
--- a/coverage/__init__.py
+++ b/coverage/__init__.py
@@ -1,11 +1,11 @@
 # Licensed under the Apache License: http://www.apache.org/licenses/LICENSE-2.0
 # For details: https://github.com/nedbat/coveragepy/blob/master/NOTICE.txt
-
 """Code coverage measurement for Python.
 
 Ned Batchelder
 https://nedbatchelder.com/code/coverage
 
+PYTEST_DONT_REWRITE
 """
 
 import sys
--- a/coverage/pytest_plugin.py
+++ b/coverage/pytest_plugin.py
@@ -1,0 +1,53 @@
+"""
+PYTEST_DONT_REWRITE
+"""
+import argparse
+
+
+def validate_context(arg):
+    if arg != "test":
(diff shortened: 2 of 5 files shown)


nedbat:
I understand your point about the alpha/betas of coverage.py, but I would rather not have this plugin here.  This code has far too much pytest knowledge in it: it should be part of pytest-cov.  This code even has details of the internals of the pytest-cov plugin.

The overall work remaining ("customization of context, showing something in the reporting about contexts and so on") is mostly on the coverage.py side of things, not in the pytest integration.

The code should live in the repo that makes the most sense as its forever home.  If there's a pytest-cov plugin, then this code should be part of it.  If somewhere down the line, we completely fold pytest-cov into coverage.py, then this code will be part of that folding.

Can we apply your fixes to pytest-dev/pytest-cov#342 ?

ionelmc:
Oooof ... well what can I say.

PR #857, opened 2019-10-16
Print TOTAL of coverage anyway with 1 or more files

nascimento:
We use coveragepy on GitlabCi pipelines to get coverage total of a repository and get metrics for other purposes. (https://docs.gitlab.com/ee/ci/yaml/#coverage)

Some times we have repositories with 1 file and many libs (that we run coverage on another pipeline) and we need to grep (with regex) the total of coverage. If coveragepy print TOTAL every run, anyway, we can make a generic script to get the coverage.

### DIFF
--- a/coverage/summary.py
+++ b/coverage/summary.py
@@ -110,16 +110,15 @@ class SummaryReporter(object):
         for line in lines:
             self.writeout(line[0])
 
-        # Write a TOTAl line if we had more than one file.
-        if self.total.n_files > 1:
-            self.writeout(rule)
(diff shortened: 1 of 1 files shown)


nedbat:
Thanks, but I would rather keep the output reasonable for people to read.  When this has come up in the past, I've suggested that parsing the last number from the last line of output will already work in both the 1-file and N-file cases.  Does that work for you?

Also, in 5.0, we have a new "coverage json" command that writes a supported machine-readable JSON file if you want access to the results for scripting.

Issue #587, opened 2017-06-16
Option to output lcov.info format?


nedbat:
*Originally reported by* **Matthew Crenshaw (Bitbucket: [sgtsquiggs](https://bitbucket.org/sgtsquiggs), GitHub: [sgtsquiggs](https://github.com/sgtsquiggs))**

----------------------------------------

So that we can use coverage.py with IDE plugins that read the standard lcov.info format and highlight covered/uncovered lines. Would be super great!

----------------------------------------
- Bitbucket: https://bitbucket.org/ned/coveragepy/issue/587

nedbat:
Interesting idea.  Do you have a link to information about the file format?

nedbat:
*Original comment by* **Matthew Crenshaw (Bitbucket: [sgtsquiggs](https://bitbucket.org/sgtsquiggs), GitHub: [sgtsquiggs](https://github.com/sgtsquiggs))**

----------------------------------------

[lcov format](http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php)

       Following  is  a  quick  description of the tracefile format as used by
       genhtml, geninfo and lcov.

       A tracefile is made up of several human-readable lines of text, divided
       into sections. If available, a tracefile begins with the testname which
       is stored in the following format:

         TN:<test name>

       For each source file referenced in the .da file,  there  is  a  section
       containing filename and coverage data:

         SF:<absolute path to the source file>

       Following is a list of line numbers for each function name found in the
       source file:

         FN:<line number of function start>,<function name>

       Next, there is a list of execution counts for each  instrumented  func‐
       tion:

         FNDA:<execution count>,<function name>

       This  list  is followed by two lines containing the number of functions
       found and hit:

         FNF:<number of functions found>
         FNH:<number of function hit>

       Branch coverage information is stored which one line per branch:

         BRDA:<line number>,<block number>,<branch number>,<taken>

       Block number and branch number are gcc internal  IDs  for  the  branch.
       Taken  is either '-' if the basic block containing the branch was never
       executed or a number indicating how often that branch was taken.

       Branch coverage summaries are stored in two lines:

         BRF:<number of branches found>
         BRH:<number of branches hit>

       Then there is a list of execution counts  for  each  instrumented  line
       (i.e. a line which resulted in executable code):

         DA:<line number>,<execution count>[,<checksum>]

       Note  that  there  may be an optional checksum present for each instru‐
       mented line. The current geninfo implementation uses  an  MD5  hash  as
[...]

PR #863, opened 2019-10-29
DO-NOT-MERGE: Patched-on LCOV file format support for coveragepy v4.x

gpshead:
This is gross as is, it was originally done in a way that minized diffs.  :/

This PR is for discussion purposes only. I expect a new PR to be created by anyone interested on master to add lcov support in a non-hacky manner. :)

I'm making the PR on a branch of v4.5.x as our existing changes do not apply cleanly after some of the refactoring in the git master (v5) code.

We are using a patched 4.4.2 internally at the time of PR creation.  This patch comes from Google where we've been carrying around some variant of this patch on top of coverage for a very long time.  It clearly belongs upstream (starting that here), as the LCOV format seems to be supported by a variety of IDEs and other systems(?).

I don't expect to do future work in this PR itself.  Follow https://github.com/nedbat/coveragepy/issues/587 for actual details.

Lacking: Any LCOV format unittests. (sadface)

(I did not successfully run the coverage testsuite on this PR but that might have been my lack of trying and possibly a weird environment?  They don't seem to run properly for me when following the v4.5.x branch howto.txt instructions but I didn't try to debug why)

### DIFF
--- a/coverage/annotate.py
+++ b/coverage/annotate.py
@@ -7,6 +7,8 @@ import io
 import os
 import re
 
+from coverage import files
+
 from coverage.files import flat_rootname
 from coverage.misc import isolate_module
 from coverage.report import Reporter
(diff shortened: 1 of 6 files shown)


nedbat:
@gpshead Thanks. I have to admit, when I saw you started with "this is gross," I thought, "come on, how gross could it be?"  Then I looked at the diffs... :)

Is there a reference anywhere for the lcov file format?  I found this: http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php

nedbat:
Oops, that was already in #587 (*slurps coffee*).

nedbat:
@gpshead a couple of questions:
1. Your code outputs "BA" lines, but the man page describing the format doesn't mention them.  Do other tools understand these lines? The comment in the code is helpful for understanding the semantics.
2. This code is based on annotate, which is the least-used reporting method.  It marks some lines executed that the other reporting methods do not.  For example, non-blank lines in docstrings, or the def line after a decorator.  Will it be OK if the real lcov support changes how these lines are marked?

abitrolly:
Fixed by https://github.com/nedbat/coveragepy/pull/1289 ?

gpshead:
oh boy!  I can't wait to try that out and delete this ugly pile of code. :)

PR #864, opened 2019-10-30
Optionally skip empty files in reports

reybog90:
In a large code base it is fairly common to have many empty `__init__.py` files. For example in Django applications, you generally end up with files like:
`<app>/management/__init__.py`
`<app>/management/commands/__init__.py`
`<app>/migrations/__init__.py`
`<app>/templatetags/__init__.py`
`<app>/tests/__init__.py`
... for each Django app in the project. Most of them are generally empty and pollute the reports.

This change adds a `--skip-empty` option to skip them in the reports. It follows the same pattern as `--skip-covered`.

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -93,6 +93,10 @@ class Opts(object):
         '--skip-covered', action='store_true',
         help="Skip files with 100% coverage.",
     )
+    skip_empty = optparse.make_option(
+        '--skip-empty', action='store_true',
+        help="Skip empty files.",
+    )
     show_contexts = optparse.make_option(
         '--show-contexts', action='store_true',
         help="Show contexts for covered lines.",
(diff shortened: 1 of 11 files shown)


nedbat:
Since the `__init__.py` files will be 100% covered, why isn't `--skip-covered` enough?

reybog90:
Because I like to keep the fully covered files in the report. That makes it easier to see where all the code is and what is the state of each package. It also makes the report more optimistic (assuming the coverage is high).
The empty `__init__.py` files to me feel out of place in a code coverage report, since there was nothing to cover and many of them are not expected to have any code. In my opinion, the only time it makes sense to show them is when debugging coverage scripts or looking for empty files to delete. At the same time, filtering them out by name won't work, because some of them do contain code.
Since it was very simple to add and I find this option useful, I just went in and made the PR.

nedbat:
Thanks! Do you have a name I can put into the CONTRIBUTORS file? :)

reybog90:
Thanks a lot! You can add "Reya B", if you want :)

Issue #868, opened 2019-11-12
Expose numbits_to_nums() as a SQL function


simonw:
The `register_sqlite_functions()` function currently adds four custom SQL functions:

https://github.com/nedbat/coveragepy/blob/694435b844760ca0443cf87bf73a537f06c37d7d/coverage/numbits.py#L153-L156

I'd appreciate it if it also exposed a version of the `numbits_to_nums()` function that returns a JSON array of line numbers. SQLite has built-in JSON functions that could take advantage of this, but more importantly it would make it easier to run queries that show a human-readable version of what's going on.

simonw:
I'm planning a Datasette plugin which exposes these SQL functions (via [this hook](https://datasette.readthedocs.io/en/stable/plugins.html#prepare-connection-conn)), so I can use Datasette to muck around with my `.coverage` files.

This feature is by no means a blocker on what I'm planning to do - I can always register the exact same function in my own plugin - but I thought it would also make sense as a feature of the `register_sqlite_functions(connection):` function here.

nedbat:
Fixed in #868

PR #869, opened 2019-11-12
Expose numbits_to_nums() SQL function

simonw:
See #868

### DIFF
--- a/coverage/numbits.py
+++ b/coverage/numbits.py
@@ -132,6 +132,7 @@ def register_sqlite_functions(connection):
     * :func:`numbits_intersection`
     * :func:`numbits_any_intersection`
     * :func:`num_in_numbits`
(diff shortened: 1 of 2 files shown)


nedbat:
Thanks, I've never used the SQLite JSON support before.  One question I guess I sidestepped that now seems important: how to document the SQLite behavior of these functions? `numbits_to_nums` says "Returns a list of ints."  Will that be clear to people that it's a JSON formatted string?

simonw:
That's a good call - the documentation should definitely clarify that this returns a string that is a JSON list of integers, since many developers won't be familiar with this pattern.

nedbat:
I added a sentence: https://github.com/nedbat/coveragepy/commit/2aeb189280ad89b1fdc67529438c6c98668b9d6f.

Issue #895, opened 2019-12-18
os.path.relpath fails when saving a coverage file with coverage 5 in a folder from a different Windows drive


ogrisel:
**Describe the bug**

In the scikit-learn test suite we have a test that launches a subprocess and the atexit registry triggers the saving of a coverage file in a tempfolder in d:\ while the current working directory has been changed by `pytest --pyargs` to c:\ in the meantime.

```
E                   AssertionError: Error in atexit._run_exitfuncs:
E                   Traceback (most recent call last):
E                     File "c:\miniconda\envs\testvenv\lib\ntpath.py", line 562, in relpath
E                       path_drive, start_drive))
E                   ValueError: path is on mount 'D:', start on mount 'c:'

cmd        = ['c:\\miniconda\\envs\\testvenv\\python.exe', 'C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\tmp32ajt78f_src_test_sklearn.py']
coverage_rc = None
cwd        = 'c:\\miniconda\\envs\\testvenv\\lib\\site-packages'
...
```

I do not understand why the traceback starts at `atexit._run_exitfuncs` and then goes to `ntpath.relpath` directly without going through the function of the `coverage` module. I suspect that the culprit is:

https://github.com/nedbat/coveragepy/blob/coverage-5.0/coverage/sqldata.py#L982

When rolling back to coverage 4.5.3, the problem disappears.

**To Reproduce**

This problems happens with coverage 5.0 and Python 3.7.5 with the latest pytest and pytest-cov.

It seems tricky to reproduce, in particular I do not understand why this only happens in the subprocess tests of scikit-learn and never in the main process itself. Maybe the current working directory is restored to the d:\ tempfolder prior to writing the coverage files?

Anyhow the details can be found in the CI logs of this pull request where I have patched the `ntpath.relpath` function to display more debug info:

https://github.com/scikit-learn/scikit-learn/pull/15909

I will open a PR to fix the issue in coveragepy.

It looks similar to #824 but the error message is actually different.

nedbat:
Ironically, the relpath line in question is only there to help with a different Windows problem.  Perhaps catching a ValueError from relpath, and skipping it will be enough?

ogrisel:
Will do this.

ogrisel:
Done in #896.

nedbat:
Fixed in c09c2c203286db3cae1148bd02b1db9a9d524af1

asottile:
[...]

PR #896, opened 2019-12-18
Workaround issue with relpath on Windows

ogrisel:
Fix for #895.

Writing a non-regression test for this is going to be tricky. It might be possible using https://stackoverflow.com/a/827398/163740 to see if there is another drive / mount available on the CI machine and `chdir` to it before writing the coverage file but it's complicated.

Let me know if you want me to try to write such a non-regression test or not.

Besides, I have tested this branch on the scikit-learn CI https://github.com/scikit-learn/scikit-learn/pull/15909 and confirms that this fixes the original problem.

### DIFF
--- a/coverage/sqldata.py
+++ b/coverage/sqldata.py
@@ -979,7 +979,14 @@ class SqliteDb(SimpleReprMixin):
         # SQLite on Windows on py2 won't open a file if the filename argument
         # has non-ascii characters in it.  Opening a relative file name avoids
         # a problem if the current directory has non-ascii.
-        filename = os.path.relpath(self.filename)
+        try:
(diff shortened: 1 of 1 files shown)


ogrisel:
The travis failure is:

```
worker 'gw2' crashed while running 'tests/test_oddball.py::RecursionTest::test_long_recursion'
```

It seems unrelated to this PR.

ogrisel:
@nedbat please let me know if you need anything else from my side.

nedbat:
Looks great, thanks. I'll amend the commit to add changelog and bug number info.

ogrisel:
Cheers!

PR #908, opened 2019-12-30
Added arm64 job for Travis-CI

ossdev07:
Travis-CI has added support for ARM64. Added ARM64 jobs in Travis-CI.


ossdev07:
HI @nedbat , Can you please have a look at this PR.

nedbat:
Help me understand the value here.  Coverage.py doesn't have tricky processor-specific code that I know of.

rhenwood-arm:
hi @nedbat 
I've been working with @ossdev07 on a downstream project on AArch64 (ARM64) that has coveragepy as a dependency. As you point out, coveragepy is a pure python project with a straightforward Travis-CI build job and that doesn't have a binary release. In this case, there is very little benefit in testing on alternative architectures, and this PR can probably be held in reserve until it is needed.

nedbat:
@rhenwood-arm Actually, coverage.py has a C extension, so it does compile code.  But I didn't realize that processor differences in that C code were likely enough to warrant the extra builds and time. Do they?

.travis.yml
```diff
@@ -8,14 +8,33 @@ dist: xenial
...
+      python: 3.7
+    - arch: arm64
+      python: 3.8
+    - arch: amd64
+      python: 3.8
+    - arch: amd64
+      python: pypy2.7-6.0
+    - arch: amd64
+      python: pypy3.5-6.0
+      # Skipping for Arm64, pypy has aarch64 support - but it isn't released yet.
```
rhenwood-arm: I find this very hard to read with 'arm64' and 'amd64' interspersed. Can the architectures be grouped together to make it more readable?

Also, python 2.7 can be omitted for arm64 as it is EOL.

rhenwood-arm (changes requested):

rhenwood-arm:
Ah! indeed - I didn't spot ctracer!
Native code provides an opportunity to break on a given platform, so testing is typically valuable. As a more general goal, I am working on figuring out how Arm can support wheels and manylinux2014 activities. So this library is a good example to pursue for me.

I clicked around and I couldn't find the Arm test so I'm unsure what the delay is. @ossdev07 do you have the test duration on AArch64?

ossdev07:
Hi @nedbat
Modified PR, added arm64 job for just python3.8 as it has already a lot of existing tests which were taking considerable amount of time. 
I think it's good to have it tested on aarch64 platform as well.

nedbat:
Thanks!

PR #917, opened 2020-01-08
Support patterns in module names

kvalev:
This PR adds support for using patterns in the `source` module names. For example the following configuration will match both the `coverage`, as well as the `mycoverage` modules:
```
[run]
source = .*coverage
```

### DIFF
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -242,6 +242,7 @@ class ModuleMatcher(object):
     """A matcher for modules in a tree."""
     def __init__(self, module_names):
         self.modules = list(module_names)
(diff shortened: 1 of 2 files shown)


kvalev:
Let me know if you agree in principle with the changes in the PR and I will also update the documentation accordingly.

kvalev:
@nedbat Any chance of getting this PR approved?

nedbat:
Hi, sorry that I let this get so old.  Can you tell me more about the problem you are solving with this change?

kvalev:
We are using one framework, which generates code in temporary modules on the fly, so in case we want to measure the code coverage of the generated modules (we do), we need a way to match the dynamic module names.

PR #932, opened 2020-02-14
Add --no-skip-covered to allow negation of --skip-covered

### LINKED ISSUE #779: `--no-skip-covered` to negate `--skip-covered`

For test run outputs, I set `skip_covered = true` in my `.coveragerc` (for example [here](https://github.com/asottile/all-repos/blob/1a3d612bd90e826c99257ebf7ff6e1762742d479/.coveragerc#L14)).  I do this to avoid the terminal spew of the 50+ 100% covered files.

However, for html outputs, I'd still like to have files included when running `coverage html`, even if they are 100% covered.

I think the _simplest_ solution would be to add `--no-skip-covered` options to the various `coverage html` / `coverage report` entrypoints which would override the `--skip-covered` option.

Not really seriously considered, but `sed -i /skip_covered/d .coveragerc` before running `coverage html` -- pretty haxy though ;)

**Additional context**

I'm mostly trying to target automated CI here, in particular azure pipelines coverage reporting

asottile:
Resolves #779

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -93,6 +93,10 @@ class Opts(object):
         '--skip-covered', action='store_true',
         help="Skip files with 100% coverage.",
     )
+    no_skip_covered = optparse.make_option(
+        '--no-skip-covered', action='store_false', dest='skip_covered',
+        help="Disable --skip-covered.",
+    )
     skip_empty = optparse.make_option(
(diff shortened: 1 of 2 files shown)


nedbat:
This is now released as part of [coverage 5.2](https://pypi.org/project/coverage/5.2).

PR #942, opened 2020-02-22
Escape backslashes in windows filepath in config docs

hoefling:
[ReST uses backslashes for escaping](https://docutils.sourceforge.io/docs/user/rst/quickref.html#escaping). Using single backslashes in the windows filepath thus results in a minor rendering issue in the docs (path separator is missing). This PR fixes this.


nedbat:
Thanks!

PR #945, opened 2020-02-24
htmlfiles: Handle localStorage not accessible

### LINKED ISSUE #944: Handle cases without localStorage for sorting
I was recently trying to make my browser more "secure" by locking some domains from accessing cookies etc. And I found that somehow I broke coverage's HTML report.
I was surprised, because I didn't expect coverage to need any form of storage ...
On digging further I saw that localStorage was being used to persist the sorting column

If localStorage was causing an error for some reason - don't save the column that was used for sorting and just sort by the default column

Enabling localStorage for coverage is obviously an option :)
And something I'm going to do right now - but I'm not sure if its a good idea for coverage to throw an error in case it was unable to do just sorting ...
Note: If this JS error occurs, I can't do any sorting (even temporarily on this page)

AbdealiLoKo:
In some cases, if based on the browser's settings - localStorage
is not accessible, fallback and don't save the sort-columns into
localStorage.
While the UX is a little inconvenient, at least the page doesn't
break - sorting on columns is still possible, but not retained
between pages.

Resolves https://github.com/nedbat/coveragepy/issues/944

### DIFF
--- a/coverage/htmlfiles/coverage_html.js
+++ b/coverage/htmlfiles/coverage_html.js
@@ -172,7 +172,10 @@ coverage.index_ready = function ($) {
     // Look for a localStorage item containing previous sort settings:
     var sort_list = [];
     var storage_name = "COVERAGE_INDEX_SORT";
-    var stored_list = localStorage.getItem(storage_name);
+    var stored_list = undefined;
+    try {
+        stored_list = localStorage.getItem(storage_name);
+    } catch(err) {}
 
     if (stored_list) {
         sort_list = JSON.parse('[[' + stored_list + ']]');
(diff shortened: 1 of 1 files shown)


coverage/htmlfiles/coverage_html.js
```diff
@@ -172,7 +172,10 @@ coverage.index_ready = function ($) {
     // Look for a localStorage item containing previous sort settings:
     var sort_list = [];
     var storage_name = "COVERAGE_INDEX_SORT";
-    var stored_list = localStorage.getItem(storage_name);
+    var stored_list = undefined;
+    try {
```
AbdealiLoKo, line 176: Note - didn't look like there was any better way of testing this as merely accessing the variable caused the error. So, try/catch seemed to be the only option

nedbat:
Thanks!

nedbat:
This became 48c73b88bf2f3a5cdb404b20afa34cb2cb940e0f

nedbat:
This is now available in [coverage v5.0.4](https://pypi.org/project/coverage/5.0.4/)

PR #960, opened 2020-03-19
Displaying timezone information in HTML report

xyb:


nedbat:
Thanks for doing this, and sorry for my silence.

I'm curious why the timezone is important on the report?

I have some concerns about the code.  On my Mac, on 2.7, the timezone is shown as "+2000", but on 3.7, it's "-0400".   I'm not that comfortable with maintaining the fallback code for format_local_datetime.  How would you feel about displaying the timezone only for Python versions that can do it natively (3.6+)?

xyb:
> I'm curious why the timezone is important on the report?

I have multiple CI servers that have different time zones, so I would like to know the exact time of the test report.

> How would you feel about displaying the timezone only for Python versions that can do it natively (3.6+)?

Sounds good. The backward compatible code is ugly, let's get rid of it :)

nedbat:
Thanks! I merged it, and had to fix the test (sometimes time zones are -XXXX instead of +XXXX).

nedbat:
This is available in [coverage 5.2.1](https://pypi.org/project/coverage/5.2.1).

PR #966, opened 2020-03-31
Extending jsonreport.py to also include all branch attributes

SZVector:
The current `jsonreport` does not include the attributes for `covered_branches` and `missing_branches`.

This PR extends `jsonreport.py` such that it includes these attributes; it also updates `test_json.py` to account for these newly-dumped attributes.

Current test status:

```
  py27: commands succeeded
SKIPPED:  py35: InterpreterNotFound: python3.5
  py36: commands succeeded
SKIPPED:  py37: InterpreterNotFound: python3.7
SKIPPED:  py38: InterpreterNotFound: python3.8
SKIPPED:  py39: InterpreterNotFound: python3.9
SKIPPED:  pypy2: InterpreterNotFound: pypy2
SKIPPED:  pypy3: InterpreterNotFound: pypy3
```

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -114,6 +114,7 @@ Rodrigue Cloutier
 Roger Hu
 Ross Lawley
 Roy Williams
+Salvatore Zagaria
 Sandra Martocchia
 Scott Belden
 Sigve Tjora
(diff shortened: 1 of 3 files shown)


aytey:
@nedbat any thoughts on this? This seems like a great change, as it allows us to get access to data that would otherwise be missing from the json report.

nedbat:
Thanks, I'll add the fields to the individual files also.

nedbat:
This is available in 5.1: https://pypi.org/project/coverage/5.1/

PR #969, opened 2020-04-03
Allow plugins to alter the paths config

### LINKED ISSUE #967: paths cannot be altered by plugins
https://coverage.readthedocs.io/en/latest/config.html#paths shows how to user can set paths. This cannot be achieved by plugins due to paths not being defined as a set/get option; but rather directly as attribute on the configuration object. Due to https://github.com/nedbat/coveragepy/blob/master/coverage/control.py#L272-L276 the plugin cannot reliably access and extend this.

Came up while trying to address https://github.com/asottile/covdefaults/issues/3

gaborbernat:
Resolves #967.

### DIFF
--- a/coverage/config.py
+++ b/coverage/config.py
@@ -421,6 +421,9 @@ class CoverageConfig(object):
         `value` is the new value for the option.
 
         """
+        if option_name == "paths":
+            self.paths = value
+            return
 
         # Check all the hard-coded options.
         for option_spec in self.CONFIG_FILE_OPTIONS:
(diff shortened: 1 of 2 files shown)


gaborbernat:
@nedbat any chance to land this? 🤔

nedbat:
Thanks, this fixes an oversight.  If you don't mind me asking, what plugin will you use it for?

gaborbernat:
covdefaults, linked in the issue.

nedbat:
This is available in 5.1: https://pypi.org/project/coverage/5.1/

PR #982, opened 2020-04-30
Add --precision option to report command

teake:
This PR adds a `--precision <N>` CLI option to `coverage report`, so you don't have to write a `.coveragerc` file if you only want to adjust the precision.

Note: I did a few ad-hoc tests and those ran fine, but I couldn't get the test suite to pass. Since I didn't want to spend too much time on this I decided to submit the PR as-is. Hopefully you can fix the failing tests easily, but otherwise I won't be too embarrassed if this gets rejected.

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -146,6 +146,13 @@ class Opts(object):
             "to be run as 'python -m' would run it."
         ),
     )
+    precision = optparse.make_option(
+        '', '--precision', action='store', metavar='N', type=int,
+        help=(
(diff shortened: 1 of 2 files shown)


nedbat:
Thanks, I'll take a look.  Don't forget that you can put settings into tox.ini, setup.cfg, or pyproject.toml files also.

nedbat:
Thanks, I finished it up in 916a0c92

nedbat:
This is now released as part of [coverage 5.2](https://pypi.org/project/coverage/5.2).

PR #984, opened 2020-05-04
tests/gold/html/Makefile: support more shells

rillig:
Not all shells support the `[[` keywords. Among them are many `/bin/sh`.

### DIFF
--- a/tests/gold/html/Makefile
+++ b/tests/gold/html/Makefile
@@ -7,7 +7,7 @@ help:
 
 complete: 	## Copy support files into directories so the HTML can be viewed properly.
 	@for sub in *; do \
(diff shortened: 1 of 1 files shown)


tests/gold/html/Makefile
```diff
@@ -7,7 +7,7 @@ help:
 
 complete: 	## Copy support files into directories so the HTML can be viewed properly.
 	@for sub in *; do \
-		if [[ -f $$sub/index.html ]]; then \
+		if if [ -f "$$sub/index.html" ]; then \
```
nedbat:
```suggestion
		if [ -f "$$sub/index.html" ]; then \
```

PR #1013, opened 2020-07-19
tests/test_cmdline.py::CmdLineStdoutTest::test_cmd_help fails on large terminals

ghost:
The test `CmdLineStdoutTest::test_cmd_help` in `tests/test_cmdline.py` checks for a certain number of lines in the output.
However, optparse formats the output according to the terminals column count (often given by the environment variable `COLUMNS`; if detection fails, 80 is used).
The smaller the terminal the more line breaks occur.

The current test tests against at least 30 lines. Which is fulfilled on smallish terminals (e.g. 80 columns).
However, on larger terminals (e.g. 200 columns) every option description fits into one line reducing the count to 23 and the test fails.

This pull request changes the test to test against 20 lines (to have some leeway) to also support large terminals.

The number of lines in the help output of a command depends on the terminal.

An alternative way would be to always run the test suite with `COLUMNS=80` set as environment variable.
However, this puts the burden onto the user which have to ensure to have the "correct" environment.

### DIFF
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -861,7 +861,7 @@ class CmdLineStdoutTest(BaseCmdLineTest):
         lines = out.splitlines()
         self.assertIn("<pyfile>", lines[0])
         self.assertIn("--timid", out)
-        self.assertGreater(len(lines), 30)
+        self.assertGreater(len(lines), 20)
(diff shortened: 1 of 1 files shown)


nedbat:
Thanks!

PR #1017, opened 2020-08-04
Add better details on how to use `[coverage:run]` to exclude files

siliconninja:
This PR updates the documentation for excluding files. I initially had trouble when setting it up, and after a bit of searching, I found the name of the file that `[coverage:run]` is specified in (`setup.cfg` or something else, as described in https://coverage.readthedocs.io/en/latest/config.html?highlight=setup.cfg#configuration-reference), and I realized I had to use `[coverage:run]` and not `[run]`.

Please feel free to contact me on this PR with any questions or comments.

### DIFF
--- a/doc/source.rst
+++ b/doc/source.rst
@@ -26,7 +26,7 @@ When running your code, the ``coverage run`` command will by default measure
 all code, unless it is part of the Python standard library.
 
 You can specify source to measure with the ``--source`` command-line switch, or
-the ``[run] source`` configuration value.  The value is a comma- or
+the ``[coverage:run] source`` configuration value.  The value is a comma- or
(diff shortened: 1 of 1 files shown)


nedbat:
Hi, sorry for the difficulty with the settings.

The section name varies depending on the file you use.  In a `.coveragerc` file, the section is called `[run]`.  In other files which are not specific to coverage.py, the section needs to specify that it is for coverage, and so is called `[coverage:run]`.  In a pyproject.toml file, it's called `[tool.coverage.run]`. 

I'm not sure how to make that clearer throughout the docs.  It's not right to change the name of the section to `[coverage:run]`, since that might not be the correct section name.  I'm open to suggestions.  I'll close this PR now, but I am interested to hear improvements.

PR #1021, opened 2020-08-17
Removed python_requires="<4"

### LINKED ISSUE #1020: `python_requires="<4"` creates issues with dependency solvers
- `python_requires="<4"` is defined in https://github.com/nedbat/coveragepy/blob/aa6cb95e366f479cddd8cd30aee940393ead6296/setup.py#L122 
- This requirement causes problems with dependency solvers, such as `poetry`:

```bash
>>> poetry add --allow-prereleases --dev coverage          
Using version ^5.2.1 for coverage

Updating dependencies
Resolving dependencies... (0.0s)

[SolverProblemError]
The current project's Python requirement (>=3.6) is not compatible with some of the required packages Python requirement:
  - coverage requires Python >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <4
```

engnadeau:
- Python 4 doesn't exist, this requirement is redundant
- Fixes #1020

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -119,7 +119,7 @@ setup_args = dict(
         ),
         'Issues': 'https://github.com/nedbat/coveragepy/issues',
     },
-    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <4",
+    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*",
 )
 
 # A replacement for the build_ext command which raises a single exception


hartwork (approved):

Embeddave:
Hi @nedbat are there any plans to make this change to `python_requires`?

I am running into the same issue as #1020.

I also don't understand the need to specify "<4.0", and I don't want to "surprise" any downstream users by suddenly pinning to a major version of Python

nedbat:
From the discussion in #1020:
> can you help me understand how you ended up with ">=3.6" in your pyproject.yaml? Poetry initializes it with "^3.6", which works.

Embeddave:
Thank you Ned.

I changed the default from the caret because I don't want to break things for other people (like myself, in other projects) that haven't always specified their Python version as "<4.0". They will also get a failure on install, not just from poetry, but from pip itself.

see comment here from @henryiii 
https://mobile.twitter.com/HenrySchreiner3/status/1362183452343296004
> Practically, if you haven’t always done it, pip will pull an older release (the first without the <4), which is _not _ more likely to work with a newer Python! Python is not SemVer’d, as well. 4.0 is not supposed to be like 3.0, if ever reached

I am already running into this issue with my own set of tools that depend on each other, and now I have the same issue trying to install `pytest-cov`.

Beyond avoiding that issue, I don't understand the need to specify Python "<4.0" if there's no concrete plans for 4.0, and Python follows CalVer instead of SemVer.

from this discussion where @matthewfeickert asks
https://mobile.twitter.com/HEPfeickert/status/1362170933658738701
> Is there any need to put a "<4.0" in your #Python library's setup.cfg/setup.py python_requires? I have never done this, and it seems that most people don't, but occasionally I've seen people do this. Should we? Or are we all expecting releases out now to not be used in v4.0 era?

Happy to hear that I'm wrong but this is my current understanding.

matthewfeickert:
People may already be aware of this, so apologies if I'm reposting or missing the context if I had read more of the links in this Issue, but @NickleDave is currently having some [real world problems with other libraries specifying `python_requires = >=3.X, <4.0`](https://twitter.com/nicholdav/status/1366978826790666240?s=20). So I think his situation can be used as a case study of why this PR is a good idea. :+1:

nedbat:
I still think poetry is doing the wrong thing.  But 4.0 is a long way off, and this can relieve some pain.  If 4.0 does emerge, and needs to be excluded, we will all as a community have to revisit how these version pins work.  Maybe it will be easier then.

NickleDave:
:raised_hands: thank you so much! 

You are totally right (as far as I understand it) that Poetry's solver is being overly strict.

And yes the community will need to come to a solution.
I will try raising a couple questions in their GitHub discussion to see if I can better understand their thinking.

henryiii:
If you need some ammunition, this has been just fixed in the official docs: https://github.com/pypa/packaging.python.org/pull/850 (see the linked core dev discussion).

wimglenn:
Hi Ned, any chance we can get a release with the updated metadata?  The poetry 5.5 wheel still has this <4 thing in

nedbat:
Just for completeness: this is in coverage 6.0.

NickleDave:
Thank you @nedbat much appreciated

PR #1023, opened 2020-08-25
allow serving the html content for coverage

gtmanfred:
Hi!

I have been playing with this for as a tool for local stuff, and I was wondering if there was any interest in it being added to coveragepy, and I would spend the time to make it compatible for python2.

What it does is start a simple html server in a thread, and then use the port that is sent back by the simple http server, and then use the built webbrowser module to open the directory in the browser for viewing.

Let me know if yall are interested, if not, that is ok too.

Thanks!
Daniel

### DIFF
--- a/coverage/html.py
+++ b/coverage/html.py
@@ -3,11 +3,16 @@
 
 """HTML reporting for coverage.py."""
 
+import concurrent.futures
 import datetime
+import http.server
 import json
 import os
 import re
 import shutil
+import sys
+import threading
+import webbrowser
 
 import coverage
 from coverage import env
@@ -30,6 +35,26 @@ STATIC_PATH = [
     # Our htmlfiles directory.
     os.path.join(os.path.dirname(__file__), "htmlfiles"),
 ]
+
+
+def serve_htmldir(htmldir):
+    """ Serve htmldirectory for coverage files"""
+    future = concurrent.futures.Future()
+
+
+    def start_server():
+        httpd = http.server.HTTPServer(('127.0.0.1', 0), http.server.SimpleHTTPRequestHandler)
+        future.set_result(httpd.server_port)
+        httpd.serve_forever()
+
+
+    thread = threading.Thread(target=start_server, daemon=True)
+    thread.start()
+    webbrowser.open_new(f'http://localhost:{future.result()}/{htmldir}')
+    try:
+        thread.join()
+    except KeyboardInterrupt:
+        pass
 
 
 def data_filename(fname, pkgdir=""):
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -170,6 +170,10 @@ class Opts(object):
             "'pyproject.toml' are tried. [env: COVERAGE_RCFILE]"
         ),
     )
+    serve = optparse.make_option(
+        '', '--serve', action='store_true', default=False,
+        help='Serve htmlcov directory to default webbrowser',
+    )
     source = optparse.make_option(
         '', '--source', action='store', metavar="SRC1,SRC2,...",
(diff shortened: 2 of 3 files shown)


coverage/html.py
```diff
@@ -32,6 +37,26 @@
...
+    future = concurrent.futures.Future()
+
+
+    def start_server():
+        httpd = http.server.HTTPServer(('127.0.0.1', 0), http.server.SimpleHTTPRequestHandler)
+        future.set_result(httpd.server_port)
+        httpd.serve_forever()
+
+
+    thread = threading.Thread(target=start_server, daemon=True)
```
graingert, line 51: I don't think you need a thread here, `webbrowser.open_new` will spawn a new process

gtmanfred: The webserver has to be started before `webbrowser.open_new()` is run, otherwise, it will timeout trying to connect to a port that isn't being listened on.

coverage/html.py
```diff
@@ -32,6 +37,26 @@
 ]
 
 
+def serve_htmldir(htmldir):
+    """ Serve htmldirectory for coverage files"""
+    future = concurrent.futures.Future()
+
+
+    def start_server():
+        httpd = http.server.HTTPServer(('127.0.0.1', 0), http.server.SimpleHTTPRequestHandler)
```
graingert, line 46: should support ivp6 too

gtmanfred: Why? It is just listening on localhost?

coverage/html.py
graingert: I just tried this:

```python
import http.server
import webbrowser

def main():
    httpd = http.server.HTTPServer(('127.0.0.1', 0), http.server.SimpleHTTPRequestHandler)
    webbrowser.open_new(f'http://localhost:{httpd.server_port}/')
    httpd.serve_forever()

if __name__ == "__main__":
    main()
```

and it works because `http.server.HTTPServer` uses `socketserver.TCPServer` which synchronously binds the socket in the constructor: https://github.com/python/cpython/blob/3.8/Lib/socketserver.py#L445-L467

gtmanfred: neat, I thought I had tried that previously... hrm. looks like it is in python2.7 too.

graingert: also the threading approach wouldn't fix the problem if HTTPServer did run with `bind_and_activate=False`, consider the following race:

| main                                                                   | Thread-1                                                                                                |
|------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------|
| `thread.start()`                                                       |                                                                                                         |
|                                                                        | `httpd = http.server.HTTPServer(('127.0.0.1', 0), http.server.SimpleHTTPRequestHandler)`                |
|                                                                        | `future.set_result(httpd.server_port)` |
| `webbrowser.open_new(f'http://localhost:{future.result()}/{htmldir}')` |                                                                                                         |
|                                                                        | `httpd.serve_forever()`                                                                                 |

gtmanfred: wouldn't it wait until the future result was set before it ran the webbrowser.open_new()?

Eitherway, i will take a look at fixing it, if interest is given to have it merged into coveragepy.

nedbat:
I have to admit, this seems like something that could easily be well handled outside of coverage.py.

gtmanfred:
That is fine, no worries :) Thanks.

coverage/html.py
graingert: @gtmanfred yes, but look at the table, it waits for the `future.set_result(` but not the `httpd.serve_forever()`

gtmanfred: ahh fair, thanks

PR #1026, opened 2020-09-01
Create Way to force package even if filepath exists

### LINKED ISSUE #268: Create Way to force package even if filepath exists
*Originally reported by* **Donald Stufft (Bitbucket: [dstufft](https://bitbucket.org/dstufft), GitHub: [dstufft](https://github.com/dstufft))**

----------------------------------------

When using tox to run tests the preferred way to run tests creates throw away virtual environments and installs the package to be tested into them. This means that the code to be running is actually inside .tox/venvname/lib/pythonX.Y/site-packages/library. 

Normally coverage.py could support this BUT when doing --source library it sees the library/ directory in the current directory (which isn't being used) and doesn't do the package resolution. This leads to 0% coverage reports always when using tox. It would be great if there was a way to say "No really this is a package".

----------------------------------------
- Bitbucket: https://bitbucket.org/ned/coveragepy/issue/268

graingert:
Fixes: #268

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -99,7 +99,7 @@ class Coverage(object):
     def __init__(
         self, data_file=_DEFAULT_DATAFILE, data_suffix=None, cover_pylib=None,
         auto_data=False, timid=None, branch=None, config_file=True,
-        source=None, omit=None, include=None, debug=None,
+        source=None, source_pkgs=None, omit=None, include=None, debug=None,
         concurrency=None, check_preimported=False, context=None,
     ):
         """
@@ -188,7 +188,7 @@ class Coverage(object):
             config_file=config_file,
             data_file=data_file, cover_pylib=cover_pylib, timid=timid,
             branch=branch, parallel=bool_or_none(data_suffix),
-            source=source, run_omit=omit, run_include=include, debug=debug,
+            source=source, source_pkgs=source_pkgs, run_omit=omit, run_include=include, debug=debug,
             report_omit=omit, report_include=include,
             concurrency=concurrency, context=context,
             )
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -926,6 +926,28 @@ class SourceIncludeOmitTest(IncludeOmitTestsMixin, CoverageTest):
         self.filenames_not_in(lines, "p1b")
         self.assertEqual(lines['p1c'], 0)
 
+    def test_ambigious_source_package_as_dir(self):
+        # pkg1 is a directory and a pkg, since we cd into tests/modules/ambigious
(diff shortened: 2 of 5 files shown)


graingert:
@nedbat it might be cleaner to do something with `[paths]` eg:


```ini
[coverage:run]
source=.

[coverage:paths]
.
.tox/*/lib/*/site-packages/**
```

should be able to get coverage to cover any file in `.tox/*/lib/*/site-packages/**` that matches a file in `.`

nedbat:
> 15:23:42 <graingert> nedbat: I'm getting 0 coverage here https://github.com/PyCQA/modernize/pull/218/commits/814a71907fd5f9032a2519ac1450c8f5f98f31db#diff-380c6a8ebbbce17d55d50ef17d3cf906R60
15:24:17 <graingert> nedbat: it seems to work when I set `include = ./.tox/*/lib/*/site-packages/libmodernize/**` but that doesn't show un-imported files
19:11:16 <graingert> nedbat: https://github.com/PyCQA/modernize/pull/218/commits/a9371366292c62d304115015bd0b9f5f1beed313 this almost does what I want it finds all the files from libmodernize and it associates the files in tox site-packages with the source in the toxinidir
graingert> nedbat: it seems other people have this issue too https://stackoverflow.com/questions/58696476/tox-0-coverage#comment104421991_59072005 https://stackoverflow.com/questions/20292950/making-py-test-coverage-and-tox-work-together-init-py-in-tests-folder#comment112629840_41387398

coverage/control.py
```diff
@@ -100,7 +100,7 @@ def __init__(
         self, data_file=_DEFAULT_DATAFILE, data_suffix=None, cover_pylib=None,
         auto_data=False, timid=None, branch=None, config_file=True,
         source=None, omit=None, include=None, debug=None,
-        concurrency=None, check_preimported=False, context=None,
+        concurrency=None, check_preimported=False, context=None, **kwargs
```
nedbat: There's a clear pattern here already which you should follow.

tests/test_api.py
```diff
@@ -926,6 +926,28 @@ def test_source_package_as_package_part_omitted(self):
         self.filenames_not_in(lines, "p1b")
         self.assertEqual(lines['p1c'], 0)
 
+    def test_ambigious_source_package_as_dir(self):
```
merwok, line 929: `ambiguous` was probably intended here!

graingert: Thanks!

graingert: @merwok fixed here: https://github.com/nedbat/coveragepy/commit/c907b2ee1184dbcea1a3698588ad198a32fbea71#diff-625efbbf35a2290861f596fee1b19838R938

PR #1028, opened 2020-09-02
replace confusing comment with assertion

graingert:


tests/test_api.py
```diff
@@ -887,21 +887,30 @@ def test_source_include_exclusive(self):
...
 
-    def test_source_package_as_package(self):
+    def test_source_package_dotted_sub(self):
         lines = self.coverage_usepkgs(source=["pkg1.sub"])
         self.filenames_not_in(lines, "p2a p2b othera otherb osa osb")
         # Because source= was specified, we do search for unexecuted files.
         self.assertEqual(lines['runmod3'], 0)
 
-    def test_source_package_dotted(self):
+    def test_source_package_dotted_p1b(self):
```
graingert, line 913: @nedbat I'm not really clear what the difference between these two tests is

nedbat: pkg1.sub is a directory, pkg1.p1b is a file.

PR #1030, opened 2020-09-04
if source looks like a package add it to source_pkgs

graingert:
an alternative to https://github.com/nedbat/coveragepy/pull/1026 in which ambiguous `source` definitions use the package and the dir

### DIFF
--- a/coverage/inorout.py
+++ b/coverage/inorout.py
@@ -7,10 +7,12 @@
 import atexit
 import inspect
 import itertools
+import keyword
 import os
 import platform
 import re
 import sys
+import tokenize
 import traceback
 
 from coverage import env
@@ -108,6 +110,14 @@ def module_has_file(mod):
     return os.path.exists(mod__file__)
 
 
+_IS_PKG_RE = re.compile(
+    r"\A(((?!keywords){tokenize.Name})\.?)+\Z".format(
+        tokenize=tokenize,
+        keywords="|".join(re.escape(kw) for kw in keyword.kwlist),
+    )
+)
+
+
 class InOrOut(object):
     """Machinery for determining what files to measure."""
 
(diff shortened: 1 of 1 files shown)


coverage/inorout.py
```diff
@@ -133,10 +143,11 @@ def __init__(self, warn, debug):
     def configure(self, config):
         """Apply the configuration to get ready for decision-time."""
         for src in config.source or []:
+            if _IS_PKG_RE.match(src):
+                self.source_pkgs.append(src)
             if os.path.isdir(src):
                 self.source.append(canonical_filename(src))
-            else:
-                self.source_pkgs.append(src)
+
```
graingert:
```suggestion
```

nedbat:
I don't understand what the re.match is trying to determine. Can you write a few sentences (docs?) about what this change does?

coverage/inorout.py
```diff
@@ -108,6 +110,14 @@ def module_has_file(mod):
     return os.path.exists(mod__file__)
 
 
+_IS_PKG_RE = re.compile(
+    r"\A(((?!keywords){tokenize.Name})\.?)+\Z".format(
```
graingert:
```suggestion
    r"\A(((?!{keywords}){tokenize.Name})\.?)+\Z".format(
```

coverage/inorout.py
```diff
@@ -108,6 +110,14 @@ def module_has_file(mod):
     return os.path.exists(mod__file__)
 
 
+_IS_PKG_RE = re.compile(
```
graingert, line 113: on python3.8 _IS_PKG_RE is
```
re.compile(r"\A(((?!False|None|True|and|as|assert|async|await|break|class|continue|def|del|elif|else|except|finally|for|from|global|if|import|in|is|lambda|nonlocal|not|or|pass|raise|return|try|while|with|yield)\w+)\.?)+\Z")
```

graingert: a bit like:
```
all(s.isidentifier() and not keyword.iskeyword(s) for s in v.split("."))
```

but `s.isidentifier()` is not on python2 and the RE is faster

graingert: @nedbat something like this?

```suggestion
# a dot delimited string of names that are not keywords
_module_name_re = re.compile(
```

nedbat:
This seems like too much guessing to me. And it would change the behavior of `--source=foo` from "assume directory" to "assume package".

graingert:
Yep, makes sense

PR #1051, opened 2020-11-08
Fix the recording of arcs for non-Python files

Changaco:
Hi,

While developing new coverage plugins I found that they received invalid arcs when branch coverage was turned on.

An example of an obviously invalid arc is one which contains a line number greater than the actual number of lines in the file. This can happen when the generated python code being executed is longer than the source code it was generated from.

This problem has two causes:

1. The tracer mixes mapped and unmapped line numbers. In other words, not all line numbers are passed through the tracer's `line_number_range` method before being recorded.
2. The `DataStackEntry` struct isn't cleared before being reused. Not clearing its `last_line` attribute results in bogus arcs being recorded.

My patch fixes both problems.

I don't often write C code, so please let me know if you see something that I could have done better.

### DIFF
--- a/coverage/ctracer/datastack.c
+++ b/coverage/ctracer/datastack.c
@@ -45,6 +45,9 @@ DataStack_grow(Stats *pstats, DataStack *pdata_stack)
 
         pdata_stack->stack = bigger_data_stack;
         pdata_stack->alloc = bigger;
+    } else {
+        /* Zero the entry, because it may have been previously used and can still contain data. */
+        memset(pdata_stack->stack + pdata_stack->depth, 0, sizeof(DataStackEntry));
     }
     return RET_OK;
 }
--- a/coverage/ctracer/datastack.h
+++ b/coverage/ctracer/datastack.h
@@ -22,7 +22,8 @@ typedef struct DataStackEntry {
     PyObject * file_tracer;
 
     /* The line number of the last line recorded, for tracing arcs.
-        -1 means there was no previous line, as when entering a code object.
+       0 means there was no previous line.
+       A negative number -N means we've just entered the code object which starts at line N.
     */
     int last_line;
 
--- a/coverage/ctracer/tracer.c
+++ b/coverage/ctracer/tracer.c
@@ -296,6 +296,53 @@ error:
  * Parts of the trace function.
  */
 
+static void
+CTracer_line_number_range(CTracer *self, PyFrameObject *frame, int lineno, int *lineno_from, int *lineno_to)
+{
+    if (self->pcur_entry->file_tracer == Py_None) {
+        *lineno_from = *lineno_to = lineno;
+        return;
+    }
+
+    PyObject * from_to = NULL;
+    PyObject * pyint = NULL;
+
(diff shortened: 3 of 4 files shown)


nedbat:
Thanks, I don't often get contributions to the C code! Sorry it's taken me so long to respond.  Do you have a way to reproduce the problem this is fixing?

BTW, separately, I'd be interested to hear about the plugin you are working on.

Changaco:
I don't have a simple way of reproducing the problem. I discovered it when running [liberapay](https://github.com/liberapay/liberapay.com)'s test suite with the experimental coverage plugins, and I confirmed that the changes in this branch fix the problem, but I didn't write test code that reproduces it and could be used to prevent regressions in the future.

The plugins I worked on are for Jinja2 and [simplates](http://simplates.org/). I obtained some encouraging results, but the work isn't finished, I paused after I opened this pull request.

Changaco:
I've rebased this branch on master, replacing `frame->f_lineno` with `PyFrame_GetLineNumber(frame)` like in 7e5e28f1aba87c10b96d0ae1244352f4c520aedc.

ProsperousHeart:
since this is so old, updates have happened since, and 3 checks are not being met ... is this something that can be closed?

Changaco:
This patch is still needed. I have now rebased it on master once again, and confirmed that it still fixes the problem.

The patch has become a bit smaller because its previous version modified the `CTracer_check_missing_return` function which was dropped in 8021196662dcadf161cbeaebb3be4b0392b51803.

The patch has also been modified to use `MyFrame_GetCode(frame)` instead of `frame->f_code`, per ab48a7bf66c04754e9964587b2b4f790bb6af8d4.

nedbat:
Can you give me very specific instructions for cloning and running the test suite, and specific pointers to places where the coverage information is wrong?  I took a quick look in the liberapay repo and couldn't find a coverage plugin.

Changaco:
The unfinished coverage plugins are now publicly available in the `coverage-plugins` branch of the `liberapay.com` repository.

To run the tests, a PostgreSQL server is required. By default the tests try to use the `liberapay_tests` database of the local server. Once the database is ready, running the tests with coverage tracking should be as simple as running `make pytest-cov`.

### Results with the unpatched version of `coverage`

1. The pytest output contains quite a few warnings like this:
    ```
    simplate_coverage.py:241: UserWarning: got out of bounds line numbers: [136] > 112
    simplate_coverage.py:323: UserWarning: got out of bounds line numbers for file www/on/%platform/index.spt: [(-126, 87), (136, -62)] > 112
    ```
2. The `htmlcov/index.html` file reports a total of 3748 non-covered ("missing") lines.
3. The HTML coverage report for the `templates/exceptions/EmailAddressError.html` file erroneously claims that line 26 has been run while the surrounding lines in the same code branch haven't been.
    Screenshot_20230331_142441

### Results with the patch

1. There are no `out of bounds` warnings in the pytest output.
2. The `htmlcov/index.html` file reports a total of 3857 non-covered ("missing") lines.
3. The HTML coverage report for the `templates/exceptions/EmailAddressError.html` file correctly shows that line 26 has not been run.
    Screenshot_20230331_142504

PR #1064, opened 2020-11-29
Run lint etc in GitHub Actions

nedbat:


.github/workflows/quality.yml
```diff
@@ -0,0 +1,64 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: "Check out the repo"
+        uses: "actions/checkout@v2"
+
+      - name: "Install Python"
+        uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.8"
```
portante, line 29: Why 3.8 for quality checks but kitting is 3.7?

.github/workflows/quality.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: "Install dependencies"
+        run: |
+          set -xe
+          python -VV
+          python -m site
+          python -m pip install -r requirements/tox.pip
```
portante, line 36: How come you don't use `tox` to do these steps?

.github/workflows/quality.yml
nedbat: ¯\\_(ツ)_/¯

.github/workflows/quality.yml
nedbat: These are needed to get tox installed. Then the rest is handled by tox.

Issue #986, opened 2020-05-10
Chromium does not persist sorting order in HTML report


ptoews:
**Describe the bug**
When sorting the coverage report by a column in Chromium, clicking on a module report and then going back results in the table being sorted by the default order (name) again.

**To Reproduce**
1. What version of coverage.py are you using? The output of `coverage debug sys` is helpful
5.1
1. What commands did you run?
`coverage html -d tests/htmlcov`

Chromium Version: 81.0.4044.122

**Expected behavior**
Sorting order is persisted after navigation

**Additional context**
Does work in Firefox. Maybe related to #611

nedbat:
The sorting is persisted in Chrome.  Do you know what the difference is here between Chrome and Chromium?

ptoews:
No idea, but it might be something else.
Where is the sorting order supposed to be persisted? On the "Local Storage"? Because the Application Tab in the F12 menu doesn't show any values there.

nedbat:
I can't be certain, but I think this is fixed with #1066.  Let me know if the problem persists.

nedbat:
This is now released as part of [coverage 5.3.1](https://pypi.org/project/coverage/5.3.1).

PR #1066, opened 2020-11-30
🐛 Fix HTML report sorting order storage in jQuery unload event

tiangolo:
🐛 Fix HTML report sorting order storage in jQuery unload event.

This is related to: https://github.com/nedbat/coveragepy/issues/986

I'm having a (probably) similar issue. In the generated HTML report, on the console, I'm seeing an error:

```
Uncaught TypeError: $(...).unload is not a function
    at HTMLDocument.coverage.index_ready (coverage_html.js:227)
    at mightThrow (jquery.min.js:2)
    at process (jquery.min.js:2)
```

Using `$().unload` was deprecated and [removed in jQuery 3](https://api.jquery.com/unload/#unload-handler). The recommendation is to use `$().on("unload", handler)`.

I see that it seems the included minified jQuery JS file says it has version 1.11: https://github.com/nedbat/coveragepy/blob/master/coverage/htmlfiles/jquery.min.js#L1

Nevertheless, when I generate an HTML report from scratch in a local project, the minified jQuery I get is indeed 3.x. I have no idea why.

Either way, the method with `$().on("unload", handler)` works with both versions.

For completeness, I'm on Chrome, on Ubuntu.

### DIFF
--- a/coverage/htmlfiles/coverage_html.js
+++ b/coverage/htmlfiles/coverage_html.js
@@ -224,7 +224,7 @@ coverage.index_ready = function ($) {
     coverage.wire_up_filter();
 
     // Watch for page unload events so we can save the final sort settings:
-    $(window).unload(function () {
+    $(window).on("unload", function () {
         try {
             localStorage.setItem(storage_name, sort_list.toString())
         } catch(err) {}


nedbat:
Thanks for the detailed report.  How did you install coverage?  I have a feeling the jQuery version difference is because you installed with an Ubuntu package rather than with `pip install`, is that right?

tiangolo:
Not really, I installed through `pytest-cov`, with Fllit (equivalent to with pip), in a virtual environment.

Thanks for reviewing and merging!

nedbat:
This is now released as part of [coverage 5.3.1](https://pypi.org/project/coverage/5.3.1).

PR #1070, opened 2020-12-04
Coverage action

nedbat:


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,61 @@
...
+          python -m pip install -r requirements/ci.pip
+          python -m pip install -c requirements/pins.pip tox-gh-actions
+
+      - name: "Run tox for ${{ matrix.python-version }}"
+        env:
+          COVERAGE_COVERAGE: "yes"
+        run: |
+          set -xe
+          python -m tox
+          ls -al .*cov*
```
portante: Are these `ls` commands for debugging?

nedbat: Yes :)

nedbat: Codecov is saying my coverage went from 93.97% to 100.00%.  That isn't right.  The codecov commands are magically implicit, so I'm trying to understand what they are doing.

portante: If you do find out how that magic works, we need it in pbench.  =)

nedbat:
@portante This works now.  I use a helper (igor.py), but just "coverage run" and "coverage xml" should produce a coverage.xml file that codecov can upload.

.github/workflows/coverage.yml
nedbat: My Coverage action now works better: it combines runs from different versions of Python and uploads one coverage.xml file.

nedbat:
@portante Happy Birthday! :)  

As a present, I was hoping to show you this GitHub Action that reduces duplication using crazy YAML syntax: https://gist.github.com/474d27c80433993b533a6508fd7e617f

But it turns out that Actions don't support YAML anchors yet? https://github.community/t/support-for-yaml-anchors/16128/45

Which makes me wonder, how can they not support it yet? Do they have their own custom "YAML" parser? Ick.

Anyway, Happy Birthday.

portante:
Thanks, @nedbat!

That looks nice and clean using YAML anchors.  And yes, if GitHub uses their own YAML parser, "Ick" is right.

Do I hear a blog post coming?  ;)

Thanks for proposing a good solution for the wider community to get behind!  Now that is a good present!

PR #1072, opened 2020-12-05
A favicon for the HTML report

nedbat:


portante (approved):
Not that you need approval ... but the changes look good.

PR #1075, opened 2020-12-05
Coverage combining

nedbat:


.github/workflows/coverage.yml
```diff
@@ -48,10 +48,53 @@ jobs:
...
 
-      - name: "Upload to codecov"
-        uses: codecov/codecov-action@v1
+      - name: "Set up Python"
+        uses: "actions/setup-python@v2"
         with:
-          file: coverage.xml
+          python-version: "3.9"
+
+      - name: "Install dependencies"
```
portante, line 76: Is there a way to make a common "action" of sorts?

nedbat: I was wondering about that.  There's a lot of repetition.

Issue #577, opened 2017-05-04
combine aliases does not work on windows base dir paths (ie: "X:\")


nedbat:
*Originally reported by* **vaab (Bitbucket: [vaab](https://bitbucket.org/vaab), GitHub: [vaab](https://github.com/vaab))**

----------------------------------------

I produced some ``.coverage`` file under windows that needs to be combined.

The paths to be aliased is "T:\" (context: I'm using a netshare to automate tests in windows VM)

This doesn't get aliased properly when specifying::

    [paths]
    source = 
        .
        T:\

# Why

because in ``files.py``, in ``PathAliases.add(..)``...

## Problem 1:

 you'll have these 3 lines::

         # The pattern can't end with a wildcard component.
        pattern = pattern.rstrip(r"\/")
        if pattern.endswith("*"):
            raise CoverageException("Pattern must not end with wildcards.")
        pattern_sep = sep(pattern)

Notice that first the pattern given is stripped from it's ending slashes (first instruction), and then (last instruction) the function ``sep(..)`` is called to guess what is the pattern separator... which will not be possible since all slashes have been removed.

Moving the last instruction before the ``.rstrip(..)`` seems the way to go.

## Problem 2

Having removed the last "/" will make this path as a relative one ("T:" is relative, and  "T:/" is not)... thus these following lines will mess badly with the pattern::

        # The pattern is meant to match a filepath.  Let's make it absolute
        # unless it already is, or is meant to match any prefix.
        if not pattern.startswith('*') and not isabs_anywhere(pattern):
            pattern = abs_file(pattern)
        pattern += pattern_sep

A suggested workaround would be to add ``+ pattern_sep`` in the ``isabs_anywhere(..)`` call.

----------------------------------------
- Bitbucket: https://bitbucket.org/ned/coveragepy/issue/577

nedbat:
This is now released as part of [coverage 5.4](https://pypi.org/project/coverage/5.4).

PR #1080, opened 2020-12-14
fix: combine aliases on windows base dirs (ie: ``X:\``) (fixes: #577)

vaab:
Illustrating a valid fix for my issue, that is still happening ;-)

I had a (very quick) look at running the tests, but unfortunately I had a lot of tests failing on the pristine source prior to any change. So I figured I missed a lot of dependencies or information to get this running properly. I'm afraid that I couldn't check that I didn't break anything properly before submitting that code change.

If you think this code change is definitively worth it, I'll take the time to add tests to your convenience if required.

### DIFF
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -359,17 +359,19 @@ class PathAliases(object):
         match an entire tree, and not just its root.
 
         """
+        pattern_sep = sep(pattern)
+
         if len(pattern) > 1:
             pattern = pattern.rstrip(r"\/")
 
         # The pattern can't end with a wildcard component.
         if pattern.endswith("*"):
             raise CoverageException("Pattern must not end with wildcards.")
-        pattern_sep = sep(pattern)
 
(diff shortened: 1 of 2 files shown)


nedbat:
@vaab Thanks for sticking with this!  I took a quick look at the change, and it seems straightforward enough.  The tests are passing here on GitHub, so it seems good.  If you could add a test of the specific problem, that would be great!

vaab:
Added the test as required, kept it quite simple. Thanks for your feedback. Should I worry about the coverage drop ? It seems to happen in a test file, which was seemingly untouched by my change.

nedbat:
@vaab no worries about the codecov drop.  Ironically, I can't make heads or tails of those reports, and have turned them off.

Thanks!

nedbat:
This is now released as part of [coverage 5.4](https://pypi.org/project/coverage/5.4).

PR #1087, opened 2021-01-04
Add --always-total option

### LINKED ISSUE #1086: Add option to include TOTAL line even if there's only one file
Currently the total line is skipped when there's only one file whose coverage was computed. This is annoying when using a regex to report on coverage from a CI/CD build log (e.g. in Gitlab) because one can't use the recommended regex.

Either always print the TOTAL line, or add an option to force it even when there's only one file.

lordjabez:
Closes #1086

### DIFF
--- a/coverage/summary.py
+++ b/coverage/summary.py
@@ -120,8 +120,9 @@ class SummaryReporter(object):
         for line in lines:
             self.writeout(line[0])
 
-        # Write a TOTAl line if we had more than one file.
-        if self.total.n_files > 1:
+        # Write a TOTAl line if we had more than one file,
+        # or if configured to always total
+        if self.total.n_files > 1 or self.config.always_total:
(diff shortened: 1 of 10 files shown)


lordjabez:
@nedbat Let me know what you think. If you like it, I'll add a test or two.

coverage/summary.py
```diff
@@ -120,8 +120,9 @@ def report(self, morfs, outfile=None):
         for line in lines:
             self.writeout(line[0])
 
-        # Write a TOTAl line if we had more than one file.
-        if self.total.n_files > 1:
+        # Write a TOTAl line if we had more than one file,
```
portante, line 123: s/TOTAl/TOTAL/

lordjabez:
Per request on the #1086 going to go with a simpler path and always output the `TOTAL` line no matter what.

PR #1088, opened 2021-01-04
Prevent misbehaving dynamic context plugins from ruining coverage

s0undt3ch:
While working on a dynamic context plugin I encountered an issue where contexts are not unique(I think it's related to subprocess coverage, but I haven't been able to create a small enough example to show here).

While we can argue that the plugin itself should be fixed, this looks like an innocuous change to `coveragepy` which can prevent these misbehaving plugins from ruining most of the coverage data(not being covered).

### DIFF
--- a/coverage/sqldata.py
+++ b/coverage/sqldata.py
@@ -70,6 +70,7 @@ CREATE TABLE context (
     id integer primary key,
     context text,
     unique (context)
+    ON CONFLICT IGNORE
 );
 
 CREATE TABLE line_bits (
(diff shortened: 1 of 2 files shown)


nedbat:
Do you have any more information about what caused the conflict?  This is the type of change that can hide useful information about a problem to be fixed.  I see that it would be safe to add, but I have qualms.

s0undt3ch:
Not really, no. The information I currently know is what's mentioned on the description, sub-processes coverage.
What the misbehaving plugin(mine) is doing, can be seen [here](https://github.com/s0undt3ch/pytest-coverage-contexts/blob/master/coveragectx/pytest/plugin.py#L23) and [here](https://github.com/s0undt3ch/pytest-coverage-contexts/blob/master/coveragectx/coverage/plugin.py#L30).
Does it help understand the problem?

ProsperousHeart:
This branch is out of date, and submission is 2+ years ago. Can it be closed?

PR #1089, opened 2021-01-04
Carry static context information into subprocesses

### LINKED ISSUE #1079: CLI --context not used when multiprocessing is enabled(for child processes)
**Describe the bug**
When running coverage and passing `--context=foo`, only the main process uses that `foo` static context.
Any sub-processes, which do get coverage because we're setting `COVERAGE_PROCESS_START`, will not use that `foo` context, instead, they report an `empty` context.

**To Reproduce**
How can we reproduce the problem?  Please *be specific*. Don't just link to a failing CI job. Answer the questions below:
It's independent of the python version used.
1. What version of coverage.py are you using? The output of `coverage debug sys` is helpful.
5.3
A lot, and I don't think they're related to this issue, but I can add that to the issue if strictly necessary.
1. What code are you running?  Give us a specific commit of a specific repo that we can check out.
https://github.com/s0undt3ch/salt/tree/features/coverage-contexts
1. What commands did you run?
```
nox -e 'pytest-zeromq-3(coverage=True)' -- -vv --run-destructive --run-slow --ssh-tests --log-cli-level=debug --showlocals tests/integration/modules/test_config.py --maxfail=1
```

**Expected behavior**
I would expect that the context passed through `--context`, to be "inherited" but any subprocess which has coverage enabled.

s0undt3ch:
Fixes #1079

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -434,7 +434,10 @@ class Coverage(object):
                 raise CoverageException(                    # pragma: only jython
                     "multiprocessing is not supported on this Python"
                 )
-            patch_multiprocessing(rcfile=self.config.config_file)
+            patch_multiprocessing(
+                rcfile=self.config.config_file,
+                static_context=self.config.context or ""
+            )
 
         dycon = self.config.dynamic_context
         if not dycon or dycon == "none":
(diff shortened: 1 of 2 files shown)


s0undt3ch:
Please provide direction if this is the wrong one.

ProsperousHeart:
Is this still valid?

PR #1091, opened 2021-01-05
Always output TOTAL line.

lordjabez:
Always output the TOTAL line, and update tests to match new functionality.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,8 +24,9 @@ want to know what's different in 5.0 since 4.5.x, see :ref:`whatsnew5x`.
 Unreleased
 ----------
 
-Nothing yet.
-
+- The text report produced by ``coverage report`` now always outputs a TOTAL
+  line, even if only one Python file is reported.  This makes regex parsing
+  of the output easier.  Thanks, Judson Neer.
 
 .. _changes_531:
 
--- a/coverage/summary.py
+++ b/coverage/summary.py
@@ -120,8 +120,8 @@ class SummaryReporter(object):
         for line in lines:
             self.writeout(line[0])
 
-        # Write a TOTAl line if we had more than one file.
-        if self.total.n_files > 1:
+        # Write a TOTAL line if we had at least one file.
+        if self.total.n_files > 0:
             self.writeout(rule)
             args = ("TOTAL", self.total.n_statements, self.total.n_missing)
             if self.branches:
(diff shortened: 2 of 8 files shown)


coverage/summary.py
```diff
@@ -120,8 +120,8 @@ def report(self, morfs, outfile=None):
         for line in lines:
             self.writeout(line[0])
 
-        # Write a TOTAl line if we had more than one file.
-        if self.total.n_files > 1:
+        # Write a TOTAl line if we had at least one file.
```
portante: s/TOTAl/TOTAL/

portante (approved):
FWIW ...

coverage/summary.py
lordjabez: Ugh, I literally made that mistake a second time didn't i?

portante (approved):

CHANGES.rst
```diff
@@ -24,7 +24,7 @@ want to know what's different in 5.0 since 4.5.x, see :ref:`whatsnew5x`.
 Unreleased
 ----------
 
-Nothing yet.
+- Always output TOTAL line.
```
nedbat:
```suggestion
- The text report produced by ``coverage report`` now always outputs a TOTAL line, even if only one Python file is reported.  This makes regex parsing of the output easier.  Thanks, Judson Neer.
```

tests/test_process.py
```diff
@@ -1348,7 +1352,9 @@ def test_accented_directory(self):
         report_expected = (
             u"Name            Stmts   Miss  Cover\n"
             u"-----------------------------------\n"
-            u"\xe2%saccented.py       1      0   100%%\n" % os.sep
+            u"\xe2%saccented.py       1      0   100%%\n"
+            u"-----------------------------------\n"
+            u"TOTAL               1      0   100%%\n" % os.sep
```
nedbat: You don't need the string formatting and os.sep on this line.

nedbat:
This is great, thanks.  A little more detail in the changelog, and remove an unneeded string formatting, and we are ready to merge!

tests/test_process.py
lordjabez: If I don't move the `% os.sep` piece to the last line (or after the trailing parenthesis) the code fails to execute with a syntax error. And the substitution has to be _somewhere_

I'll tweak this a bit and hopefully it's more obvious what's going on.

nedbat:
@lordjabez Thanks. I see now that I misread the os.sep string formatting, thanks for putting up with me :)  I will merge this when the tests pass.

Issue #1090, opened 2021-01-04
Allow overriding skip_covered under the html section


gaborbernat:
For a terminal report to not show fully covered lines, one can set skip_covered to True. For HTML reports though you do want to report on all files, so you can see what covers what. Would be good to be able to specify this configuration scenario via the config only, perhaps by allowing to override ``coverage:report`` ``skip_covered`` under ``coverage:html``. Currently, you can only override via the CLI, but for example in ``pytest-cov`` these flags are not exposed to alter. 

PS. pytest-cov ticket solution for this https://github.com/pytest-dev/pytest-cov/issues/449

ryboe:
I'd like to be able to write this in `pyproject.toml`.
```toml
[tool.coverage.html]
skip_covered = true
```

Thank you! 🙏

nedbat:
This is implemented in https://github.com/nedbat/coveragepy/pull/1099.  Does it do what you need?

ryboe:
Yes it does! Thank you! That was so fast! 🙏

nedbat:
This is now released as part of [coverage 5.4](https://pypi.org/project/coverage/5.4).

PR #1099, opened 2021-01-11
skip_covered and skip_empty for HTML. #1090

nedbat:
The ``skip_covered`` and ``skip_empty`` settings in the configuration file
  can now be specified in the ``[html]`` section, so that text reports and HTML
  reports can use separate settings.  The HTML report will still use the
  ``[report]`` settings if there isn't a value in the ``[html]`` section.


gaborbernat:
Looking forward to this release 🙇🏻

PR #1104, opened 2021-01-19
Support regex patterns in module names

kvalev:


nedbat:
Hi, thanks.  Could you provide some explanation of what problem this is solving?  I thought there was an issue written about this, but I can't find it now.

kvalev:
Hi, I had another PR, but decided to create a new one, as the other one was getting quite stale.

The PR will be helpful for us, as we are using one framework (airflow), which generates code in temporary modules on the fly. So in case we want to measure the code coverage of the generated modules (we do), we need a way to match the dynamic module names and this is where the regex support comes into play.

ProsperousHeart:
@kvalev  - what was the original PR? opening a new one just because the old one is stale makes it difficult for maintainers.

Is this request still valid at all?

kvalev:
@ProsperousHeart Here is the original PR - https://github.com/nedbat/coveragepy/pull/917 and yes, the request is still valid

PR #1114, opened 2021-02-05
add command line option to show current configuration

smcavallo:
It's useful to be able to determine in advance of running an analysis what the current configuration is.
This adds support for exposing the end resulting configuration via cmd
ex.
`coverage config --show-omit -o json`

Signed-off-by: smcavallo <smcavallo@hotmail.com>

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -114,6 +114,10 @@ class Opts(object):
         '--show-contexts', action='store_true',
         help="Show contexts for covered lines.",
     )
+    show_omit = optparse.make_option(
+        '--show-omit', action='store_true',
+        help="Show omit entries in coverage config files.",
+    )
     omit = optparse.make_option(
         '', '--omit', action='store',
         metavar="PAT1,PAT2,...",
@@ -212,6 +216,7 @@ class CoverageOptionParser(optparse.OptionParser, object):
             append=None,
             branch=None,
             concurrency=None,
+            config=None,
             context=None,
             debug=None,
             directory=None,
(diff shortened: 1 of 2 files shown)


nedbat:
Thanks for the pull request.  I haven't tried running it, but isn't this similar to what `coverage debug config` shows? If so, maybe we need to make it more apparent in the docs?

smcavallo:
> 
> 
> Thanks for the pull request. I haven't tried running it, but isn't this similar to what `coverage debug config` shows? If so, maybe we need to make it more apparent in the docs?

Yes, thank you!!!  I didn't know that existed.  Very similar to what we are looking for.
We'd like to be able to:
* Select a specific section of the config to query.  Ex. `coverage debug config run_omit`
* Have more control over the output format.  Ex. `coverage debug config run_omit -o json`

Our use case is that we are using multiple code coverage tools, coveragepy being the primary and "source of truth."  We need to use a consistent set of rules across all the tools.  We need to be able to query the coveragepy rules  in order to "share" those rules with our other downstream python linters and quality gates.
It would be easier if those other tools had native support for reading coveragepy files but they don't and I'm not sure it would make sense for them to know all the logic of how to do that anyways.
We are looking for a way to get the exclusion rules from coveragepy in a format we can easily parse.
I can add those additional command line options to `coverage debug config` if that's the best route and you would accept that contribution.

nedbat:
I can see the value of parseable output from `coverage debug config`.  I don't think I'd want to add a way to select the section to output though: once you have json output, the filtering can be done downstream.

ProsperousHeart:
Is this still valid?

PR #1153, opened 2021-04-30
Fix documentation code comment formatting

storymode7:
Hey!

I was going through [Excluding code from coverage.py](https://coverage.readthedocs.io/en/latest/excluding.html) and found formatting issues in the comments used in code examples. Fixed those and a couple more from other pages.

I tried to keep it inline with [PEP8](https://www.python.org/dev/peps/pep-0008/#inline-comments)

Depending on the type of change I made and their need, I have separated the changes into 3 commits.
With each commit message explaining the reasoning behind the change.

### DIFF
--- a/doc/branch.rst
+++ b/doc/branch.rst
@@ -108,7 +108,7 @@ tell coverage.py that you don't want them flagged by marking them with a
 pragma::
 
     i = 0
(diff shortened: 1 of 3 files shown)


nedbat:
Thanks, and I appreciate the separate commits, though I'll take all of them (squashed).

PR #1155, opened 2021-05-02
Remove redundant Python 2 code

hugovk:
Also upgrade Python syntax using https://github.com/asottile/pyupgrade and remove some unused imports.


nedbat:
Thanks, I wasn't sure I cared enough about this code to do this, but why not? :)

PR #1164, opened 2021-05-24
fix(python): handle source decoding errors

### LINKED ISSUE #1160: `get_python_source` gets called on shared object file and causes SyntaxError
**Describe the bug**

While testing with `pytest-cov` we started noticing some failures due to an "internal" error. Digging deeper, it looks like somehow a shared object (.so) file has made its way to `get_python_source`, which then fails with

~~~
INTERNALERROR> SyntaxError: invalid or missing encoding declaration
~~~

with the actual issue being

~~~
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/Users/gabriele.tornetta/.pyenv/versions/3.5.10/lib/python3.5/tokenize.py", line 392, in find_cookie
INTERNALERROR>     line_string = line.decode('utf-8')
INTERNALERROR> UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcf in position 0: invalid continuation byte
~~~

Whilst it's maybe the case that the issue is more upstream (maybe .so files should not make it this far?), perhaps `get_python_source` should be enhanced to catch these potential cases. For instance, the extension check at https://github.com/nedbat/coveragepy/blob/c776c901aa9b3214be815ceee9f179f47bcd86d8/coverage/python.py#L42-L45 could be enhanced?

**To Reproduce**

Unfortunately, it's not clear to me how to reproduce this problem, as it has appeared all of a sudden, and I don't see any changes in the dependency versions that are currently being pulled in the CI jobs. The project that is being tested contains some Cythonized code, and this is where the shared object is coming from.

1. What version of Python are you using? **3.5**, **3.6**, **3.8**
1. What version of coverage.py are you using? The output of `coverage debug sys` is helpful. **5.5**
1. What code are you running?  Give us a specific commit of a specific repo that we can check out.
1. What commands did you run?

**Expected behavior**

No internal errors 🙂

P403n1x87:
As described in #1160, occasionally some shared object paths make their way into the coverage report. When they get to `get_python_source` the call to `source_encoding` fails because the content of these files is binary. The change allows `get_python_source` to handle these cases gracefully while still reporting the issue.

Fixes #1160.

### DIFF
--- a/coverage/python.py
+++ b/coverage/python.py
@@ -63,7 +63,10 @@ def get_python_source(filename):
 
     # Replace \f because of http://bugs.python.org/issue19035
     source = source.replace(b'\f', b' ')
-    source = source.decode(source_encoding(source), "replace")
+    try:
+        source = source.decode(source_encoding(source), "replace")
+    except Exception as e:
+        raise NoSource("Cannot decode source '%s: %s" % (filename, e))
 
     # Python code should always end with a line with a newline.
     if source and source[-1] != '\n':


coverage/python.py
```diff
@@ -63,7 +63,10 @@ def get_python_source(filename):
 
     # Replace \f because of http://bugs.python.org/issue19035
     source = source.replace(b'\f', b' ')
-    source = source.decode(source_encoding(source), "replace")
+    try:
+        source = source.decode(source_encoding(source), "replace")
+    except Exception as e:
```
nedbat, line 68: I would much rather catch a more specific error, in a more specific place.  The exception is happening inside source_encoding, so why not put the except-clause in that function?

P403n1x87: This felt like a good place for catching generic decoding errors and not just the one in the linked issue. Of course, I could add a specific exception handling in `source_encoding` if that's preferable.

nedbat: The reason this is a bad place for the except is that `s.decode(..., "replace")` can't raise an exception. In that line of code, it's only the `source_encoding()` call that can fail.

P403n1x87: Oh but this could be fixed by just moving `source_encoding` on its own line and wrapping it with `try...except` 😜 . Anyway, I'll have a closer look as soon as I can.

P403n1x87: Right, I've had another look and the idea of adding checks to `_source_encoding_py2` is not very appealing. Not to mention that they would have to be duplicated in `_source_encoding_py3`.

nedbat: Well, _source_encoding_py2 is gone on master, we've dropped Python 2 support.

P403n1x87: Ah I've just realised that support for Python 3.5 is also being dropped, which is not ideal for our usage. This prompted me to test on Python 3.7 with the master branch of `coveragepy` and the problem does not present itself 🙁 .

P403n1x87:
This has now been fixed with #1252

PR #1196, opened 2021-07-30
open tomli file with newline=''

graingert:


coverage/tomlconfig.py
```diff
@@ -40,7 +40,7 @@ def read(self, filenames):
         filename = os.fspath(filenames)
 
         try:
-            with open(filename, encoding='utf-8') as fp:
+            with open(filename, encoding='utf-8', newline='') as fp:
```
graingert, line 53: @hukkin do you have a test file that breaks without `newline=''`?

hukkin: Any file that has a CR character not immediately followed by a LF character. The most simple example is a one byte file with just the CR character (0x0D).

That is invalid TOML and should raise an error, but doesn't if `newline=''` is not set.

graingert: Thanks, I'll push a test case tomorrow

ProsperousHeart:
I know this is a draft, but is it still valid?

hukkin:
Yes, this is still valid.

nedbat:
@hukkin can you show me a toml file that causes a problem?  Use `hexdump -C` to show the actual bytes.  I tried a simple file with only `\r` line endings in Python 3.9, and it was fine:
```
% hexdump -C pyproject.toml
00000000  5b 74 6f 6f 6c 2e 63 6f  76 65 72 61 67 65 2e 72  |[tool.coverage.r|
00000010  75 6e 5d 0d 62 72 61 6e  63 68 3d 74 72 75 65 0d  |un].branch=true.|
00000020
```

hukkin:
> can you show me a toml file that causes a problem?

I don't think it's likely this will ever really cause a problem. But the "problem" is that the file you generated was parsed fine even though it is invalid TOML. A `TOMLDecodeError` should be raised by tomli/tomllib instead. That can't happen if a CR is translated to LF (which is what `open` does by default).

nedbat:
Thanks for the explanation.  I don't see the need to enforce this kind of strictness, since no one will have a TOML file like this.

PR #1201, opened 2021-08-02
Fix parsing 'command_line' option

paveltsialnou:
Coverage config files supports 'command_line' parameters. However, only '-m' is parsed properly.

The line 'command_line = "--module unittest discover"' is considered as a path, not a module option.

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -709,7 +709,7 @@ class CoverageScript:
             command_line = self.coverage.get_option("run:command_line")
             if command_line is not None:
                 args = shlex.split(command_line)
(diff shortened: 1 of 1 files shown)


nedbat:
Python doesn't accept `--module` as a command-line option, so coverage.py shouldn't either.  Where did you get `--module` from?

paveltsialnou:
From `coverage run —help`. This is used to set `module` option to `True` and it is then cut off from passed args. So it is not related to `python -m`.

To test use the following `.coveragerc` lines:
```
[run]
command_line = —module unittest discover
```

PR #1210, opened 2021-08-06
Fix an incompatibility with pyarmor

glacials:
Hi thanks for all your amazing and dedicated work on Coverage.py! I ran across a compatibility issue and wanted to propose a fix, feedback welcome :) 

I was running into an error that I traced to frames whose first argument is called `self` according to `frame.f_code.co_varnames`, but which have no `frame.f_locals["self"]`:

```
...
../../.pyenv/versions/3.9.6/lib/python3.9/site-packages/coverage/context.py:41: in should_start_context_test_function
    return qualname_from_frame(frame)
../../.pyenv/versions/3.9.6/lib/python3.9/site-packages/coverage/context.py:51: in qualname_from_frame
    self = frame.f_locals["self"]
E   KeyError: 'self'
```

Specifically, I can reproduce frames like this out of [pyarmor][1]ed code -- based on light research it seems there's nothing theoretically wrong with a frame like this, they just generally don't exist in practice.

These frames only trip up Coverage.py if  `dynamic_context = test_function` ("who tests what") is set, and if a function that starts with "test" and takes `self` as a first argument gets called.

I ran into this issue refactoring our pytest plugin that uses a class to sort tests, so was doing things like

```python
for test in sorter.tests_in_bucket(1):
```

inside pytest hooks, which set off this chain reaction 😄 

I've set up a simplistic reproduction example here: https://github.com/glacials/_coverage_error_repro

[1]: https://github.com/dashingsoft/pyarmor

### DIFF
--- a/coverage/context.py
+++ b/coverage/context.py
@@ -48,7 +48,7 @@ def qualname_from_frame(frame):
     fname = co.co_name
     method = None
     if co.co_argcount and co.co_varnames[0] == "self":
-        self = frame.f_locals["self"]
+        self = frame.f_locals.get("self", None)
         method = getattr(self, fname, None)
 
     if method is None:
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -5,6 +5,7 @@
 
 import inspect
 import os.path
+from unittest import mock
 
 import coverage
 from coverage.context import qualname_from_frame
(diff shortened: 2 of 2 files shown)


nedbat:
This looks straightforward enough, though I am curious how the function has a self argument, but then doesn't have a local called self?

glacials:
I don't know the reason specifically, but I'm sure it's something to do with how pyarmor obfuscates code. Maybe @jondy has some insight?

nedbat:
This is now released as part of [coverage 6.0.1](https://pypi.org/project/coverage/6.0.1).

jondy:
@glacials if the scripts are obfuscated by pyarmor with super mode, `frame.f_locals` is protected and can't be visited out of obfuscated scripts. `frame.f_locals` is an empty dict in the outer scripts.

PR #1217, opened 2021-08-30
Add link to coverage.py twitter account to project urls

asherf:
pypi has supported twitter links [for a while now](https://github.com/pypa/warehouse/pull/878/files#diff-057bf4a747c150d9f3f76db209a6c8f689b023a8e7cd41d96fcc549492889da7R31) so it makes sense to link to coverage.py official twitter account from the pypi project page, similar to other projects, like [pytest](https://pypi.org/project/pytest/)
Screen Shot 2021-08-30 at 2 11 36 PM


nedbat:
Thanks!

PR #1238, opened 2021-10-12
fix: find_writable_pth_directory must expect failure from open()

mgorny:
Fix find_writable_pth_directory() to expect the OSError from the open()
call.  In general, this is what's going to happen if the file is not
writable.  The current logic could only be triggered e.g. if the file
was writable but the underlying filesystem run out of space.


nedbat:
Thanks!

PR #1248, opened 2021-10-17
Remove jQuery dependency

### LINKED ISSUE #1118: Drop jQuery dependency and legacy browsers
It came to my attention that `coveragepy` requires jQuery. As most of the features jQuery provides are now supported by all modern (i.e. evergreen) browsers I think it would be nice to get rid of jQuery.

Furthermore does the way how this tools is shipped in debian only list jQuery as recommended leading to a degraded user experience (error messages when generating HTML reports).

I would be willing to perform this transformation if you agree with my points.

septatrix:
This is what I got so far trying to get rid of jQuery.
Everything should be working on modern browsers (I think IE11 too but I would have to double check that).

I have not adjusted the tests so far but currently also not really the time.

Fixes #1118.


nedbat:
Wow, thanks! I added a sticky header a few days ago, which is causing the conflicts.  I should be able to work on this over the weekend.

septatrix:
No problem. I actually learned some new things that javascript can do (like getting the element at a pixel - and that even in very old browsers) :D I hope everything is understandable, otherwise feel free to ask.

nedbat:
I've merged, tweaked, and squashed this at commit 91ace30b.  Thanks for your amazing work! I hope it's ok how I've credited you: https://github.com/nedbat/coveragepy/commit/b5a53740700e59dbebbf655d4914bbaf9470870a

septatrix:
Thank you for resolving the conflicts and the credit is fine, thanks.

nedbat:
This is now released as part of [coverage 6.1](https://pypi.org/project/coverage/6.1).

PR #1260, opened 2021-10-24
Build wheel for PyPy3.8

hugovk:


nedbat:
Hmm, I've merged this, but now I wonder if it's needed?  Can't PyPy use the source wheels?

hugovk:
Yes it can, for example:

https://github.com/hugovk/test/runs/3990236369?check_suite_focus=true#step:4:13

So not urgent to release, but it's good to also provide wheels, like the pp36 and pp37 ones.

Issue #1189, opened 2021-07-13
"coverage html" fails on systems with FIPS enabled


cjbland:
**Is your feature request related to a problem? Please describe.**
My development system recently got an "upgrade" which enabled FIPS and has thus blocked all access to creating MD5 hashes. Because of this, `coverage html` does not working anymore. Instead I get a stacktrace that looks like this (I can't copy and paste the exact error because it's on a lab machine that is not connected to the internet):

```
  File: "....../venv/lib64/python3.6/site-packages/coverage/html.py" line 461, in check_global_data
    m = Hasher()
  File: "....../venv/lib64/python3.6/site-packages/coverage/misc.py" line 203, in __init__
    self.md5 = hashlib.md5()
ValueError: unknown reasons
```

**Describe the solution you'd like**
It would be nice if this class used `sha256` instead of `md5` for the hashing algorithm.

**Describe alternatives you've considered**
We have not explored any other alternatives other than manually modifying the code to work.

**Additional context**
The workstation I'm using is configured as follows:

CentOS 7.9
Python 3.6
coverage==5.5.0
pytest==6.2.4
pytest-cov==2.12.1

I will create a PR for this as the changes seemed pretty straightforward and non-impacting.

nedbat:
Fixed in 367f7c45.

nedbat:
This is now released as part of [coverage 6.0b1](https://pypi.org/project/coverage/6.0b1).

nedbat:
This is now released as part of [coverage 6.0](https://pypi.org/project/coverage/6.0).

PR #1289, opened 2021-12-08
Lcov testing

bradb423:
Hi @nedbat. As a follow up to our messages last week, we have produced an LCOV report generator, taking inspiration from @chaychoong's [cli converter](https://github.com/chaychoong/coveragepy-lcov).
The LCOV generator is run similarly to the HTML or JSON generators using the command `coverage lcov`, and produces an LCOV file (defaulting to `coverage.lcov`).

As of right now, the LCOV reporter does not support branch coverage, however this feature has the potential to be added in the future.

The generator also provides checksums for each line in the `coverage.lcov` file, as outlined [here](http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php)

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -127,6 +127,11 @@ class Opts:
         '', '--pretty-print', action='store_true',
         help="Format the JSON for human readers.",
     )
+    lcov = optparse.make_option(
+        '-o', '', action='store', dest='outfile',
+        metavar="OUTFILE",
+        help="Write the LCOV report to this file. Defaults to 'coverage.lcov'"
+    )
     parallel_mode = optparse.make_option(
         '-p', '--parallel-mode', action='store_true',
         help=(
@@ -473,6 +478,20 @@ CMDS = {
         usage="[options] [modules]",
         description="Generate an XML report of coverage results."
     ),
+
+    'lcov': CmdOptionParser(
+        "lcov",
+        [
+            Opts.fail_under,
+            Opts.ignore_errors,
+            Opts.include,
+            Opts.lcov,
+            Opts.omit,
+            Opts.quiet,
+        ] + GLOBAL_ARGS,
+        usage="[options] [modules]",
+        description="Generate an LCOV report of coverage results."
+    )
 }
 
 
@@ -657,6 +676,12 @@ class CoverageScript:
                 show_contexts=options.show_contexts,
                 **report_args
                 )
+        elif options.action == "lcov":
+            total = self.coverage.lcov_report(
+                outfile=options.outfile,
+                **report_args
+                )
+
         else:
             # There are no other possible actions.
             raise AssertionError
@@ -854,6 +879,7 @@ HELP_TOPICS = {
             report      Report coverage stats on modules.
             run         Run a Python program and measure code execution.
             xml         Create an XML report of coverage results.
+            lcov        Create an LCOV report of coverage results.
 
         Use "{program_name} help <command>" for detailed help on any command.
     """,
--- a/coverage/config.py
+++ b/coverage/config.py
@@ -227,6 +227,9 @@ class CoverageConfig:
         self.json_pretty_print = False
         self.json_show_contexts = False
 
+        # Default output filename for lcov_reporter
+        self.lcov_output = "coverage.lcov"
+
         # Defaults for [paths]
         self.paths = collections.OrderedDict()
 
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -26,6 +26,7 @@ from coverage.files import PathAliases, abs_file, relative_filename, set_relativ
 from coverage.html import HtmlReporter
 from coverage.inorout import InOrOut
 from coverage.jsonreport import JsonReporter
+from coverage.lcovreport import LcovReporter
 from coverage.misc import bool_or_none, join_regex, human_sorted, human_sorted_items
 from coverage.misc import DefaultValue, ensure_dir_for_file, isolate_module
 from coverage.plugin import FileReporter
@@ -1048,6 +1049,25 @@ class Coverage:
             json_show_contexts=show_contexts
         ):
             return render_report(self.config.json_output, JsonReporter(self), morfs, self._message)
+
+    def lcov_report(
+        self, morfs=None, outfile=None, ignore_errors=None,
+        omit=None, include=None, contexts=None,
+    ):
+        """Generate an LCOV report of coverage results.
+
+        Each module in 'morfs' is included in the report. 'outfile' is the
+        path to write the file to, "-" will write to stdout.
+
+        See :meth 'report' for other arguments.
+
+        .. versionadded:: 6.3
+        """
+        with override_config(self,
+            ignore_errors=ignore_errors, report_omit=omit, report_include=include,
+            lcov_output=outfile,  report_contexts=contexts,
+        ):
+            return render_report(self.config.lcov_output, LcovReporter(self), morfs, self._message)
 
     def sys_info(self):
         """Return a list of (key, value) pairs showing internal information."""
--- a/coverage/lcovreport.py
+++ b/coverage/lcovreport.py
@@ -1,0 +1,106 @@
+# Licensed under the Apache License: http://www.apache.org/licenses/LICENSE-2.0
+# For details: https://github.com/nedbat/coveragepy/blob/master/NOTICE.txt
+
+"""LCOV reporting for coverage.py."""
+
+import sys
+import base64
+from hashlib import md5
+
+from coverage.report import get_analysis_to_report
+
+
+class LcovReporter:
+    """A reporter for writing LCOV coverage reports."""
+
+    report_type = "LCOV report"
+
+    def __init__(self, coverage):
+        self.coverage = coverage
+        self.config = self.coverage.config
+
+    def report(self, morfs, outfile=None):
+        """Renders the full lcov report
+
+        'morfs' is a list of modules or filenames
+
+        outfile is the file object to write the file into.
+        """
+
+        self.coverage.get_data()
+        outfile = outfile or sys.stdout
+
+        for fr, analysis in get_analysis_to_report(self.coverage, morfs):
+            self.get_lcov(fr, analysis, outfile)
+
+    def get_lcov(self, fr, analysis, outfile=None):
+        """Produces the lcov data for a single file
+
+        get_lcov currently supports both line and branch coverage,
+        however function coverage is not supported.
+
+        """
+
+        outfile.write("TN:\n")
+        outfile.write(f"SF:{fr.relative_filename()}\n")
+        source_lines = fr.source().splitlines()
+        for covered in sorted(analysis.executed):
+            # Note: Coveragepy currently only supports checking *if* a line has
+            # been executed, not how many times, so we set this to 1 for nice
+            # output even if it's technically incorrect
+
+            # The lines below calculate a 64 bit encoded md5 hash of the line
+            # corresponding to the DA lines in the lcov file,
+            # for either case of the line being covered or missed in Coveragepy
+            # The final two characters of the encoding ("==") are removed from
+            # the hash to allow genhtml to run on the resulting lcov file
+            if source_lines:
+                line = source_lines[covered - 1].encode("utf-8")
+            else:
+                line = b""
+            hashed = str(base64.b64encode(md5(line).digest())[:-2], encoding="utf-8")
+            outfile.write(f"DA:{covered},1,{hashed}\n")
+        for missed in sorted(analysis.missing):
+            if source_lines:
+                line = source_lines[missed - 1].encode("utf-8")
+            else:
+                line = b""
+            hashed = str(base64.b64encode(md5(line).digest())[:-2], encoding="utf-8")
+            outfile.write(f"DA:{missed},0,{hashed}\n")
+        outfile.write(f"LF:{len(analysis.statements)}\n")
+        outfile.write(f"LH:{len(analysis.executed)}\n")
+
+        # More information dense branch coverage data
+        missing_arcs = analysis.missing_branch_arcs()
+        executed_arcs = analysis.executed_branch_arcs()
+        for block_number, block_line_number in enumerate(
+            sorted(analysis.branch_stats().keys())
+        ):
+            for branch_number, line_number in enumerate(
+                sorted(missing_arcs[block_line_number])
+            ):
+                # The exit branches have a negative line number,
+                # this will not produce valid lcov, and so setting
+                # the line number of the exit branch to 0 will allow
+                # for valid lcov, while preserving the data
+                line_number = max(line_number, 0)
+                outfile.write(f"BRDA:{line_number},{block_number},{branch_number},-\n")
+            # The start value below allows for the block number to be
+            # preserved between these two for loops (stopping the loop from
+            # resetting the value of the block number to 0)
+            for branch_number, line_number in enumerate(
+                sorted(executed_arcs[block_line_number]),
+                start=len(missing_arcs[block_line_number]),
+            ):
+                line_number = max(line_number, 0)
+                outfile.write(f"BRDA:{line_number},{block_number},{branch_number},1\n")
+
+        # Summary of the branch coverage
+        if analysis.has_arcs():
+            branch_stats = analysis.branch_stats()
+            brf = sum(t for t, k in branch_stats.values())
+            brh = brf - sum(t - k for t, k in branch_stats.values())
+            outfile.write(f"BRF:{brf}\n")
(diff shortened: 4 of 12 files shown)


tests/test_lcov.py
```diff
@@ -0,0 +1,148 @@
...
+            "DA:10,0,jodMK26WYDizOO1C7ekBbg\n",
+            "DA:11,0,LtxfKehkX8o4KvC5GnN52g\n",
+            "LF:8\n",
+            "LH:4\n",
+            "end_of_record\n",
+        ]
+        actual_result = self.get_lcov_report_content()
+        assert expected_result == actual_result
+
+    # TODO: Add tests for branch coverage here for when this
```
TLATER: Hmm, I thought the issue was that you didn't think this was possible with coveragepy because none of the other examples tried it?

It's worth noting that lcov [has two different kinds of branch coverage reporting](https://github.com/linux-test-project/lcov/blob/master/man/geninfo.1#L552):

1. Per-line "is this branch covered"
2. Summary of branch coverage

Having had a quick glance, the latter shouldn't be all that hard to do, see https://github.com/nedbat/coveragepy/blob/1c29ef3e4b871508bb2defd6b47b9a745547b626/coverage/results.py#L140 for getting the data.

I think you'd simply need to add something like:

```python
brf, brh = analysis.branch_stats()
outfile.write(f"BRF:{brf}\n")
outfile.write(f"BRH:{brh}\n")
```

The former can be done with [`arcs`](https://github.com/nedbat/coveragepy/blob/1c29ef3e4b871508bb2defd6b47b9a745547b626/coverage/results.py#L79), I believe, but I don't know enough about what those represent to actually recommend anything here without playing a bit with it myself.

I'm guessing arcs don't line up perfectly with lines (e.g. `something if cond else something_else`), and that this therefore requires a bit of thought, but this shouldn't be a complete show-stopper, especially since C also has the option of using blocks in lines around branches?

coverage/lcovreport.py
```diff
@@ -0,0 +1,61 @@
...
+
+        for fr, analysis in get_analysis_to_report(self.coverage, morfs):
+            self.get_lcov(fr, analysis, outfile)
+
+    def get_lcov(self, fr, analysis, outfile=None):
+        """Produces the lcov data for a single file"""
+
+        outfile.write("TN:\n")
+        outfile.write(f"SF:{fr.relative_filename()}\n")
+        sourceLines = fr.source().splitlines()
```
TLATER: Besides branch coverage, [function name and execution info](https://github.com/linux-test-project/lcov/blob/master/man/geninfo.1#L538) is also missing. This is totally valid lcov, but it's probably worth pointing out in a comment here at least.

I don't think there's a good way of implementing that with coveragepy currently?

bradb423:
I have added the branch coverage support for the LCOV generator. Both the summary, and the line by line branch coverage were possible in the end, and appropriate tests for branch coverage have now been added. 

There is still the problem of pypy3 failing currently, this seems to be due to the ordering of the lines in the LCOV file, a further patch will be added to rectify this.

bradb423:
I have now fixed the pypy3 issue from earlier.

nedbat:
I really appreciate the work you are putting into this.  I'm holding off on specific review while I continue to think through where this code should live (inside coverage.py or outside, roughly).

bradb423:
Thank you for your time, @nedbat. I'd be more than happy to answer any questions that you may have, along with any review comments that you may have in the future.

From what I can tell, there seems to be a bug in the test for pypy3 on MacOS. It looks like the pypy3 cannot be found for the run, as you can see [here](https://github.com/nedbat/coveragepy/runs/4519927975?check_suite_focus=true), do you know what might be the cause of this?

nedbat:
@bradb423 I fixed the pypy3 thing on master. If you rebase, you'll get the fix.

bradb423:
@nedbat thank you!

bradb423:
Hi @nedbat , thank you for your patience and for your help over the last couple of weeks. I encountered a bug with the code this morning (causing the block numbers to provide the wrong values) which has now been fixed.

I believe that as of right now, this PR is in a state at which it does not require any more commits. Please let me know your thoughts, and your questions if you have any.

coverage/cmdline.py
```diff
@@ -127,6 +127,11 @@ class Opts:
         '', '--pretty-print', action='store_true',
         help="Format the JSON for human readers.",
     )
+    lcov = optparse.make_option(
+        '-l', '--lcov', action='store', dest='outfile',
```
nedbat: This should be `-o --output` I think.  There's an earlier PR that used `-l`, but that was hacking it into an existing command.

coverage/cmdline.py
```diff
@@ -473,6 +478,20 @@ def get_prog_name(self):
...
+
+    'lcov': CmdOptionParser(
+        "lcov",
+        [
+            Opts.fail_under,
+            Opts.ignore_errors,
+            Opts.include,
+            Opts.omit,
+            Opts.quiet,
+            Opts.lcov,
```
nedbat: I've tried to keep the options alphabetized (roughly) in the lists.

coverage/cmdline.py
```diff
@@ -658,6 +677,12 @@ def command_line(self, argv):
                 show_contexts=options.show_contexts,
                 **report_args
             )
+        elif options.action == "lcov":
+            outfile = options.outfile
```
nedbat: I thought this pre-assignment was odd, but I see now that the other reporting clauses do it like that.  I'm changing them :)

coverage/config.py
```diff
@@ -227,6 +227,9 @@ def __init__(self):
         self.json_pretty_print = False
         self.json_show_contexts = False
 
+        # Defaults for [lcov]
```
nedbat: This comment is a little misleading, since there is no parsing of an `[lcov]` section in the config file.  Do you want to support setting the lcov output file in the config file?

coverage/control.py
```diff
@@ -1049,6 +1050,23 @@ def json_report(
...
+    def lcov_report(
+        self, morfs=None, outfile=None, ignore_errors=None,
+        omit=None, include=None, contexts=None,
+    ):
+        """Generate an LCOV report of coverage results.
+
+        Each module in 'morfs' is included in the report. 'outfile' is the
+        path to write the file to, "-" will write to stdout.
+
+        See :meth 'report' for other arguments.
```
nedbat, line 1062: This docstring should have:
```
.. versionadded:: 6.3
```

coverage/lcovreport.py
```diff
@@ -0,0 +1,106 @@
...
+
+import sys
+import base64
+from hashlib import md5
+
+from coverage.report import get_analysis_to_report
+
+
+class LcovReporter:
+    """A reporter for writing LCOV coverage reports"""
```
nedbat: This is a super-nit, but I prefer that docstrings and comment use sentence punctuation, with a final period.

coverage/lcovreport.py
```diff
@@ -0,0 +1,106 @@
...
+
+        get_lcov currently supports both line and branch coverage,
+        however function coverage is not supported.
+
+        """
+
+        outfile.write("TN:\n")
+        outfile.write(f"SF:{fr.relative_filename()}\n")
+        sourceLines = fr.source().splitlines()
+        for covered in sorted(list(analysis.executed)):
```
nedbat: No need for `list()` here.

coverage/lcovreport.py
```diff
@@ -0,0 +1,106 @@
...
+            # output even if it's technically incorrect
+
+            # The lines below calculate a 64 bit encoded md5 hash of the line
+            # corresponding to the DA lines in the lcov file,
+            # for either case of the line being covered or missed in Coveragepy
+            # The final two characters of the encoding ("==") are removed from
+            # the hash to allow genhtml to run on the resulting lcov file
+            try:
+                line = sourceLines[covered - 1].encode("utf-8")
+            except IndexError:
```
nedbat: According to the commit comment, this except is only for the case of an empty file? If so, it would be better to use an explicit `if sourceLines:` test.

coverage/lcovreport.py
```diff
@@ -0,0 +1,106 @@
...
+        """Produces the lcov data for a single file
+
+        get_lcov currently supports both line and branch coverage,
+        however function coverage is not supported.
+
+        """
+
+        outfile.write("TN:\n")
+        outfile.write(f"SF:{fr.relative_filename()}\n")
+        sourceLines = fr.source().splitlines()
```
nedbat: `sourceLines` is an unusual variable name.  Can we use `source_lines` instead?

coverage/lcovreport.py
```diff
@@ -0,0 +1,106 @@
...
+            # The lines below calculate a 64 bit encoded md5 hash of the line
+            # corresponding to the DA lines in the lcov file,
+            # for either case of the line being covered or missed in Coveragepy
+            # The final two characters of the encoding ("==") are removed from
+            # the hash to allow genhtml to run on the resulting lcov file
+            try:
+                line = sourceLines[covered - 1].encode("utf-8")
+            except IndexError:
+                line = b""
+            hashed = str(base64.b64encode(md5(line).digest())[:-2], encoding="utf-8")
```
nedbat, line 61: I guess tastes vary, but I like `.strip("=")` here more than `[:-2]` since it's more explicit what it's doing.
And "ascii" as encoding also seems more appropriate.

nedbat: But this is where we need to decide what to do about md5: it's not available on FIPS-enabled systems.  Here are some choices:
1. Don't use md5 at all, and leave the hash out of the report.
2. Use md5, but catch the ValueError it will raise on FIPS systems (https://github.com/nedbat/coveragepy/issues/1189)
3. Make lcov unavailable on FIPS systems
4. Let lcov just fail if used on FIPS systems

I don't know what the effect would be of omitting the hash, so I don't know which to pick.

nedbat (changes requested):
Thanks for pushing forward on this feature. I haven't review the whole pull request, but wanted to get you some feedback. More tomorrow.

tests/test_lcov.py
```diff
@@ -0,0 +1,290 @@
...
+
+"""Test LCOV-based summary reporting for coverage.py"""
+
+import coverage
+
+
+from tests.coveragetest import CoverageTest
+
+
+class lcovTest(CoverageTest):
```
nedbat: `LcovTest` :)

tests/test_lcov.py
```diff
@@ -0,0 +1,290 @@
...
+                    self.assertAlmostEqual(cuboid_volume(2),8)
+                    self.assertAlmostEqual(cuboid_volume(1),1)
+                    self.assertAlmostEqual(cuboid_volume(0),0)
+                    self.assertAlmostEqual(cuboid_volume(5.5),166.375)
+
+        """,
+        )
+
+    def run_coverage(self):
+        """Run coverage.py on main_file.py, and create an LCOV report."""
```
nedbat: This function doesn't make an lcov report as it claims to.

tests/test_lcov.py
```diff
@@ -0,0 +1,290 @@
...
+        cov = coverage.Coverage(source=["."])
+        self.start_import_stop(cov, "main_file")
+        return cov
+
+    def get_lcov_report_content(self):
+        """Return the content of the LCOV report for `module`."""
```
nedbat: This function doesn't have a module parameter.

tests/test_lcov.py
```diff
@@ -0,0 +1,290 @@
...
+        filename = "coverage.lcov"
+        line_list = []
+        with open(filename) as f:
+            for line in f:
+                line_list.append(line)
+            return line_list
```
nedbat: I'm not a fan of returning a list of lines here.  I think the tests will read better if this function returns the text as one string, and the literals are multi-line strings instead of lists of lines.

tests/test_lcov.py
```diff
@@ -0,0 +1,290 @@
...
+        cov.lcov_report()
+        actual_result = self.get_lcov_report_content()
+        assert expected_result == actual_result
+
+    def test_simple_line_coverage_two_files(self):
+        """Test that line coverage is created when coverage is run,
+        and matches the output of the file below"""
+        self.create_initial_files()
+        cov = coverage.Coverage(source=".")
+        self.start_import_stop(cov, "main_file")
```
nedbat: Shouldn't the main_file be run by the test_file, instead of directly?

nedbat (changes requested):
OK, I've read the whole thing.  We'll also need the docs updated, if you want to do that.

coverage/lcovreport.py
TLATER: The hash is mainly used to ensure the correct lines are marked as covered, to my knowledge. Upstream's `genhtml`, for example, which uses an lcov file and the original source tree to produce an html page with annotated source files, will throw an error if it finds a hash that mismatches: https://github.com/linux-test-project/lcov/blob/master/bin/genhtml#L5308.

Genhtml is of course an unlikely usecase with pycoverage, but it's the reference consumer.

It is up to the implementation of the consumer to define what the md5 is used for, but it is marked as optional, so any tools that break on a missing md5 are inherently broken. There shouldn't be any issue with simply omitting the hash when it can't be set, and I can't think of many other uses than "has our user accidentally changed their source". I think that was probably a more common use case around the time lcov added that feature :)

It *would* be sad if pycoverage produced different outputs depending on the machine it is run on. Maybe hence it's best to remove this functionality in general.

I think a fifth option is to lobby upstream for supporting a more modern hash, as much as that seems unnecessary. It's honestly a bit silly that python mandates thinking about FIPS compliance with no opt-out if you use its md5 hash, and even more silly that some default python distributions seem to enable FIPS compliance (this seems to be a MacOS thing?).

It'll be hard to come up with a backwards-compatible design for that, so I suspect the solution would be to deprecate having hashes at all, since those don't really seem that useful anymore in 2022 when checking for accidental source changes is anyway so easy and often unnecessary.

nedbat: Thanks for the backstory on the hash values.

> It's honestly a bit silly that python mandates thinking about FIPS compliance with no opt-out if you use its md5 hash

Is this Python's fault? FIPS forbids md5 as an algorithm.

> some default python distributions seem to enable FIPS compliance (this seems to be a MacOS thing?)

I haven't encountered this.  Can you provide more information?

TLATER: > Is this Python's fault? FIPS forbids md5 as an algorithm.

No, not at all. I just think the situation is a bit silly for backwards-compatibility with older formats that use md5 for legitimate reasons (or even just for migration purposes). It's besides the point, sorry for venting here.

> I haven't encountered this. Can you provide more information?

[This support page](https://support.apple.com/en-us/HT209638) seems to suggest a FIPS-compliance mode is enabled by default. Reading further, it may come down to processor support for the hash, rather than something explicitly disabled on the software side, but this is all ultimately besides the point.

I think I'll spend a bit more time reading up on where exactly the limitation comes from, and then raise an issue on the lcov upstream to see if they're interested enough in allowing lcov to be run in FIPS-compliant environments. No need for coveragepy to be involved in that :)

nedbat:
btw, for extra credit, it would be good to have tests in test_cmdline.py also.

tests/test_lcov.py
bradb423: Hi there,

Is there a way to do the above? I am just about finished with making the changes, but this change in the tests is stumping me at the moment.

Cheers,

Brad

nedbat: Sorry, I'm not sure what you are asking. My point was simply that you don't need to call `start_import_stop` on both the main_file and the test_file.  test_file.py imports main_file, so just calling it on test_file will be enough, and is how coverage would normally be run.

bradb423: Thank you, it seems I misunderstood the review comment. I have amended the tests now, thank you!

coverage/lcovreport.py
bradb423: @TLATER is right here on the use of the hash within LCOV. I am perfectly happy to remove the md5 hash if you would prefer @nedbat.

nedbat:
@bradb423 Good to see progress! Let me know when you are ready.

bradb423:
Hi @nedbat Thanks for your patience. I'd say that this is ready for another look. Please let me know how you feel about the current state, along with the md5 hash :)

nedbat:
Thanks, this looks good to merge.  We'll keep the md5 as it is, and see if we get any concerns from people.  I'll do some styling and clean-up after the merge.

Thanks for doing this! <3

nedbat (approved):

nedbat:
This is now released as part of [coverage 6.3](https://pypi.org/project/coverage/6.3).  Thanks again! <3

Issue #624, opened 2017-12-20
COVERAGE_FILE should be a command-line parameter instead.


nedbat:
*Originally reported by* **Anonymous**

----------------------------------------

Hi, Ned. First of all, thank you for this wonderful tool.

In a UNIX environment, I would expect a tool to allow me to specify the output path from the command line, instead of setting it through an environment variable. Much so that I think coverage.py is the first tool I have seen with such behavior. In light of this, I am proposing to remove the `COVERAGE_FILE` environment variable, and add an option to `run` subcommand that exactly serves in place of the mentioned environment variable. 

Comparing the existing API,

```bash
# Regular use
COVERAGE_FILE=my_file coverage run my_program.py

# alternatively
export COVERAGE_FILE=my_file
coverage run my_program.py

# Write to stdout of the process, to enable piping
COVERAGE_FILE=/dev/stdout my_program.py
```

with my proposal,

```bash
# Regular use
coverage run -o my_file my_program.py

# Write to stdout of the process, to enable piping
coverage run -o /dev/stdout my_program.py
```

I believe the command line argument option is more elegant, and something a programmer would expect. I personally had to search documents on how to specify the output file.

If you think this is a good idea, I am willing to do the work on the codebase to implement this feature.

----------------------------------------
- Bitbucket: https://bitbucket.org/ned/coveragepy/issue/624

nedbat:
I suppose we could add -o support, though you're the first person to ask for it in a, so I don't see much demand for it.

We are not going to remove the environment variable support. I'm not sure why you would propose to.

Piping to stdout seems like an odd thing to want.  I try to discourage people from dealing with the data file directly, so I hope people are not building tools that will read one from stdin.

nedbat:
*Original comment by* **Cengiz Kaygusuz (Bitbucket: [cngkaygusuz](https://bitbucket.org/cngkaygusuz), GitHub: [cngkaygusuz](https://github.com/cngkaygusuz))**

----------------------------------------

I'm the author of the issue. I posted it thinking I was logged in :/

Can't really comment on the demand of other people. Maybe there will be more of such demands in the future, who knows. My specific use case is that I need to run a test suite both in python 2 and python 3, and combine them. To do so, I need to write the output of `coverage` to two distinct files. Running two commands appended with something like `COVERAGE_FILE=my_file` felt awkward, hence this proposal.

[...]

PR #1304, opened 2022-01-15
coverage run: add a -o option

nbfalcon:
Add a way to specify the coverage filename manually for `coverage run` (option: `-o`) and correspondingly for the reporting commands and `combine` (`-c`, `--input-coverage`):

```shell
python3 -m coverage run -o output.coverage test.py
python3 -m coverage xml -c output.coverage -o -
```

Use case: I am writing an IDE plugin to integrate `coverage.py`, and I would like to use the python package unmodified. However, I'd like to avoid the creation of a `.coverage` file at the project root, and would instead like the coverage to be written to a file of my choosing (somewhere in the IDE's data directory).

This idea could (possibly) be taken further, if use cases are found (for my use case, the feature as implemented in this PR is sufficient), so I have not implement them here:
- `coverage erase` could take a list of coverage files instead of only one (with this PR: via `-c`)
- `input-coverage` for `coverage run`: allowing the coverage data of a previous run to be extended while writing to a new file (would require further core changes, introducing an `input_data_file` argument to the `Coverage` class)

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -384,7 +384,7 @@ CMDS = {
     ),
 
     'erase': CmdOptionParser(
-        "erase", GLOBAL_ARGS,
+        "erase", [Opts.input_coverage] + GLOBAL_ARGS,
         description="Erase previously collected coverage data.",
     ),
 
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -129,6 +129,12 @@ class Opts:
         metavar="OUTFILE",
         help="Write the recorded coverage information to this file. Defaults to '.coverage'"
     )
+    input_coverage = optparse.make_option(
+        '-c', '--input-coverage', action='store', dest="input_coverage",
+        metavar="INPUT",
+        help="Read coverage data for report generation from this file (needed if you have "
+             "specified -o previously). Defaults to '.coverage'"
+    )
     json_pretty_print = optparse.make_option(
         '', '--pretty-print', action='store_true',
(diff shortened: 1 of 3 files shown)


nbfalcon:
I don't understand why `tox` (docs and quality/pylint) is failing.

nedbat:
Hi, thanks for doing this. It's been requested before (https://github.com/nedbat/coveragepy/issues/624), but I've relied on the COVERAGE_FILE environment variable to cover the need.

I haven't reviewed the code yet, though I think a `--data-file` option for both run and report would be the clearest way to expose it on the command line.

About the tests: I added some explanation to the tox.ini and the cmd.rst files in the latest commit on master.  Let me know if they help make it clear?

nbfalcon:
Thanks for the suggestion, I've changed the options to --data-file (I agree that it is less ambiguous than -c and -o). As for the tests, they run through on my machine, but there are some failing GitHub checks (I don't know if that is a problem), and I don't know why they fail.

I've somehow overlooked COVERAGE_FILE (I should have looked more closely at the docs), and that env var is actually a better solution for my use-case, since it won't require the latest coverage.py version for my plugin, so I no longer personally need this PR (though if you consider this feature worthwhile, I'll implement all suggestions you have).

nedbat:
I squashed these commits and merged them in commit ba884e41. Thanks for getting this over the line!

nedbat:
This is now released as part of [coverage 6.3](https://pypi.org/project/coverage/6.3).  Thanks again! <3

PR #1313, opened 2022-01-27
reset CoverageData._lock at fork

graingert:
might help with #1310


nedbat:
Thanks, though this doesn't seem to help #1310. :(

ProsperousHeart:
Is this still valid?

KRRT7:
@graingert is this still valid?

graingert:
I don't know - I suspect it is but it's too outdated to fix here I think

PR #1314, opened 2022-01-27
restore old signal handler asap

graingert:
might help with #1310

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -605,10 +605,10 @@ class Coverage:
 
     def _on_sigterm(self, signum_unused, frame_unused):
         """A handler for signal.SIGTERM."""
+        signal.signal(signal.SIGTERM, self._old_sigterm)
         self._atexit("sigterm")
         # Statements after here won't be seen by metacov because we just wrote
         # the data, and are about to kill the process.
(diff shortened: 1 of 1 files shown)


coverage/control.py
```diff
@@ -605,10 +605,10 @@ def _atexit(self, event="atexit"):
 
     def _on_sigterm(self, signum_unused, frame_unused):
         """A handler for signal.SIGTERM."""
+        signal.signal(signal.SIGTERM, self._old_sigterm)
         self._atexit("sigterm")
```
graingert, line 609: this way if someone is furiously mashing ctrl+c they don't start multiple `self._atexit` calls

graingert: might also be prudent to check if `self._atexit.__code__` is in the `frame_unused` currently

graingert:
This seems to help a lot! However I still got a flake here https://github.com/graingert/distributed/runs/4967453793?check_suite_focus=true#step:13:1240

nedbat:
Thanks, but this doesn't seem to fix #1310.  Detailed reproduction steps are at https://github.com/nedbat/coveragepy/issues/1310#issuecomment-1024178941

nedbat:
I've added a setting to opt-in to SIGTERM handling.

PR #1317, opened 2022-02-01
Use precision argument on cmd

Kludex:
Precision gets default to 0 when using as cmd argument.

I guess it's related to https://github.com/pytest-dev/pytest-cov/issues/403

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -55,8 +55,8 @@ jobs:
         uses: "actions/setup-python@v2"
         with:
           python-version: "${{ matrix.python-version }}"
-          cache: pip
-          cache-dependency-path: 'requirements/*.pip'
+          #cache: pip
+          #cache-dependency-path: 'requirements/*.pip'
 
       - name: "Install dependencies"
         run: |
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -4,7 +4,7 @@
 """Command-line support for coverage.py."""
 
 import glob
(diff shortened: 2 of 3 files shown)


Kludex:
The test doesn't seem a very good one... As it passes without my modifications.

nedbat:
Thanks.  Can you show the behavior you are fixing?  It looks like it works to me:
```
$ coverage report --precision=1
Name      Stmts   Miss    Cover
-------------------------------
igor.py     274    233    15.0%
-------------------------------
TOTAL       274    233    15.0%

$ coverage report --precision=4
Name      Stmts   Miss       Cover
----------------------------------
igor.py     274    233    14.9635%
----------------------------------
TOTAL       274    233    14.9635%
```

Kludex:
I guess on your example, would be:

```bash
$ coverage report --fail-under=15.0 --precision=4
Name      Stmts   Miss       Cover
----------------------------------
igor.py     274    233    14.9635%
----------------------------------
TOTAL       274    233    14.9635%
```
You can change the value `fail-under` to 14.97 as well...

Kludex:
It's the total coverage that gets rounded using `precision=0`.

nedbat:
Thanks, that's the part I was missing.

nedbat:
I see what you mean about the test. I'll fix up the test after merging this. Thanks.

Kludex:
Thank you. :)

nedbat:
This is now released as part of [coverage 6.3.1](https://pypi.org/project/coverage/6.3.1).

PR #1329, opened 2022-02-21
added set timeout function

qerogram:
Hi~

First of all, saying thank you for creating a good library. I was using the function well through the coverage run command line, but I wanted to measure the coverage of the program that continued to operate without turning off based on the timeout value. Unfortunately, "coverage" does not support timeout-based measurements. So I've developed it simply, and I hope you'll review it. 

ex) 
```
$ coverage run --timeout=10 app.py
```

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -222,6 +222,11 @@ class Opts:
         help="Display version information and exit.",
     )
 
+    timeout = optparse.make_option(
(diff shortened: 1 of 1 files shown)


nedbat:
Thanks, but I think this would be better done in a program of your own.  You can use the Coverage API to do this in your own test suite.

PR #1352, opened 2022-04-08
Use PyFrame_GetLasti()

vstinner:
Replace MyFrame_lasti() with PyFrame_GetLasti(): Python 3.11.0b1 adds
PyFrame_GetLasti().

Add pythoncapi_compat.h header file to get PyFrame_GetLasti() on
Python 3.10 and older. File copied from:
https://github.com/python/pythoncapi_compat

Drop support for alpha versions of Python 3.11 which don't have
PyFrame_GetLasti().

### DIFF
--- a/coverage/ctracer/pythoncapi_compat.h
+++ b/coverage/ctracer/pythoncapi_compat.h
@@ -1,0 +1,503 @@
+// Header file providing new C API functions to old Python versions.
+//
+// File distributed under the Zero Clause BSD (0BSD) license.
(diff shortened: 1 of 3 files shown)


vstinner:
https://github.com/nedbat/coveragepy/pull/1353 is an attempt to keep support for Python 3.11 alpha versions.

vstinner:
Since https://github.com/nedbat/coveragepy/pull/1353 tests pass on Python 3.11a5, I close this PR.

PR #1353, opened 2022-04-08
Use PyFrame_GetLasti()

vstinner:
Replace MyFrame_lasti() with PyFrame_GetLasti(): Python 3.11.0b1 adds
PyFrame_GetLasti().

Add pythoncapi_compat.h header file to get PyFrame_GetLasti() on
Python 3.10 and older. File copied from:
https://github.com/python/pythoncapi_compat

Keep compatibility code for alpha versions of Python 3.11 (alpha 1 to
alpha 7). Remove PyFrame_GetLocals(), PyFrame_GetGlobals() and
PyFrame_GetBuiltins() in pythoncapi_compat.h which are not compatible
with alpha versions of Python 3.11. Hack pythoncapi_compat.h to not
define PyFrame_GetLasti() on Python 3.11a1 and newer.

### DIFF
--- a/coverage/ctracer/pythoncapi_compat.h
+++ b/coverage/ctracer/pythoncapi_compat.h
@@ -1,0 +1,519 @@
+// Header file providing new C API functions to old Python versions.
+//
+// File distributed under the Zero Clause BSD (0BSD) license.
+// Copyright Contributors to the pythoncapi_compat project.
+//
+// Homepage:
+// https://github.com/python/pythoncapi_compat
+//
+// Latest version:
+// https://raw.githubusercontent.com/python/pythoncapi_compat/master/pythoncapi_compat.h
+//
+// SPDX-License-Identifier: 0BSD
+
+#ifndef PYTHONCAPI_COMPAT
+#define PYTHONCAPI_COMPAT
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <Python.h>
+#include "frameobject.h"          // PyFrameObject, PyFrame_GetBack()
+
+
+// Compatibility with Visual Studio 2013 and older which don't support
+// the inline keyword in C (only in C++): use __inline instead.
+#if (defined(_MSC_VER) && _MSC_VER < 1900 \
+     && !defined(__cplusplus) && !defined(inline))
+#  define PYCAPI_COMPAT_STATIC_INLINE(TYPE) static __inline TYPE
+#else
+#  define PYCAPI_COMPAT_STATIC_INLINE(TYPE) static inline TYPE
+#endif
+
+
+// C++ compatibility: _Py_CAST() and _Py_NULL
+#ifndef _Py_CAST
+#  ifdef __cplusplus
+extern "C++" {
+    namespace {
+        template <typename type, typename expr_type>
+            inline type _Py_CAST_impl(expr_type *expr) {
+                return reinterpret_cast<type>(expr);
+            }
+
+        template <typename type, typename expr_type>
+            inline type _Py_CAST_impl(expr_type const *expr) {
+                return reinterpret_cast<type>(const_cast<expr_type *>(expr));
+            }
+
+        template <typename type, typename expr_type>
+            inline type _Py_CAST_impl(expr_type &expr) {
+                return static_cast<type>(expr);
+            }
+
+        template <typename type, typename expr_type>
+            inline type _Py_CAST_impl(expr_type const &expr) {
+                return static_cast<type>(const_cast<expr_type &>(expr));
+            }
+    }
+}
+#    define _Py_CAST(type, expr) _Py_CAST_impl<type>(expr)
+#  else
+#    define _Py_CAST(type, expr) ((type)(expr))
+#  endif
+#endif
+#ifndef _Py_NULL
+#  ifdef __cplusplus
+#    define _Py_NULL nullptr
+#  else
+#    define _Py_NULL NULL
+#  endif
+#endif
+
+// Cast argument to PyObject* type.
+#ifndef _PyObject_CAST
+#  define _PyObject_CAST(op) _Py_CAST(PyObject*, op)
+#endif
+
+
+// bpo-42262 added Py_NewRef() to Python 3.10.0a3
+#if PY_VERSION_HEX < 0x030A00A3 && !defined(Py_NewRef)
+PYCAPI_COMPAT_STATIC_INLINE(PyObject*)
+_Py_NewRef(PyObject *obj)
+{
+    Py_INCREF(obj);
+    return obj;
+}
+#define Py_NewRef(obj) _Py_NewRef(_PyObject_CAST(obj))
+#endif
+
+
+// bpo-42262 added Py_XNewRef() to Python 3.10.0a3
+#if PY_VERSION_HEX < 0x030A00A3 && !defined(Py_XNewRef)
+PYCAPI_COMPAT_STATIC_INLINE(PyObject*)
+_Py_XNewRef(PyObject *obj)
+{
+    Py_XINCREF(obj);
+    return obj;
+}
+#define Py_XNewRef(obj) _Py_XNewRef(_PyObject_CAST(obj))
+#endif
+
+
+// bpo-39573 added Py_SET_REFCNT() to Python 3.9.0a4
+#if PY_VERSION_HEX < 0x030900A4 && !defined(Py_SET_REFCNT)
+PYCAPI_COMPAT_STATIC_INLINE(void)
+_Py_SET_REFCNT(PyObject *ob, Py_ssize_t refcnt)
+{
+    ob->ob_refcnt = refcnt;
+}
+#define Py_SET_REFCNT(ob, refcnt) _Py_SET_REFCNT(_PyObject_CAST(ob), refcnt)
+#endif
+
+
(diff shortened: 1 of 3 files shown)


vstinner:
> Add pythoncapi_compat.h header file to get PyFrame_GetLasti()

With this header file, it also becomes possible to drop MyFrame_GetCode() from coverage/ctracer/util.h and use PyFrame_GetCode() on all Python versions. It would allow to fix a **reference leak** in coverage affecting Python 3.9 and newer: PyFrame_GetCode() returns a strong reference, whereas MyFrame_GetCode() returns a borrow reference on Python 3.8 and older. Currently, coverage always make the assumption that MyFrame_GetCode() returns a borrow reference which is wrong on Python 3.9 and newer.

I chose to restrict this PR to PyFrame_GetLasti() to keep it short.

vstinner:
Current implementation of PyFrame_GetLasti() in **upstream** pythoncapi_compat.h:

```
// bpo-40421 added PyFrame_GetLasti() to Python 3.11.0b1
#if PY_VERSION_HEX < 0x030B00B1 && !defined(PYPY_VERSION)
PYCAPI_COMPAT_STATIC_INLINE(int)
PyFrame_GetLasti(PyFrameObject *frame)
{
#if PY_VERSION_HEX >= 0x030A00A7
    // bpo-27129: Since Python 3.10.0a7, f_lasti is an instruction offset,
    // not a bytes offset anymore. Python uses 16-bit "wordcode" (2 bytes)
    // instructions.
    if (frame->f_lasti < 0) {
        return -1;
    }
    return frame->f_lasti * 2;
#else
    return frame->f_lasti;
#endif
}
#endif
```

In this  PR, I replaced ``PY_VERSION_HEX < 0x030B00B1`` (beta1) with ``PY_VERSION_HEX < 0x030B00A1`` (alpha1) to prevent a compiler error on Python 3.11 alpha versions. On Python 3.11.0 final will be released, coverage code can be simplified to remove support for Python 3.11 alpha versions.

IMO it's better to rely on pythoncapi_compat.h since this code is non trivial.

vstinner:
@nedbat: Are you ok to use pythoncapi_compat.h? As I wrote, it would allow to fix the coverage MyFrame_GetCode() refleak (I can work on a fix, once this PR is merged).

coverage/ctracer/pythoncapi_compat.h
```diff
@@ -0,0 +1,466 @@
...
+_PyFrame_GetBackBorrow(PyFrameObject *frame)
+{
+    return PYCAPI_COMPAT_CAST(PyFrameObject *,
+                              _Py_XStealRef(PyFrame_GetBack(frame)));
+}
+#endif
+
+
+// bpo-40421 added PyFrame_GetLasti() to Python 3.11.0b1
+#if PY_VERSION_HEX < 0x030B00A1 && !defined(PYPY_VERSION)
```
jouve, line 209: `0x030B00B1` ?

nedbat:
@vstinner Yes, I am fine with using pythoncapi_compat.h.  But when I try this pull request on main, I still have compile errors:
```
coverage/ctracer/tracer.c:533:48: error: no member named 'co_code' in 'struct PyCodeObject'
    PyObject * pCode = MyFrame_GetCode(frame)->co_code;
                       ~~~~~~~~~~~~~~~~~~~~~~  ^
coverage/ctracer/tracer.c:534:43: error: no member named 'f_lasti' in 'struct _PyInterpreterFrame'
    real_call = (PyBytes_AS_STRING(pCode)[PyFrame_GetLasti(frame) + 1] == 0);
                                          ^~~~~~~~~~~~~~~~~~~~~~~
coverage/ctracer/util.h:20:47: note: expanded from macro 'PyFrame_GetLasti'
#define PyFrame_GetLasti(f)    ((f)->f_frame->f_lasti * 2)
                                ~~~~~~~~~~~~  ^
coverage/ctracer/tracer.c:701:56: error: no member named 'co_code' in 'struct PyCodeObject'
            PyObject * pCode = MyFrame_GetCode(frame)->co_code;
                               ~~~~~~~~~~~~~~~~~~~~~~  ^
coverage/ctracer/tracer.c:702:25: error: no member named 'f_lasti' in 'struct _PyInterpreterFrame'
            int lasti = PyFrame_GetLasti(frame);
                        ^~~~~~~~~~~~~~~~~~~~~~~
coverage/ctracer/util.h:20:47: note: expanded from macro 'PyFrame_GetLasti'
#define PyFrame_GetLasti(f)    ((f)->f_frame->f_lasti * 2)
                                ~~~~~~~~~~~~  ^
```
I'm going to work on this, but if you have suggestions, I would welcome them!

vstinner:
I deleted my comment. I went 1 week in holiday and it is way enough for me to forget the context :-) This PR is supposed to support Python 3.11 alpha versions. I'm not sure what's going on.

coverage/ctracer/util.h
```diff
@@ -5,25 +5,19 @@
...
+// Python 3.11.0b1 adds PyFrame_GetLasti().
 #include <internal/pycore_frame.h>
-#define MyFrame_lasti(f)    ((f)->f_frame->f_lasti * 2)
-#elif PY_VERSION_HEX >= 0x030A00A7
-// The f_lasti field changed meaning in 3.10.0a7. It had been bytes, but
-// now is instructions, so we need to adjust it to use it as a byte index.
-#define MyFrame_lasti(f)    ((f)->f_lasti * 2)
-#else
-#define MyFrame_lasti(f)    ((f)->f_lasti)
+#define PyFrame_GetLasti(f)    ((f)->f_frame->f_lasti * 2)
```
Fidget-Spinner: The version check is off because since 3.11a1, `f->f_frame` refers to the internal `_PyInterpreterFrame`/`_interpreter_frame` struct, not `PyFrameObject`.  Unfortunately I suspect you need to add the `_PyInterpreterFrame_LASTI` macro and use it like how `GetLastI` uses it for 3.11a1-a7, like what CPython does https://github.com/python/cpython/blob/64a54e511debaac6d3a6e53685824fce435c440c/Objects/frameobject.c#L1162.
`_PyInterpreterFrame_LASTI`:
https://github.com/python/cpython/blob/main/Include/internal/pycore_frame.h#L69.

Alternatively, you can manually expand that macro in `pythoncapi_compat.h` :).

Fidget-Spinner (commented):
@nedbat as a drive-by reviewer, I hope I'm not repeating anything you already know:

`_PyInterpreterFrame` is the internal frame object used in CPython since 3.11a1. We no longer use `PyFrameObject` except when a debugger requests it. When a debugger *does* ask for it, it's created on the fly.

Fidget-Spinner:
Also, `co_code` has been gone since https://github.com/python/cpython/commit/2bde6827ea4f136297b2d882480b981ff26262b6. Which is 3.11a7 and up only. For that you'd need the internal `_PyCode_GetCode` (Huh, I just realised we should probably expose that in CPython's API).

You *could* access `co_code_adaptive` directly, but that would include the quickened bytecodes, which don't allow tracing.

vstinner:
> Also, co_code has been gone since https://github.com/python/cpython/commit/2bde6827ea4f136297b2d882480b981ff26262b6. Which is 3.11a7 and up only. For that you'd need the internal _PyCode_GetCode (Huh, I just realised we should probably expose that in CPython's API).

No. Please don't use an internal C API to read co_code. This attribute is accessible in Python, so we should provide a public function for that.

vstinner:
Or you can already use ``PyObject_GetAttrString(code, "co_code")``.

vstinner:
See https://github.com/python/cpython/issues/91397 "Move the PyCodeObject structure to the internal C API (make the structure opaque)".

Fidget-Spinner:
> No. Please don't use an internal C API to read co_code. This attribute is accessible in Python, so we should provide a public function for that.

Sorry I wasn't trying to suggest people use it. Anyways, its in the internal API and completely private/hidden so its impossible to use :).  Also I highly do not recommend accessing the struct member `co_code_adaptive` directly.

nedbat:
I've solved the `co_code` problem using `PyObject_GetAttrString(code, "co_code")` in #1368.  Can you take a look to see if there's something I've done wrong?

nedbat:
@vstinner I've gotten things to work in 956f0fde, but I would still rather use this approach when you have time to help me with it.

kumaraditya303:
@nedbat I have a WIP branch [^1] which uses `pythoncapi_compat` but it does not support alpha versions but as the beta is already out so supporting alpha versions is less important. It will make coveragepy a little slower because `pythoncapi_compat` uses strong references instead of current borrowed references though.

[^1]:https://github.com/kumaraditya303/coveragepy

vstinner:
> @nedbat I have a WIP branch [1](https://github.com/nedbat/coveragepy/pull/1353#user-content-fn-1-b7c3bebea816c70ceec203f0e04e9cf7) which uses pythoncapi_compat

Can you create a PR for that?

vstinner:
I rebased my PR on the master branch and I updated pythoncapi_compat.h. My PR drops support for Python 3.11 alpha versions. IMO it's too complicated to support 3.11 alpha versions and it's not worth it.

I prefer to write a separated PR for PyCode_GetCode(), since it requires to add Py_DECREF(). I prefer to keep this PR easy to review.

PR #1354, opened 2022-04-09
Add timestamp and version info in header and footer.

### LINKED ISSUE #1351: In HTML report, display coverage creation time at the top of the page.
Often I found myself not realizing that I was looking at an outdated coverage report.

I know that the the report includes the creation time at the bottom of the page. When working in large codebase, we would need to scroll down first before seeing that information. It's not too obvious that the info exists, and it's not always the first thing I'd look at.

I think it would be great if the creation time can be displayed at the top of the page as well.

Display the "created at .." at the top.

Sample mockup:

index.html
Screen Shot 2022-04-05 at 1 43 53 PM

module coverage report
Screen Shot 2022-04-05 at 1 45 25 PM

ammaraskar:
This fixes #1351 by adding the timestamp and version info to both the header and footer. As mentioned in the issue, this is mostly useful for larger projects where long files push down this important info far down the page.

I also lined up the footer to be on the same place in both the index page and the file listing page.

| Page          | Before                                                                                                        | After                                                                                                         |
|---------------|---------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------|
| Light (Index) | image | image |
| Dark (Index)  | image | image |
| Light (File)  | image | image |
| Dark (File)   | image | image |

### DIFF
--- a/coverage/htmlfiles/index.html
+++ b/coverage/htmlfiles/index.html
@@ -48,6 +48,13 @@
             <input id="filter" type="text" value="" placeholder="filter..." />
         </form>
     </div>
+
+    <nav>
+        <p>
+            <a class="nav" href="{{__url__}}">coverage.py v{{__version__}}</a>,
+            created at {{ time_stamp }}
+        </p>
+    </nav>
 </header>
 
 <main id="index">
@@ -109,12 +116,12 @@
 </main>
 
 <footer>
-    <div class="content">
+    <nav>
         <p>
             <a class="nav" href="{{__url__}}">coverage.py v{{__version__}}</a>,
             created at {{ time_stamp }}
         </p>
-    </div>
+    </nav>
 </footer>
 
 </body>
--- a/coverage/htmlfiles/pyfile.html
+++ b/coverage/htmlfiles/pyfile.html
@@ -73,6 +73,13 @@
             <button type="button" class="button_first_chunk" data-shortcut="1">Goto first highlighted chunk</button>
         </div>
     </div>
+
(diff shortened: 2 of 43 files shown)


nedbat:
I appreciate the pull request (with pictures!) and the attention to detail.  But the footer should remain aligned with the code and header, as it was on the index page.  The file page can change to match.

coverage/htmlfiles/index.html
```diff
@@ -48,6 +48,13 @@ <h1>{{ title|escape }}:
             <input id="filter" type="text" value="" placeholder="filter..." />
         </form>
     </div>
+
+    <nav>
```
nedbat, line 52: I don't think I would use a `<nav>` element here.  The information here isn't navigation.

nedbat:
I implemented this slightly differently in commit 53e5dd3e.

Mariatta:
It doesn't seem like the timestamp is showing at the top of the `index.html` page, but it is showing up correctly everywhere else. 


Would it be possible to add it in to the `index.html` page as well?

Thank you.

nedbat:
OK, as of 6ef78073 it's at the top of the index page too.  I feel like it might be in too many places now...  Maybe we don't need it at the bottom, but we'll see if anyone else voices an opinion (they won't).

Mariatta:
Thanks! Just tried it and it looks great to me.

Screen Shot 2022-05-20 at 2 42 58 PM

nedbat:

PR #1359, opened 2022-04-16
Use tomllib on Python 3.11

hauntsaninja:
Python 3.11 includes tomllib in the stdlib for parsing TOML (PEP 680).


hauntsaninja:
(draft PR since coverage doesn't currently seem to work on Python 3.11 alpha 7, which is the first to include tomllib)

nedbat:
Thanks!

nedbat:

PR #1364, opened 2022-04-25
Add keyboard shortcuts

jftsang:
Couple of keyboard shortcuts for navigating between pages on the generated reports

[ and ] will go to the previous or next file
u will take you up to the index page
? will show or hide the help panel

### DIFF
--- a/coverage/html.py
+++ b/coverage/html.py
@@ -273,7 +273,6 @@ class HtmlReporter:
         else:
             next_html = "index.html"
 
-        print(prev_html, html_filename, next_html)
         ensure_dir(self.directory)
         if not os.listdir(self.directory):
             self.directory_was_empty = True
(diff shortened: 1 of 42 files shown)


nedbat:
Thanks, this is great, but there is one problem that is blocking me from merging it.  If you use `--skip-covered` or `--skip-empty`, then the next/prev shortcuts link to pages that don't exist.  Do you have an idea about how to fix that?

nedbat:
Also, I might be merging #1354, which would cause some conflicts here that will need rebasing.

nedbat:
I've merged this and fixed the next/prev links so they take skipped files into account, in 78c0915d1638e2077d55989b1de856fe40378fce

Thanks for this!

jftsang:
Awesome! Sorry I've not been able to help with that last bit

nedbat:

PR #1366, opened 2022-04-29
Build and test tweaks to help with cross-compilation

zooba:
For context, I've been testing whether a range of popular libraries are going to work on Windows ARM64. (This requires compiling on a regular x64 machine and then copying the wheel to an ARM64 one for testing.)

The good news is, coverage seems to be just fine without any changes. However, because of a few assumptions in the test suite about always testing an in-place build, I had to make some tweaks to be able to run tests. My proposed tweaks should be fine for current uses, but they also allow the following:

* support `SETUPTOOLS_EXT_SUFFIX` when building to override the pyd tag on Windows (used with `setup.py build_ext -L <path>` to point at [pythonarm64](https://www.nuget.org/packages/pythonarm64/) import libraries to do the cross-compile)
* allow `COVERAGE_INSTALL_ARGS` to override how the tests install coverage into a venv (allows to point at a wheel rather than rebuilding from the sources)
* allow `python igor.py remove_extension --from-install` to delete the extension module from `site-packages` rather than only the source tree
* other changes to allow removing the `coverage` directory from the source tree before tests so that the installed copy will be used instead.

I've tested these on my own Windows ARM64 machine, though unfortunately there aren't any available on CI services yet. If you wanted to start releasing (preview) wheels for win-arm64 you can, but there's no support (yet) in `cibuildwheel` or `build` to do it (because those tools don't really have a concept of cross-compilation for Windows at all... yet ;-) ).

### DIFF
--- a/igor.py
+++ b/igor.py
@@ -56,7 +56,7 @@ def do_show_env():
         print(f"  {env} = {os.environ[env]!r}")
 
 
-def do_remove_extension():
+def do_remove_extension(*args):
     """Remove the compiled C extension, no matter what its name."""
 
     so_patterns = """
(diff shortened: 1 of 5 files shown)


nedbat:
Thanks! I've squashed the two commits and moved your PR description into the commit message on master.

PR #1377, opened 2022-05-15
Use tomllib instead of tomli on Python 3.11

agronholm:
The "toml" extra should not cause tomli to be installed on 3.11.


nedbat:
Thanks, i should have done my research better before giving you the go-ahead on IRC: #1359 is basically the same.

Issue #1339, opened 2022-03-08
Code runs 77x times slower when using Decimal running with coverage.py on PyPy 3.9 (7.3.8)


felipou:
**Describe the bug**
Running with coveragepy on PyPy drastically slows down programs using Decimal:
```
~/tmp/pypyperftest$ time poetry run coverage run -m covperftest.covperftest 10000000
10000000

real	7m37,942s
user	7m36,118s
sys	0m1,348s
~/tmp/pypyperftest$ time poetry run python -m covperftest.covperftest 10000000
10000000

real	0m5,891s
user	0m5,267s
sys	0m0,606s
```
I've noticed this issue while experimenting with PyPy's newest versions in the past, trying to migrate to Python 3.8. At the time I didn't pay much attention, just gave up. Now that PyPy launched a new version supporting 3.9, I tried to investigate a bit more and got the results above. At first I thought that maybe it was because PyPy is using Decimal with a pure Python implementation, but forcing the C implementation yields similar results (if not slower).

**To Reproduce**
1. What version of Python are you using?
    - **Pypy 3.9 (7.3.8) - but I've noticed this problem while using earlier versions of PyPy.**
1. What version of coverage.py shows the problem?
    - **I tested with the latest release, version 6.3.2.**
1. What versions of what packages do you have installed?
    - **I tried to test in a virtualenv with as few packages as possible:**
```
cffi==1.15.0
coverage==6.3.2
covperftest==1.0.0
greenlet==0.4.13
hpy==0.0.3
readline==6.2.4.1
tomli==2.0.1
```
1. What code shows the problem?
    - **I'm not sure where the problem in coveragepy is (or if there's even a problem at all, maybe it's something related to how Decimal works on PyPy, but I thought it would be better to report this here).**
1. What commands did you run?
    - **I created a sample script to test this, which I executed in a virtualenv managed using poetry:**
```
import random
import sys
from decimal import Decimal

size = int(sys.argv[1]) if len(sys.argv) > 1 else 100000

values = []
for _ in range(size):
    values.append(Decimal(3) * random.randint(0, 10000))
print(len(values), file=sys.stderr)
```

**Expected behavior**
Although a worse performance is to be expected when getting the coverage, it gets much worse when using Decimal. I tried the same code using just an integer, and the execution time was around 4.5x slower when using coveragepy. But when using Decimal, it was 77x slower.

nedbat:
[...]

Issue #482, opened 2016-03-14
Branch coverage does weird things for yield statements


nedbat:
*Originally reported by* **Tristan Seligmann (Bitbucket: [mithrandi](https://bitbucket.org/mithrandi), GitHub: [mithrandi](https://github.com/mithrandi))**

----------------------------------------

Consider this example:
```python
def gen(): 
    print 'hi mom!' 
    yield 42 
    print 'not covered' 
print gen().next() 
```

If I run it without --branch, coverage tells me the first print and yield are executed, but the third print isn't; all as expected.

If I run it with --branch, coverage tells me the first print was executed, but that the yield was not executed at all. Huh?

----------------------------------------
- Bitbucket: https://bitbucket.org/ned/coveragepy/issue/482

nedbat:
*Original comment by* **Pi Delport (Bitbucket: [pjdelport](https://bitbucket.org/pjdelport), GitHub: [pjdelport](https://github.com/pjdelport))**

----------------------------------------

This makes partial sense: the `yield` is an expression that branches control flow, and doesn't actually fully cover the line until control is resumed back into the generator. For example, consider:

```python
def gen():
    print('hi mom!')
    print((yield 42), 1 / 0)
    print('not covered')

print gen().next()
```

However, that doesn't explain why it counts the `yield` line as completely missed instead of partial.

nedbat:
Duplicate of #440.

PR #1381, opened 2022-05-24
speed up python tracer with frame.f_trace_lines = False

cfbolz:
This is a Python-only version of #791, in response to the extreme slowdowns of coverage on pypy reported in #1339.

the PR uses the python >= 3.7 feature of being able to disable line tracing by setting the frame attribute f_trace_lines to False. This can be used for the frames of functions that we aren't collecting coverage information for (eg those functions in the stdlib).

I tried to used the `perf/` code to look into the impact, but as @nedbat wrote on #1339, the code had bitrotted and even after fixing I couldn't really use it, since it doesn't *have* any code that shouldn't be traced. Instead, I used [pytest-html](https://github.com/pytest-dev/pytest-html) as a coverage benchmark (because somebody had reported extremely slow PyPy coverage runs there as well, see [#482](https://github.com/pytest-dev/pytest-html/issues/482)).

Running coverage of pytest-html with the python tracer gives the following numbers:

py3.10 timid:
before: 457.84s
after: 48.63s

pypy3.9-dev:
before: 290.38
after: 159.96

So quite a substantial improvement on both CPython and PyPy (I had to re-measure CPython a few times before I believed the speedup, but it was consistent for me).

This PR intentionally does not contain the equivalent change to the C tracer. I briefly tried this but it did not show a huge change in performance. So somebody might have to investigate that more carefully.

### DIFF
--- a/coverage/pytracer.py
+++ b/coverage/pytracer.py
@@ -159,6 +159,8 @@ class PyTracer:
                 if tracename not in self.data:
                     self.data[tracename] = set()
                 self.cur_file_data = self.data[tracename]
+            else:
+                frame.f_trace_lines = False
             # The call event is really a "start frame" event, and happens for
             # function calls and re-entering generators.  The f_lasti field is
             # -1 for calls, and a real offset for generators.  Use <0 as the


nedbat:
Thanks, the results are great!

I threw together a benchmark.py program and compared this PR to 6.4 in a few configurations on the pytest-html suite.  I also got good results, though not as great as yours. We are probably measuring different amounts of the overhead, but it doesn't matter: even my lesser results are good improvements:
```
python3.10, cov=none: 17.990s
python3.10, cov=6.4: 24.888s
python3.10, cov=6.4 timid: 103.707s
python3.10, cov=PR 1381: 23.754s
python3.10, cov=PR 1381 timid: 54.133s
pypy3.9, cov=none: 36.442s
pypy3.9, cov=6.4: 193.428s
pypy3.9, cov=PR 1381: 134.366s
```

nedbat:
This is now released as part of [coverage 6.4.1](https://pypi.org/project/coverage/6.4.1).

nedbat:
This is now released as part of [coverage 6.4.1](https://pypi.org/project/coverage/6.4.1).

PR #1388, opened 2022-05-30
Pytracer improvements

cfbolz:
Here are two further improvements of the python tracer performance to help with #1339:
- cache the bound method of `_trace`
- use the observation that if we stay within the same file from one function to a called function, the decision whether we should trace or not does not change to save dictionary lookups

unfortunately I couldn't get the benchmark script to run (I fixed the hardcoded path, but even then tests stopped running after a run or two, didn't dig deeper). But here are some numbers for attrs and pytest-html:

| Implementation               | none | Coverage master timid | Coverage PR timid | Speedup |
|------------------------------|------|----------------------:|------------------:|---------|
| CPython 3.10 attrs           | 87s  |                  335s |              274s | 22%     |
| PyPy 3.9 nightly attrs       | 62s  |                  575s |              356s | 62%     |
| CPython 3.10 pytest-html     | 33s  |                   63s |               59s | 7%      |
| PyPy 3.9 nightly pytest-html | 45s  |                  185s |              156s | 19%     |

This change is a little bit more complex than the previous one, so I would understand if you don't deem it worth the extra complexity.

### DIFF
--- a/coverage/pytracer.py
+++ b/coverage/pytracer.py
@@ -151,20 +151,29 @@ class PyTracer:
                     self.started_context,
                 )
             )
-            filename = frame.f_code.co_filename
-            self.cur_file_name = filename
-            disp = self.should_trace_cache.get(filename)
-            if disp is None:
-                disp = self.should_trace(filename, frame)
-                self.should_trace_cache[filename] = disp
 
-            self.cur_file_data = None
-            if disp.trace:
-                tracename = disp.source_filename
-                if tracename not in self.data:
-                    self.data[tracename] = set()
-                self.cur_file_data = self.data[tracename]
-            else:
+            # improve tracing performance: when calling a function, both caller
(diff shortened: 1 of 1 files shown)


cfbolz:
(I messed up the numbers in the table, fixing momentarily)

cfbolz:
fixed, I had swapped some of the numbers, sorry about that and thanks @mattip for catching!

nedbat:
I'm surprised it's worthwhile to use an attribute to avoid a dict lookup.  I would have thought the extra bytecodes would be worse than just doing the dict lookup, but I see the results.  This seems to be about a 10% improvement on my benchmark.

nedbat:
This is now released as part of [coverage 6.4.1](https://pypi.org/project/coverage/6.4.1).

nedbat:
This is now released as part of [coverage 6.4.1](https://pypi.org/project/coverage/6.4.1).

Issue #1350, opened 2022-03-30
Option to excluded no-context lines from total calculation


bbeattie-phxlabs:
**Is your feature request related to a problem? Please describe.**

When I run coverage against my code that has _zero_ tests run, it's frustrating to see a coverage total above 0%.

As a simple example:

```
$ cat myfile.py

def foo(a, b):
    return a + b
```

```
$ pytest myfile.py --cov=. --cov-report term-missing:skip-covered --quiet

---------- coverage: platform linux, python 3.10.2-final-0 -----------
Name        Stmts   Miss  Cover   Missing
-----------------------------------------
myfile.py       2      1    50%   2
-----------------------------------------
TOTAL           2      1    50%

no tests ran in 0.01s
```

**Describe the solution you'd like**

Executed lines are annotated with their contexts as per https://coverage.readthedocs.io/en/latest/contexts.html#dynamic-contexts. An option could use this to say "consider lines with no context as though they were commented with `# pragma: no cover`, excluding them from total calculations."

**Describe alternatives you've considered**

Alternative solutions would be to take the resultant `.coverage` sqlite DB and manually compute the contextless-excluded total from there, but that seems notably hackier and wouldn't affect any of the generated reported numbers

**Additional context**

I'm happy to put the work into a pull request to address this feature request if the maintainers see value in pursuing it!

bbeattie-phxlabs:
At first glance, augmenting `Analysis.__init__` with 

```
self.no_context = set(
    line
    for line, contexts in self.data.contexts_by_lineno(self.filename).items()
    if not any(contexts)
)
```

and the instantiation of self.numbers with

```            
  n_statements=len(self.statements - self.no_context),
  n_excluded=len(self.excluded - self.no_context),
  n_missing=len(self.missing - self.no_context),
```

which results in an output of 

```
---------- coverage: platform linux, python 3.10.2-final-0 -----------
Name        Stmts   Miss  Cover   Missing
-----------------------------------------
myfile.py       1      1     0%   2
-----------------------------------------
TOTAL           1      1     0%
```

where 0% is the result I'm looking for. I've written no tests so far.

The above snipped needs to sit behind a configuration option, but figured I'd get clear go-ahead before creating a PR.

nedbat:
[...]

PR #1389, opened 2022-05-30
Ignoring contextless lines during reporting

bbeattie-phxlabs:
As detailed in https://github.com/nedbat/coveragepy/issues/1350, coveragepy currently reports a single file with zero tests like

```
def foo_a(x, y):
  return x + y
```

as having 50% coverage. With [dynamic contexts](https://coverage.readthedocs.io/en/latest/contexts.html#dynamic-contexts) we can make use of line context to exclude lines executed outside of tests (imports, function definitions, etc) to arrive at an arguably more accurate lower bound of 0% for files without tests.

I suspect the PR will need to be augmented with additional documentation and tests to confirm this behaviour, but I figured getting the PR in for discussion might move things in the right direction.


coverage/results.py
```diff
@@ -26,6 +26,13 @@ def __init__(self, data, precision, file_reporter, file_mapper):
         self.executed = executed
         self.missing = self.statements - self.executed
 
+        # Optionally ignore statements that have no context
```
bbeattie-phxlabs, line 29: This change and the change below are the crux of this PR.

nedbat:
I'm still wondering whether this extra configuration is worth it.  I know you have an example where no tests results in 50% coverage, but in real projects the number will be much lower.  We could simply add an FAQ that explains that many lines get executed just by importing files.

bbeattie-phxlabs:
To the thrust of my initial ticket that prompted the PR:

> My motivation is to coax developers in my org towards best practices without immediately requiring 100% code coverage. So a project template that says `--cov-fail-under=1` requires new projects to at least write one test, and we can go from there. But given coverage's current inclusion of let's call it "contextless code", I can't set a stable lower threshold at the moment.

I read on the overall size and complexity of the changes requested as fairly minimal, but I get that at the end of the day coveragepy is your baby. Happy to defer to your call on this. 👍

bbeattie-phxlabs:
Alternatively, I'm just as happy to maintain my own fork with the `ignore_contextless` behaviour as a hardcoded default if there isn't sufficiently broad appeal to this option. :)

PR #1399, opened 2022-06-10
When stashing the singleton to sys.modules, use an actual module object.

yilei:
At least this won't trip anyone iterating through sys.modules and expecting the values are actual modules.

### DIFF
--- a/coverage/debug.py
+++ b/coverage/debug.py
@@ -12,6 +12,7 @@ import os
 import pprint
 import reprlib
 import sys
+import types
 import _thread
 
 from coverage.misc import isolate_module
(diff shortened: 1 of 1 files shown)


nedbat:
Hi, thanks for this.  I'm curious what lead you to this? Did you run into a problem?

yilei:
What leads to this: yes, we have some (old) code that implements mocking by iterating through sys.modules and examine the module content.

This is an improvement on its own so that https://github.com/python/typeshed/blob/b4c3e2c3e16b4d96de0cfdcf23e91ad0240d9166/stdlib/sys.pyi#L44 still holds.

nedbat:
Thanks.

nedbat:
This is now released as part of [coverage 6.4.2](https://pypi.org/project/coverage/6.4.2).

PR #1401, opened 2022-06-11
perf: cache the lookup of should_trace_cache

nedbat:
Doesn't seem to improve performance by more than about .5%

| pyver | proj | without cache | with cache | compare |
|----|----|----|----|----|
| python3.11 | bug1339.py | 0.767 s | 0.819 s | 106.77% |
| python3.11 | bm_sudoku.py | 50.346 s | 50.142 s | 99.60% |
| python3.11 | bm_spectral_norm.py | 60.180 s | 59.727 s | 99.25% |


nedbat:
This is about what I expected, since the dict lookup is only once per function call, and dict lookups are fast.

PR #1403, opened 2022-06-14
Fix paths when running coverage from root

arthurio:
I'm running coverage from a docker image where my source code is at `/`.
I couldn't figure out why the paths were all starting with `/` but not if I moved things in a subfolder (e.g. `/subfolder`).
With the current implementation, `RELATIVE_DIR` would be equal to `//` instead of `/` because we suffix no matter what.

I tested this change manually in my docker image with `python@3.10.4` and `coverage==6.0.2`.

### DIFF
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -26,7 +26,7 @@ def set_relative_directory():
 
     # The current directory
     abs_curdir = abs_file(os.curdir)
-    if abs_curdir != os.sep:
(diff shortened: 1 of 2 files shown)


arthurio:
The simple tests I added are heavily mocked... let me know if you can think of something better!

nedbat:
Thanks.  I see what you mean about the mocking, but i don't see a better way.

nedbat:

PR #1405, opened 2022-06-17
Fix globbing of filenames that contain special glob characters

earthman1:
When specifying a coverage filename using the `COVERAGE_FILE` environment variable, special characters are not handled correctly. I use that facility to set the name of the coverage file to match the name of the test that I am running in pytest, so that I can associate the coverage with the specific test that generated that coverage. Some test names contain brackets, so the name of the test might be `exampletest[b-a]`, and I transform that name into `.coverage.exampletest[b-a]` and then run
```
COVERAGE_FILE=.coverage.exampletest[b-a] pytest --cov "exampletest[b-a]"
```
Which results in the error:
```
re.error: bad character range b-a at position 5
```
Any special characters in a filename specified for creating a new file should be escaped before globbing for similar filenames.

### DIFF
--- a/coverage/data.py
+++ b/coverage/data.py
@@ -70,7 +70,7 @@ def combinable_files(data_file, data_paths=None):
         if os.path.isfile(p):
             files_to_combine.append(os.path.abspath(p))
         elif os.path.isdir(p):
-            pattern = os.path.join(os.path.abspath(p), f"{glob.escape(local)}.*")
+            pattern = glob.escape(os.path.join(os.path.abspath(p), local)) +".*"
             files_to_combine.extend(glob.glob(pattern))
         else:
             raise NoDataError(f"Couldn't combine from non-existent path '{p}'")
--- a/coverage/data.py
+++ b/coverage/data.py
@@ -70,7 +70,7 @@ def combinable_files(data_file, data_paths=None):
         if os.path.isfile(p):
             files_to_combine.append(os.path.abspath(p))
         elif os.path.isdir(p):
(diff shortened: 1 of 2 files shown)


coverage/data.py
```diff
@@ -70,7 +70,7 @@ def combinable_files(data_file, data_paths=None):
         if os.path.isfile(p):
             files_to_combine.append(os.path.abspath(p))
         elif os.path.isdir(p):
-            pattern = os.path.join(os.path.abspath(p), f"{local}.*")
```
BenjaminSchubert, line 73: The `p` here should also be escaped, there is no guarantee that it won't have `[` or `]` in it

coverage/sqldata.py
```diff
@@ -774,7 +774,7 @@ def erase(self, parallel=False):
         file_be_gone(self._filename)
         if parallel:
             data_dir, local = os.path.split(self._filename)
-            localdot = local + ".*"
+            localdot = glob.escape(local) + ".*"
             pattern = os.path.join(os.path.abspath(data_dir), localdot)
```
BenjaminSchubert: The `data_dir` should also be escaped here, there is no guarantee that it won't have `[` or `]` in it either

BenjaminSchubert (commented):
I was hitting the same bug, but my directory itself had the `[3.9]` in it.

Overall both parts of the `join` should be escaped. It might be nicer to first create the full path, and escape it, then add `.*` at the end maybe?

Otherwise it seems good to me, though I am not a maintainer :)

nedbat:

PR #1409, opened 2022-06-20
force / in actual_path() on Windows

altendky:
This is intended to make relative paths generated on Windows usable in Linux and macOS.

https://github.com/nedbat/coveragepy/issues/991

Draft for:
- [ ] Tests
- [ ] Discussion around better solutions or other places that should be addressed.

### DIFF
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -133,7 +133,7 @@ if env.WINDOWS:
                     tail = f
                     break
             actpath = os.path.join(head, tail)
-        actpath.replace("\\", "/")
+        actpath = actpath.replace("\\", "/")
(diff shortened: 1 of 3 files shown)


altendky:
So this trivial change ran ok in my single test usage at https://github.com/Chia-Network/chia-blockchain/runs/6972851503?check_suite_focus=true, but a bunch of tests fail here.  I'll keep looking around and trying to get a better feel but if you have some design guidance in mind, what with you actually knowing the code base already `:]`, I'm all ears.  Or eyes, I suppose.

nedbat:
I think this has been handled in commit 7df8609f.

PR #1414, opened 2022-07-01
fixed "mNan" id in scroll markers

LordBaryhobal:
Scroll markers had the id "mNan" because the line number was not extracted from the correct elements.


nedbat:
Thanks, though I'm curious how you noticed? :)

LordBaryhobal:
I was just inspecting the HTML and looking at how they were created, noticed they all had "mNan" as id and tracked down the error in the source code :)

Issue #1434, opened 2022-08-09
type stubs


KotlinIsland:
I would like some type stubs for coverage, I see that you support >= 3.7 so types in the code would be possible (although perhaps not desirable).

I'm happy to drive this. just need to know if you want the stubs inline or in a separate package, or maybe even hosted on typehsed.

nedbat:
I'd be interested in adding annotations right into the code.  I'd been using pycontracts as a pre-historic way of checking types, but that package is unmaintained and is causing pain.

If there are annotations, I want to run mypy, so we'd need some new tooling for that too.

KotlinIsland:
Sounds cool, I'll give it a shot and make a PR, would you be interested in [basedmypy](https://github.com/KotlinIsland/basedmypy)?

KotlinIsland:
While I'm here should I implement black?

nedbat:
That's generous of you, but I don't want black for this repo, thanks.  I don't know anything about basedmypy, but I'm guessing I don't want that either, but maybe I'm wrong?

KotlinIsland:
It's basically a fork of mypy with additional functionality.
basedmypy:
```py
def foo(a=1)
    print(a)
```
mypy:
```py
def foo(a: int = 1) -> None:
    print(a)
```

andersk:
For projects that use `coverage` as a library (such as test runners), it would be better if the type annotations worked with standard mypy and didn’t require a custom fork.

nedbat:
I've started adding annotations.  About 3/4 of the coverage/*.py files now check cleanly.  Please let me know if I should do something differently.

KotlinIsland:
@nedbat this is awesome! One outstanding task is to add a 'py.typed' file. See #1550

PR #1442, opened 2022-08-29
WIP type annotations

KotlinIsland:
Very WIP 

Contains:
- partial type annotations for some modules
- github workflow integration

addresses: #1434


coverage/__init__.py
```diff
@@ -10,13 +10,13 @@
 
 import sys
 
-from coverage.version import __version__, __url__, version_info
+from coverage.version import __version__ as __version__, __url__ as __url__, version_info as version_info
```
nedbat, line 13: Why add these `as` clauses? They're the same names they already were.

nedbat:
I haven't looked at this closely, but do you have a way to run mypy as part of the tools?

coverage/__init__.py
KotlinIsland: In python typing, this nonsense is semantically understood as "re-export this import as a public member of the module".

I believe this crude hack was made because they didn't want to introduce any new syntax for typing.

It is mentioned in [pep-484](https://peps.python.org/pep-0484/#stub-files)

KotlinIsland:
> I haven't looked at this closely, but do you have a way to run mypy as part of the tools?

I'm unfamiliar with your project configuration (`make`, `pip-compile`) but I can give it a go.

nedbat:
Thanks for taking this all on.  I [added some details](https://github.com/nedbat/coveragepy/commit/fcb782df20dee25944f27a8a3104e8d5d150c249) to the Contributing section of the docs about dependencies and `make upgrade`, though I wonder if I've added enough? Can you take a look and let me know? Thanks so much.

nedbat:
What is the future of this pull request?  It might be too hard to get everything type-checked, since the master branch is continuing development.  Is there a way to get the tooling in place, and gradually make progress on adding the annotations?

KotlinIsland:
Hmm, sure. I could clean this up and get a minimal state ready to merge when I get time.

Kludex:
> What is the future of this pull request? It might be too hard to get everything type-checked, since the master branch is continuing development. Is there a way to get the tooling in place, and gradually make progress on adding the annotations?

@nedbat On `uvicorn`, we created an issue to track our progress: https://github.com/encode/uvicorn/issues/998.

Each PR had an entry on the mypy configuration file (`setup.cfg` for us).

I can create a similar issue for coverage, and help to coordinate, if wanted.

nedbat:
I started a more gradual approach.  As of 212f6520, four files on master are checked with mypy.  I'd be happy to have help adding more files gradually, and eventually making the checks more strict and complete.  I'll close this PR now, but feel free to re-open it if you can make this way work with the latest code.

nedbat:
As of now (commit b059a67f), all of the files in coverage/ and tests/ are checked by mypy.

PR #1460, opened 2022-10-04
🐛 Fix the branch protection gate check @ GHA

webknjaz:
This adds a GHA job that reliably determines if all the required dependencies have succeeded or not.

It is now in use in aiohttp (and other aio-libs projects), CherryPy, some of the Ansible repositories, pip-tools, all of the jaraco's projects (like `setuptools`, `importlib_metadata`), some of hynek's projects, some PyCQA, PyCA, PyPA and pytest projects, a few AWS Labs projects (to my surprise).

The story behind this is explained in more detail at https://github.com/marketplace/actions/alls-green#why.

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -91,13 +91,19 @@ jobs:
              exit 1
           fi
 
-  # A final step to give a simple name for required status checks.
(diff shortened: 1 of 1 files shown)


nedbat:
Thanks so much!  I've slightly edited this and merged it as commit fdc40d01.

Issue #1418, opened 2022-07-08
Support report terminal output in Markdown Table format


sullivancolin:
The current terminal report is great, but wondered if there would be interest in adding a new output format or flag to support markdown table syntax.

This would enable nice display of coverage reports in rendered html that contains the markdown output.  Some examples where this might be nice:

1. Users could used [Cog](https://nedbatchelder.com/code/cog/) to automate the inclusion of coverage report information in documentation.
2.  GitHub actions workflow summaries could have nicely formatted tables of coverage information, [example here](https://til.simonwillison.net/github-actions/job-summaries)

Current appearance:
```
Name                      Stmts   Miss  Cover
---------------------------------------------
my_program.py                20      4    80%
my_module.py                 15      2    86%
my_other_module.py           56      6    89%
---------------------------------------------
TOTAL                        91     12    87%
```

Sample Markdown syntax:
````
| Name               | Stmts      | Miss     |    Cover |
| ------------------ | ---------- | -------- | -------- |
| my_program.py      |         20 |        4 |      80% |
| my_module.py       |         15 |        2 |      86% |
| my_other_module.py |         56 |        6 |      89% |
| **TOTAL**          |     **91** |   **12** |  **87%** |
````

Sample Rendering:

| Name               | Stmts      | Miss     |    Cover |
| ------------------ | ---------- | -------- | -------- |
| my_program.py      |         20 |        4 |      80% |
| my_module.py       |         15 |        2 |      86% |
| my_other_module.py |         56 |        6 |      89% |
| **TOTAL**          |     **91** |   **12** |  **87%** |

**Describe the solution you'd like**
Ideally the a flag to the default `report` subcommand could indicate to use markdown syntax like:

```sh
$ coverage report --markdown
| Name               | Stmts      | Miss     |    Cover |
| ------------------ | ---------- | -------- | -------- |
| my_program.py      |         20 |        4 |      80% |
| my_module.py       |         15 |        2 |      86% |
| my_other_module.py |         56 |        6 |      89% |
| **TOTAL**          |     **91** |   **12** |  **87%** |
$
```

This would allow the output to be piped to a file in the GitHub action summary, or captured by cog to inject into a markdown file.

**Describe alternatives you've considered**
If the `report` subcommand is not suitable place to add this flag then a new subcommand specifically for markdown reporting might be a good alternative with `coverage markdown`

```sh
$ coverage markdown
| Name               | Stmts      | Miss     |    Cover |
[...]

PR #1479, opened 2022-10-23
Support report terminal output in Markdown Table format #1418

stepeos:
```
$ tox -e py37
py37 develop-inst-noop: foo/bar/coveragepy
py37 installed: atomicwrites==1.4.1,attrs==22.1.0,backports.functools-lru-cache==1.6.4,cffi==1.15.1,colorama==0.4.5,-e git+ssh://git@github.com/stepeos/coveragepy.git@b3179db61fcf593fa41ce3f862f0af39d3383f7a#egg=coverage,decorator==5.1.1,distlib==0.3.6,dnspython==2.2.1,eventlet==0.33.1,exceptiongroup==1.0.0rc9,execnet==1.9.0,filelock==3.8.0,flaky==3.7.0,future==0.18.2,gevent==21.12.0,greenlet==1.1.3,hypothesis==6.54.6,importlib-metadata==4.12.0,iniconfig==1.1.1,packaging==21.3,platformdirs==2.5.2,pluggy==1.0.0,py==1.11.0,PyContracts @ https://github.com/slorg1/contracts/archive/c5a6da27d4dc9985f68e574d20d86000880919c3.zip,pycparser==2.21,pyparsing==3.0.9,pytest==7.1.3,pytest-forked==1.4.0,pytest-xdist==2.5.0,qualname==0.1.0,six==1.16.0,sortedcontainers==2.4.0,tomli==2.0.1,typing_extensions==4.3.0,virtualenv==20.16.5,zipp==3.8.1,zope.event==4.5.0,zope.interface==5.4.0
py37 run-test-pre: PYTHONHASHSEED='1893709823'
py37 run-test: commands[0] | python igor.py zip_mods
py37 run-test: commands[1] | python setup.py --quiet build_ext --inplace
py37 run-test: commands[2] | python -m pip install -q -e .
WARNING: You are using pip version 22.0.4; however, version 22.3 is available.
You should consider upgrading via the 'foo/bar/coveragepy/.tox/py37/bin/python -m pip install --upgrade pip' command.
py37 run-test: commands[3] | python igor.py test_with_tracer c
=== CPython 3.7.13 with C tracer (.tox/py37/bin/python) ===
bringing up nodes...
.............................................................................................................................x.x.......................ss.s........... [ 13%]
.................................................................................................................................................................... [ 27%]
.................................................................................................................................................................... [ 41%]
...................................................................................................................................................................... [ 55%]
.s.........s....................................................................................................................................s...............s... [ 69%]
.............s........................................................s.........................................................................s.................... [ 83%]
............................................................s....................................................................................................... [ 97%]
............................                                                                                                                                         [100%]
1168 passed, 11 skipped, 2 xfailed in 98.02s (0:01:38)
py37 run-test: commands[4] | python igor.py remove_extension
py37 run-test: commands[5] | python igor.py test_with_tracer py
=== CPython 3.7.13 with Python tracer (.tox/py37/bin/python) ===
bringing up nodes...
...........................................................................................................................x..x.........................ss..s....... [ 13%]
.........................................................................................................ssssss.s......s...ss....................................... [ 27%]
.................................................................................................................................................................... [ 41%]
.................................................................................................................................................................... [ 55%]
......s........s.............................................................................................................................s..............s....... [ 69%]
.s............s..............................sssssssssssssssssssss...........s..............s....................................................................... [ 83%]
........s.........................................................s................................................................................................. [ 97%]
...........................ss....                                                                                                                                    [100%]
1133 passed, 46 skipped, 2 xfailed in 88.22s (0:01:28)
_________________________________________________________________________________ summary __________________________________________________________________________________
  py37: commands succeeded
  congratulations :)

```


nedbat:
Thanks for taking this on.  I have some concerns with the code as it stands now, but you'll have to tell me how long you want to continue working on this, or whether I should take it over at some point.

1. (easy) The option should be `--format`, not `--format-text`.
2. (real work) There's too much code duplicated between the two formatting methods.  The original code mixed together the logic of the table (sorting, how many columns, which lines to include) with the formatting of the table, which was not great.  This pull request now duplicates the logic parts between `_report_text` and `_report_markdown`, which is a real problem.  Better would be to have one function that pulls together the data and gets it ready for formatting, then uses either a text or markdown formatting function.  Someone someday is going to ask for another format, and I would hate to see the logic duplicated a third time.
3. (not too bad) How to test these things is always tricky.  Once step 2 is done, we can use unit testing of the formatting function to avoid having to add too much duplication in the test methods.  A few cases like you've done it would probably be enough.
4. (easy) The quality jobs will pass if you run `make prebuild` locally to fix up some of the computed content in the docs.

Let me know how I can help with any of these.

stepeos:
> 1. option should be `--format`, not `--format-text`

I implemented the cmdline argument as --format, but the variable is called format_text since format is also a builtin python method. format was linted on my editor as function. Still change?

> 2\. have one function that pulls together the data and gets it ready for formatting

this was my idea as well, and I tried implementing it that way. _report_text takes the data and the command line options and builds the format string and inserts the values.
The major 'logic' part that i can see is duplicated is the sorting of the lines.
I'd love to give it one more try if you tell me which data should be input to _report_markdown and _report_text, and what exact parts should be moved to report()

nedbat:
> > 1. option should be `--format`, not `--format-text`
> 
> I implemented the cmdline argument as --format, but the variable is called format_text since format is also a builtin python method. format was linted on my editor as function. Still change?

You can do something like this: 
```python
  output_format = optparse.make_option(
        '', '--format', action='store', metavar="FORMAT", dest="output_format",
        help="Output format, either text (default) or markdown",
    )
```
and it will put the option into an argument called `output_format`.

> > 2. have one function that pulls together the data and gets it ready for formatting
> 
> this was my idea as well, and I tried implementing it that way. _report_text takes the data and the command line options and builds the format string and inserts the values. The major 'logic' part that i can see is duplicated is the sorting of the lines. I'd love to give it one more try if you tell me which data should be input to _report_markdown and _report_text, and what exact parts should be moved to report()

I haven't looked closely at the code to see exactly how this would work out, but the logic includes sorting of lines, what columns to include, whether to include a total line, and so on.  I imagine one function producing a list of tuples: each list element is one row, the tuples are the cells for each column in that row.  The header and total lines could be the first and last elements in the list, or it might work out better to pass them separately, I'm not sure.

stepeos:
@nedbat thanks for you patience & suggestions, I really appreciate it (I'm learning a lot).

I have refactored summary.py. I moved all logic to the top level method `report()`.
Data for Header, lines, TOTAL line and summary-lines ("x files skipped") is generated in the `report` and passed to either markdown- or text-formatted report-methods.

`make prebuild` did not fix the docs error unfortunately :/

nedbat:
Thanks, I'll take a look very soon.

For the docs error, you have to go to the action, and find the failing line, which can be tricky, because you have to [keep opening sections until you find it](https://github.com/nedbat/coveragepy/actions/runs/3323798890/jobs/5494600718#step:5:23):
```
/home/runner/work/coveragepy/coveragepy/coverage/control.py:
docstring of coverage.control.Coverage.__init__:19:
Unknown directive type "veresionadded".
```

coverage/summary.py
```diff
@@ -30,6 +30,112 @@ def writeout(self, line):
...
+            self.writeout(rule)
+            line_items = [
+                h_form[item].format(str(value),
+                name_len=max_name, n=max_n-1) for item, value in zip(header, total_line)]
+            text = "".join(line_items)
+            self.writeout(text)
+
+        for end_line in end_lines:
+            self.writeout(end_line)
+        return self.total.n_statements and self.total.pc_covered
```
nedbat: You don't need to return this from here, or from `_report_markdown`.

coverage/summary.py
```diff
@@ -30,6 +30,112 @@ def writeout(self, line):
...
+        header_str = "".join(header_items)
+        rule = "-" * len(header_str)
+
+        # Write the header
+        self.writeout(header_str)
+        self.writeout(rule)
+
+        column_order = dict(name=0, stmts=1, miss=2, cover=-1)
+        if self.branches:
+            column_order.update(dict(branch=3, brpart=4))
```
nedbat: `column_order` isn't used in this function (or in `_report_markdown`), so these lines can be deleted.

coverage/summary.py
```diff
@@ -30,6 +30,112 @@ def writeout(self, line):
         self.outfile.write(line.rstrip())
         self.outfile.write("\n")
 
+    def _report_text(self, header, lines_values, total_line, end_lines):
+        "internal method to print report data in text format"
+        # Prepare the formatting strings, header, and column sorting.
+        max_name = max([len(fr.relative_filename()) for (fr, analysis) in \
```
nedbat: You don't need a backslash to continue a line if you are in unbalanced brackets (as you are here).  Also, I prefer to have the closing bracket on a line of its own.  Take a look through other code in the repo to get a sense of the style.

coverage/summary.py
```diff
@@ -30,6 +30,112 @@ def writeout(self, line):
         self.outfile.write(line.rstrip())
         self.outfile.write("\n")
 
+    def _report_text(self, header, lines_values, total_line, end_lines):
+        "internal method to print report data in text format"
```
nedbat: Docstrings should always use triple-quote style, and complete English grammar.

coverage/summary.py
```diff
@@ -84,54 +173,58 @@ def report(self, morfs, outfile=None):
...
         if sort_option == "name":
-            lines = human_sorted_items(lines, reverse=reverse)
+            lines_values.sort(key=lambda tup: (human_key(tup[0]), tup[1]),
+                reverse=reverse)
         else:
-            position = column_order.get(sort_option)
-            if position is None:
-                raise ConfigError(f"Invalid sorting option: {self.config.sort!r}")
-            lines.sort(key=lambda l: (l[1][position], l[0]), reverse=reverse)
+            lines_values.sort(key=lambda tup: (tup[sort_idx], tup[0]),
```
nedbat: I use a line-length of 100, so no need to wrap this line, and probably others in this file.

coverage/summary.py
```diff
@@ -30,6 +30,112 @@ def writeout(self, line):
         self.outfile.write(line.rstrip())
         self.outfile.write("\n")
 
+    def _report_text(self, header, lines_values, total_line, end_lines):
+        "internal method to print report data in text format"
+        # Prepare the formatting strings, header, and column sorting.
+        max_name = max([len(fr.relative_filename()) for (fr, analysis) in \
+            self.fr_analysis] + [5]) + 2
+        n = self.config.precision
```
nedbat: It seems odd to me to be looking at `config.precision` here, since you are getting all of the coverage totals as strings.  You shouldn't need to think about the precision, just deal with the strings as they are, sort of how `max_name` is calculated.

tests/test_api.py
```diff
@@ -590,15 +590,15 @@ def test_source_and_include_dont_conflict(self):
...
-            Name    Stmts   Miss  Cover
-            ---------------------------
-            b.py        1      0   100%
-            ---------------------------
-            TOTAL       1      0   100%
+            Name     Stmts   Miss  Cover
+            ----------------------------
+            b.py         1      0   100%
+            ----------------------------
+            TOTAL        1      0   100%
```
nedbat: Why did this change by one space?

tests/test_api.py
```diff
@@ -1054,13 +1054,17 @@ def pretend_to_be_nose_with_cover(self, erase=False, cd=False):
...
-            Name           Stmts   Miss  Cover   Missing
-            --------------------------------------------
-            no_biggie.py       4      1    75%   4
-            --------------------------------------------
-            TOTAL              4      1    75%
+        cov.load()
+        report = self.get_report(cov, squeeze=False)
+
+        print(report)
+        # cov.report(["no_biggie.py"], show_missing=True)
```
nedbat: This looks like debugging left in the file?

tests/test_plugins.py
```diff
@@ -416,19 +416,20 @@ def test_plugin2_with_text_report(self):
...
+            include=["*.html"], omit=["uni*.html"], show_missing=True)
+        # Name          Stmts   Miss Branch BrPart  Cover  Missing
+        # --------------------------------------------------------
+        # bar_4.html        4      2      0      0    50% 1, 4
+        # foo_7.html        7      5      0      0    29% 1-3, 6-7
+        # --------------------------------------------------------
+        # TOTAL            11      7      0      0    36%
+        assert report.split("\n")[2] == \
+            "bar_4.html        4      2      0      0    50% 1, 4"
+        total = int(report.split("\n")[-2].split(" ")[-1][:-1])
```
nedbat: Why change the way the total is accessed by the test?

tests/test_summary.py
```diff
@@ -317,6 +423,22 @@ def not_covered():
         assert squeezed[6] == "1 file skipped due to complete coverage."
         assert self.last_command_status == 0
 
+        report = self.report_from_command(
```
nedbat: I'm torn about adding all these markdown outputs to the tests.  Is there a way we can test `_report_text` and `_report_markdown` more directly to reduce the overhead?

nedbat (changes requested):
Thanks for pushing forward on this.  Let me know if I'm getting too picky, or if it's too much.

coverage/summary.py
```diff
@@ -30,6 +30,112 @@ def writeout(self, line):
...
+            line_items = [
+                h_form[item].format(str(value).replace("_", "\\_"),
+                name_len=max_name, n=max_n-1) for item, value in zip(header, values)]
+            text = "".join(line_items)
+            self.writeout(text)
+
+        # Write the TOTAL line
+        if total_line:
+            total_form = dict(
+                Name="| {:>{name_len}}** |", Stmts="{:>5}** |", Miss="{:>5}** |",
```
nedbat: I don't understand why the trailing stars are in the format string, but the leading stars are in the values on line 131?

coverage/summary.py
```diff
@@ -30,6 +30,112 @@ def writeout(self, line):
...
+            # build string with line values
+            h_form.update(dict(Cover="{:>{n}}% |"))
+            line_items = [
+                h_form[item].format(str(value).replace("_", "\\_"),
+                name_len=max_name, n=max_n-1) for item, value in zip(header, values)]
+            text = "".join(line_items)
+            self.writeout(text)
+
+        # Write the TOTAL line
+        if total_line:
```
nedbat: I ran a coverage report locally, and this line (and the similar one in `_report_text`) say the condition is always true.  Is it true that these functions can never be called with an empty `total_line`, or do we just not have a test for that case?

coverage/summary.py
stepeos: It was just to less code. If I inserted the stars like 
`"**" + str(value) + "**"`
then without special if else case, the Cover entry would look like:
`****`
On the other hand, I already had to make an if else statement for the missing column, so I'll rewrite it.

coverage/summary.py
stepeos: The total line is always printed, if there is coverage data to report.

The `report` method raises an Exception before creating the total line, if there is no coverage data to report.

tests/test_plugins.py
stepeos: I'll remove most of them and test the few cases that are necessary.

nedbat:
@stepeos this is coming along nicely!

1. I notice now that `coverage report --precision=3` (or other values) can make "100.000%" in the Cover column jam into the previous column with no space between them.  I'm a little lost in the logic about why that is.  Also, I thought I would have had a test to check that it didn't happen, but I guess I don't?
2. I still have formatting tweaks I'd like to make.  I can make a pull request into your branch to do them if you like, or are you still working on that?

nedbat:
Hmm, there is a test that shows precision is fine, so I'm not sure what the difference is with my local running.... I'm still digging into it.

nedbat:
OK, I worked out the conditions for the poor formatting, and added a test in https://github.com/stepeos/coveragepy/pull/1.

nedbat:
I can also commit that new test to master, and you rebase on top of it, since it looks like you'll need to do that soon anyway. Your choice.

stepeos:
Sounds great!
I added you to the repo.

nedbat:
Sorry, I meant I could add the test to my master branch, and you would rebase your pull request on top of it.  But you've already merged the pull request, so never mind :)

stepeos:
To fix that, we'd have to get the  max length of all lines for every column , I reckon.
It'd be easier to increase the width of the coverage column, but you want to keep it like it was before, right?

nedbat:
I've made a styling update pull request: https://github.com/stepeos/coveragepy/pull/2

nedbat (approved):

nedbat:
Thanks again for doing all this.  I'll squash and merge it now. If you have more you want to do to it, we can start another pull request.

nedbat:
BTW, how should I credit you in the changelog? I don't know your name :)

stepeos:
I'm Steve Oswald! And thanks :^)

nedbat:
This is now released as part of [coverage 7.0.0b1](https://pypi.org/project/coverage/7.0.0b1).

PR #1489, opened 2022-11-16
PythonFileReporter: do not expect files without a filename extension to contain Python code

### LINKED ISSUE #1488: [low-prio] Possible bug: files without extensions misidentified as Python?
**Describe the bug**
While running `pytest` with `coverage` (v6.5.0) for `sphinx`, I noticed an unusual warning in the output logs:

```
sphinx/venv/lib/python3.10/site-packages/coverage/report.py:81: CoverageWarning: Couldn't parse Python file '/home/jka/Documents/personal/sphinx/sphinx/templates/quickstart/Makefile_t' (couldnt-parse)
  coverage._warn(msg, slug="couldnt-parse")
```

Digging into the code to figure out where that's coming from, I found this conditional block:

https://github.com/nedbat/coveragepy/blob/0ac24533e7f167518b8653d423de5ff7d6229a42/coverage/python.py#L240-L242

**To Reproduce**
How can we reproduce the problem?  Please *be specific*. Don't link to a failing CI job. Answer the questions below:
1. What version of Python are you using?  `3.10`
1. What version of coverage.py shows the problem? `6.5.0`
1. What code shows the problem?  https://github.com/sphinx-doc/sphinx/commit/cd3f2e435000835dd98a11497440bc16a79ec31c
1. What commands did you run? `pytest --cov=sphinx --cov-config=setup.cfg`

**Expected behavior**
No warning should appear.

**Additional context**
I realize that perhaps there could be input types / environments where files don't have extensions that have been supported for historic reasons - and I've read some of the context in #82.  Even so it seems like maybe the file shouldn't be considered Python-y.

jayaddison:
Possible fixup; could resolve #1488 (if that's confirmed to be a bug worth addressing).

### DIFF
--- a/coverage/python.py
+++ b/coverage/python.py
@@ -244,9 +244,6 @@ class PythonFileReporter(FileReporter):
         # Anything named *.py* should be Python.
         if ext.startswith(".py"):
             return True
-        # Jinja templates are evaluated using Python.
-        if ext.startswith('.jinja'):
-            return True
         # Everything else is probably not Python.
         return False
 
--- a/coverage/python.py
+++ b/coverage/python.py
@@ -244,9 +244,6 @@ class PythonFileReporter(FileReporter):
         # Anything named *.py* should be Python.
         if ext.startswith(".py"):
             return True
-        # A file with no extension should be Python.
-        if not ext:
-            return True
         # Everything else is probably not Python.
         return False
 
--- a/coverage/python.py
+++ b/coverage/python.py
@@ -244,6 +244,9 @@ class PythonFileReporter(FileReporter):
         # Anything named *.py* should be Python.
         if ext.startswith(".py"):
             return True
+        # Jinja templates are evaluated using Python.
+        if ext.startswith('.jinja'):
+            return True
         # Everything else is probably not Python.
         return False
 
--- a/tests/test_filereporter.py
+++ b/tests/test_filereporter.py
@@ -102,3 +102,10 @@ class FileReporterTest(UsingModulesMixin, CoverageTest):
         z1z1 = PythonFileReporter(zip1.zip1)
         assert z1.source() == ""
         assert "# My zip file!" in z1z1.source().splitlines()
+
+    def test_python_extensions(self):
+        f1 = PythonFileReporter("afile.py")
+        f2 = PythonFileReporter("Makefile")
+
+        assert f1.should_be_python()
+        assert not f2.should_be_python()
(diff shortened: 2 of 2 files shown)


jayaddison:
Awaits sphinx-doc/sphinx#11165 (or similar)

ProsperousHeart:
@nedbat is this still valid?

jayaddison:
It's completely possible that it'll be made redundant by other changes, @ProsperousHeart - it's a fairly innocuous warning message that most people probably won't notice (and likely rarely indicates a genuine problem).

That said: I (for no good reason) have the time and energy to follow it, even if it takes months or years, so for me it remains relevant.  It's less like the Sagrada Familia and more like a small signpost that's misaligned on a minor countryside road; we're aware of it and it can be fixed at some point.

(that analogy doesn't quite make sense, because in this case we're dependent on other signposts changing elsewhere.. it's an effort to explain it, though :shrug:)

jayaddison:
(a small note to mention that I continue to be interested in progressing this at some point.  roughly speaking, the things on my tasklist ahead of this are: a bunch of Sphinx `linkcheck`-related changes (pending review, acceptance and stability assurance), before then refreshing and offering an updated Sphinx `.jinja` template suffix changeset, and then finally after that also being merged, accepted, proven stable -- optimistic perhaps -- then a revisit here)

jayaddison:
The Sphinx linkcheck-related changes I had in mind have been merged and release in Sphinx v7.1.0 - I'm going to wait for another week or two to allow time for any feedback and bugreports to arrive, and then I'll get started again on updating the Jinja template suffix logic... and then after that is complete, then this will be ready for reconsideration.

nedbat:
I'm not fond of saying that ".jinja" means Python.  As #1488 shows, the Python code made by Jinja will be named whatever the original template file was called.  So it could be ".j2" or ".rst_t", etc.

jayaddison:
Yep, that makes sense.  I'll revert 69e6ef0de4530c5c1b6ff687fcbfe1a2749c2019, also bringing the branch back into alignment with the issue description and this pull request's title.

jayaddison:
As I've mentioned in a [bugreport comment a few moments ago](https://github.com/nedbat/coveragepy/issues/1488#issuecomment-2241234927), I don't think I had understood a detail about the extent of the requirement to support filenames without dotted suffixes - and so I think that this should indeed be closed.  My apologies for being defensive about closing it; at the time I felt that this would be a long process but one that could be handled with co-ordination and updates solely in Sphinx.

PR #1490, opened 2022-11-17
build: enable CIFuzz

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.
This is a service offered by OSS-Fuzz, on which coveragepy already runs.

CIFuzz can help detect catch regressions and fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzer is built on a pull request and will run for 300 seconds.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'coveragepy'
+        dry-run: false
+        language: python
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'coveragepy'
+        fuzz-seconds: 300
+        dry-run: false
+        language: python
+    - name: Upload Crash
+      uses: actions/upload-artifact@v3
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts


nedbat:
Thanks! There was also some mention some place of moving the fuzzing code itself into this repo so that we could extend it.  How do we make that happen?

DavidKorczynski:
> There was also some mention some place of moving the fuzzing code itself into this repo so that we could extend it. How do we make that happen?

Move this code https://github.com/google/oss-fuzz/blob/master/projects/coveragepy/fuzz_parse.py into this library. We can move this anytime now -- feel free to make a PR on OSS-Fuzz that remove it and moves it up here. We should adjust https://github.com/google/oss-fuzz/blob/master/projects/coveragepy/build.sh and https://github.com/google/oss-fuzz/blob/master/projects/coveragepy/Dockerfile accordingly, but this is likely just switching a link.

In order to build `fuzz_parse.py` in the OSS-Fuzz environment, we use https://github.com/google/oss-fuzz/blob/master/projects/coveragepy/build.sh It would be great to continue using `compile_python_fuzzer` for the OSS-Fuzz environment since it makes it possible to get the coverage reports (thanks to this repo :) !) as well as introspector reports (still early stage for Python: https://github.com/ossf/fuzz-introspector).

If we place the fuzzer in `test/fuzzers/` of this repo, then we can adjust the build script on OSS-Fuzz to use:
```
for fuzzer in $(find ./test/fuzzers/ -name 'fuzz_*.py'); do
  compile_python_fuzzer $fuzzer
done
```
The source folder in the `find` command is the only thing changed. That will make it possible to simply put new fuzzers in `test/fuzzers/*` and OSS-Fuzz will start running them continuously automatically. 

I can also make a PR that does the above -- let me know if you'd like me to and I'll do this tomorrow Friday.

ProsperousHeart:
Are this and #1497 still valid @nedbat ?

nedbat:
@DavidKorczynski Sorry I've let this and #1497 sit for a few years.  Are these still current?  I'm a tiny bit reluctant to take on new actions, but also am baffled when I get a fuzzing report from the centralized runs that Google does.

PR #1497, opened 2022-11-30
test: add fuzzing set up

DavidKorczynski:
This is an effort to migrate fuzzers from OSS-Fuzz to upstream.

I've added it such that it's also a unit test that is run by way of pytest. The reason for doing so is to make it easier to test that the fuzzers won't break during development.

If you are interested in this addition, then I can refine the build script on the OSS-Fuzz side (https://github.com/google/oss-fuzz/blob/master/projects/coveragepy/build.sh) such that it builds all files prefixed `test_fuzz` in the tests/ directory as fuzzers and will run them on OSS-Fuzz accodingly. This will make it trivial to add new fuzzers to the continuous fuzzing set up.

Ref:
- https://github.com/google/oss-fuzz/pull/8161
- https://github.com/nedbat/coveragepy/pull/1490

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/tests/test_fuzz_parser.py
+++ b/tests/test_fuzz_parser.py
@@ -1,0 +1,65 @@
+# Licensed under the Apache License: http://www.apache.org/licenses/LICENSE-2.0
+
+"""Fuzz test PythonParser.parse_source
+
+This runs on OSS-Fuzz where coveragepy's set is located at:
+https://github.com/google/oss-fuzz/tree/master/projects/coveragepy
+
+It is configured to be a unit test as well, which makes it easier to test
+during development, e.g. to catch breaking changes.
+
+The goal of the fuzzing by way of OSS-Fuzz is to:
+- Find any uncaught illegitimate exceptions.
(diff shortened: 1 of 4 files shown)


tests/test_fuzz_parser.py
```diff
@@ -0,0 +1,65 @@
...
+
+from coverage.exceptions import NotPython
+from coverage.parser import PythonParser
+
+
+@pytest.mark.parametrize(
+    "data",
+    [
+        b"random_data",
+        b"more random data"
```
nedbat, line 31: I don't understand how atheris works.  What does it do with these two data strings?

ProsperousHeart:
Are this and #1490 still valid?

tests/test_fuzz_parser.py
DavidKorczynski: I only added these strings just to make it such that the fuzzer could be tested with pytest -- to e.g. check that the fuzzer won't break.

The data from `pytest.mark.parametrize(` is not used in the actual fuzzing runs.

PR #1498, opened 2022-12-01
Only rerun the failed tests in CI

webknjaz:
This patch simplifies attempting to execute the tests again on failures in CI. It also limits the scope of the tests being rerun to include only those that failed on the first try. Additionally, it enables the maximum verbosity so that the retry is more useful for troubleshooting.

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -74,23 +74,14 @@ jobs:
           # python -c "import urllib.request as r; exec(r.urlopen('https://bit.ly/pydoctor').read())"
 
       - name: "Run tox for ${{ matrix.python-version }}"
-        continue-on-error: true
-        id: tox1
         run: |
           python -m tox -- -rfsEX
 
       - name: "Retry tox for ${{ matrix.python-version }}"
-        id: tox2
-        if: steps.tox1.outcome == 'failure'
-        run: |
-          python -m tox -- -rfsEX
-
-      - name: "Set status"
-        if: always()
-        run: |
-          if ${{ steps.tox1.outcome != 'success' && steps.tox2.outcome != 'success' }}; then
-             exit 1
-          fi
+        if: failure()
+        run: >-  # `exit 1` makes sure that the job remains red with flaky runs
+          python -m tox -- -rfsEX --lf -vvvvv
+          && exit 1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/testsuite.yml
```diff
@@ -74,23 +74,14 @@ jobs:
...
-
-      - name: "Set status"
-        if: always()
-        run: |
-          if ${{ steps.tox1.outcome != 'success' && steps.tox2.outcome != 'success' }}; then
-             exit 1
-          fi
+        if: failure()
+        run: >-  # `exit 1` makes sure that the job remains red with flaky runs
+          python -m tox -- -rfsEX --lf -vvvvv
```
webknjaz, line 83: @nedbat did I get it right that the posargs are being forwarded to pytest?

.github/workflows/testsuite.yml
```diff
@@ -74,23 +74,14 @@ jobs:
           # python -c "import urllib.request as r; exec(r.urlopen('https://bit.ly/pydoctor').read())"
```
webknjaz, line 74: _(side note)_

@nedbat JFYI https://github.com/marketplace/actions/debugging-with-tmate is a thing.

Additionally, it should be possible to check if the workflow is running in debug mode by checking `${{ secrets.ACTIONS_RUNNER_DEBUG == 'true' || secrets.ACTIONS_STEP_DEBUG == 'true' }}`. So this could be uncommented but guarded. (https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging)

.github/workflows/testsuite.yml
nedbat: yes

webknjaz: Then, this should work as intended. I've started practicing this approach relatively recently and it works well.

nedbat:
Thanks, this is much better than what I had.  I've merged this with a slight edit in commit 535060c9.

webknjaz:
@nedbat there's one unobvious thing in this implementation, though. The retry step may be triggered if any of the earlier steps fail, not just the first pytest invocation. I suppose that may be a bit confusing but shouldn't steal CPU time since it'll probably fail instantly in such a case anyway.

PR #1526, opened 2023-01-04
fix: Handle ast.Match not defined in Python 3.10.0-alpha.1

SylvainDe:
Coverage 7.0.2 introduced some logic relying on ast.Match which a check that the Python version is 3.10 or more (see https://github.com/nedbat/coveragepy/commit/cceadff1d3d33c046042b606d40e01f41e23ec5d )

Unfortunately, on version 3.10.0-alpha.1, ast.Match does not seem to be defined, leading to uncaught exception:
  AttributeError: module 'ast' has no attribute 'Match'

More details can be found on https://github.com/SylvainDe/DidYouMean-Python/issues/59 .

The quickfix is to perform a more precise check on the Python version (another option could be to try to
access ast.Match and catch the exception).

### DIFF
--- a/coverage/parser.py
+++ b/coverage/parser.py
@@ -1079,7 +1079,7 @@ class AstArcAnalyzer:
         exits |= self.add_body_arcs(node.orelse, from_start=from_start)
         return exits
 
-    if sys.version_info > (3, 10, 0, 'alpha', 1):
+    if hasattr(ast, "Match"):  # Python 3.10 and more recent
         def _handle__Match(self, node: ast.Match) -> Set[ArcStart]:
             start = self.line_for_node(node)
             last_start = start
--- a/coverage/parser.py
+++ b/coverage/parser.py
@@ -9,7 +9,6 @@ import ast
 import collections
 import os
 import re
-import sys
 import token
 import tokenize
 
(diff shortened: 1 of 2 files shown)


SylvainDe:
Hello @nedbat ,

- I'm not sure if you'd be interested in a fix for such an unlikely issue but I gave it a try as it is just as quick as writing a bug report
- one of the tests is failing ( https://github.com/nedbat/coveragepy/actions/runs/3836719057/jobs/6531195010 ) but I'm not sure if this is related to my commit.
And obviously, thanks for your worlk!

nedbat:
Wow, this is a deep dive into version compatibility.  Also, it's odd that it's coming from the CI on an explicitly unmaintained repo!

The failures here are due to mypy not understanding your new version check.  I'd have to figure out how best to handle that.

nedbat:
@SylvainDe how important is it to fix this?

SylvainDe:
@nedbat As far as I'm concerned, it's not important at all. As you've noticed, the issue was detected on an unmaintained repo (the CI is kept alive/updated only because in the past it was useful to detect CPython bugs). I understand if it is not worth fixing this in coverage directly (and in that case, I'll just add a tweak in my CI).

This being said, if you do want to fix it anyway, I can rewrite my patch differently with a simple `if hasattr(ast, "Match")` and see how it goes.

SylvainDe:
Hello @nedbat , it looks like the original issue is now fixed. Thanks for whatever was done on your side. I guess this can be closed now :)

Edit: the solution is probably https://github.com/nedbat/coveragepy/commit/4ea850a695e3ab8e42d400dc9dceaebea9246081

PR #1547, opened 2023-01-30
Add purge_files method to CoverageData + unit tests for it

wingware-dev:
This new method makes it possible for users to selectively remove coverage data, and then optionally update it via the existing add_lines and add_arcs methods.

My motivation for this change:  I'm using it in an IDE to track coverage data across edits, in order to invalidate unit test results based on which code was edited.  It produces a pretty good approximation of tests that need to be re-run, useful in applications with massive test suites that take days or longer to run.  Eventually you have to rerun all tests, of course, but this helps catch errors earlier in the development process.  Obviously, there is way more to the implementation of that feature in an IDE, but this method is needed to make it possible.  I'm currently monkey-patching it in, but would prefer it to be part of the CoverageData API.

Of course this new method may also have other uses I'm not thinking of.

### DIFF
--- a/coverage/sqldata.py
+++ b/coverage/sqldata.py
@@ -617,29 +617,29 @@ class CoverageData(AutoReprMixin):
 
     def purge_files(self, filenames, context=None):
         """Purge any existing coverage data for the given `filenames`.
-    
+
         If `context` is given, purge only data associated with that measurement context.
         """
-    
+
         if self._debug.should("dataop"):
             self._debug.write(f"Purging {filenames!r} for context {context}")
         self._start_using()
         with self._connect() as con:
-    
+
             if context is not None:
                 context_id = self._context_id(context)
                 if context_id is None:
                     raise DataError("Unknown context {context}")
             else:
                 context_id = None
-    
+
             if self._has_lines:
                 table = 'line_bits'
             elif self._has_arcs:
                 table = 'arcs'
             else:
                 return
-    
+
             for filename in filenames:
                 file_id = self._file_id(filename, add=False)
                 if file_id is None:
--- a/coverage/sqldata.py
+++ b/coverage/sqldata.py
@@ -615,7 +615,7 @@ class CoverageData(AutoReprMixin):
                     # Set the tracer for this file
                     self.add_file_tracers({filename: plugin_name})
 
-    def purge_files(self, filenames, context=None):
+    def purge_files(self, filenames: Iterable[str], context: Optional[str] = None) -> None:
         """Purge any existing coverage data for the given `filenames`.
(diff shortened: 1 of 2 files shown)


wingware-dev:
Hmm, well, checks all passed (in [14b7a78]) until I merged from master which seems to have an issue under Python 3.9 :-/

nedbat:
Thanks! I'm going to do a little refactoring after merging.

coverage/sqldata.py
```diff
@@ -615,6 +615,42 @@ def touch_files(self, filenames: Collection[str], plugin_name: Optional[str] = N
...
+            elif self._has_arcs:
+                table = 'arcs'
+            else:
+                return
+
+            for filename in filenames:
+                file_id = self._file_id(filename, add=False)
+                if file_id is None:
+                    continue
+                self._file_map.pop(filename, None)
```
nedbat, line 647: After I merged, I noticed this line: if you are only removing one context for the file, then we don't want to remove the file from the file_map, correct?  Why is it important to remove it from the map, and what should we do if it's only a single context being removed?

sdeibel: Good catch.  It looks like in Wing i'm only calling this with context==None and in all cases I immediately set updated data with set_context and add_lines so all the contexts are restored, but on the right new line numbers.  Should I go ahead and just remove the context arg?  I think it was a mistake that came from trying to follow what other API calls support for arguments.  Now I can't really think of a reason to have that argument.

I do remember popping from _file_map was needed or some aspect of the update did not work because measured_files was called.  I'm going to investigate that further since popping here may indeed be the wrong way to solve that part of it.

Another option would be to replace this with a call that purges/replaces data with a new set of lines all in one operation.  My initial idea was the low-level approach was a better match for this API, but I could have been wrong.

nedbat: It looks like the original `cur.execute` call didn't actually remove data because the cursor wasn't iterated. I replaced it with `cur.execute_void` and it had a better effect.

If you don't need the per-context operation, then let's definitely simplify it to only remove the entire file.  I think I'd have it remove the data for the file, but leave the file in the set of measured files, and state that it's possible for a "measured file" to have no data.  Would that work for you?

If we find a need for per-context purging later, we could add it.

wingware-dev: I've implemented this in https://github.com/nedbat/coveragepy/pull/1551 - wasn't sure how to add to this pull request so ended up with a new one.  Hope that is OK...

nedbat:
This is now released as part of [coverage 7.2.0](https://pypi.org/project/coverage/7.2.0).

PR #1550, opened 2023-02-02
(🎁) add py.typed

KotlinIsland:
Type checkers look for this file to determine if type annotations should be used or not. It's required for downstream projects to have access to coverage's type annotations.


nedbat:
Thanks!

nedbat:
This is now released as part of [coverage 7.2.0](https://pypi.org/project/coverage/7.2.0).

PR #1551, opened 2023-02-06
Removed contexts are from purge_files and don't pop from _file_map

wingware-dev:
This works for me and I agree it makes more sense to keep file_ids around so they remain constant over time for the same file.  Please let me know if you would prefer just to remove the second test, which is creating contexts but not adding much anymore, other than just checking that measured_contexts() still has them (sort of obvious / unlikely to be an issue that needs to be tested here specifically).

### DIFF
--- a/coverage/sqldata.py
+++ b/coverage/sqldata.py
@@ -615,23 +615,18 @@ class CoverageData(AutoReprMixin):
                     # Set the tracer for this file
                     self.add_file_tracers({filename: plugin_name})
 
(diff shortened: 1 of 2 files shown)


nedbat:
I'd already started moving some code around in my own branch, so I'll take care of this, thanks.

wingware-dev:
Great, thank you!

nedbat:
This is finished in commit 423fa596.

PR #1560, opened 2023-02-21
fix: don’t measure all third-party packages if source is in third-party location

manueljacob:
There is logic to not measure third-party packages inside configured sources. However, when a (i.e. another) configured source was inside a third-party location, this logic was previously disabled completely.

This caused a problem if a virtual env is set up inside a configured source directory and a configured source package gets installed inside the virtual env. Previously in this case, coverage was measured for all files in the virtual env for the reason described in the previous paragraph.

This commit changes the code to collect all configured source directories inside third-party locations and disable coverage for code in third-party locations only if its not in one of these collected source directories.

### DIFF
--- a/coverage/inorout.py
+++ b/coverage/inorout.py
@@ -262,7 +262,7 @@ class InOrOut:
         # Check if the source we want to measure has been installed as a
         # third-party package.
         # Is the source inside a third-party area?
-        self.source_in_third = False
+        self.source_in_third_paths = set()
         with sys_modules_saved():
             for pkg in self.source_pkgs:
                 try:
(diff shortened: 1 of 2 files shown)


nedbat:
I'd like to understand the problematic scenario better.  Do you have an example of a configuration that was measured incorrectly?

nedbat:
I fixed the tests that were triggering the CodeQL warnings.  We have to figure out what's going on with the failing Windows tests.

manueljacob:
I’ve created an example in a [repository](https://github.com/manueljacob/coverage_source_in_third_example/tree/venv). The problem gets triggered when running `run_tests_in_venv.sh`.

nedbat:
Thanks for the demo, I get it now :)    and thanks for the pull request!

nedbat:
Though I have to ask: if your `example_package` is installed in the venv, why use `source = .` in the coverage settings?

manueljacob:
> Though I have to ask: if your `example_package` is installed in the venv, why use `source = .` in the coverage settings?

In this particular example (where the `test_example_module.py` file is in `.`), to enable coverage measurement for the test file. Is there any other way to achieve this if a to-be-measured file is in the top-level directory and `source_pkgs` needs to be used for other files?

Usually, I’d put tests into a directory called `tests` or `testing`, though. Originally, the problem appeared in a project where I wanted to enable coverage measurement for `setup.py` and some other file that was referenced to from `setup.py`.

Building the package and running the tests are two separate invocations of `coverage run`. Originally, I wanted to specify all sources centrally in the configuration, but this caused the error that is fixed by this pull request during the test coverage run and a `module-not-imported` warning during the build coverage run. What works now is to pass only the package as source to the test coverage run and only `.` as source to the build coverage run. If the error is fixed, is there any reason to prefer one way of specifying the sources over the other?

Another reason why I decided to use `source = .` is that I wanted coverage to report unexecuted files. However, it turns out that this works only for importable files (and files in `.` itself). Is there a better way to make coverage report all unexecuted files?

nedbat:
I fixed the Windows tests on 3.7 and 3.8 in https://github.com/manueljacob/coveragepy/pull/1.

nedbat:
Thanks, I've merged this along with my Windows tweak in 4574ecf128ae51c2b950f6c9cb2486b86f5354e7.

manueljacob:
Thank you for looking into this again! Sorry, I was a bit clueless about the Windows problem, so I didn’t push this any further.

nedbat:
This is now released as part of [coverage 7.2.2](https://pypi.org/project/coverage/7.2.2).

PR #1570, opened 2023-03-03
(🐞) fix: Remove missing type parameter

KotlinIsland:
This missing type parameter is causing errors in some usages


nedbat:
Can you explain what usages are causing errors?

KotlinIsland:
`Coverage` doesn't actually support a `os.PathLike[bytes]` for `config_file`:
```py
a: os.PathLike[bytes] = get_bytes_path()
Coverage(config_file=a)
```
```
...
  File "/coveragepy/coverage/config.py", line 558, in config_files_to_try
    assert isinstance(config_file, str)
AssertionError
```

Also in basedmypy you will get an error if you call a partially typed function, so any usage of `Coverage` report:
```py
Coverage(data_file=path)  # error: Call to incomplete function "Coverage" in typed context  [no-untyped-call]
```

PR #1587, opened 2023-03-16
Put contexts dict in <script> and reveal them using JavaScript

### LINKED ISSUE #1584: coverage html with show_contexts = True may produce a 196MiB file that can't be opened in a browser
**Describe the bug**
In case a file is covered by thousands of contexts, the HTML file produces is too big to be opened in Firefox - the browser simply hangs.

**To Reproduce**
Our codebase's `.coverage` is around 55MiB and looks like this:
```
sqlite> select count(*) from file;
1249
sqlite> select count(*) from context;
26184
sqlite> select count(*) from line_bits;
567073
```
Running `coverage html` with `show_contexts = True` takes a long while.

without show_contexts:

    HTML size = 124MiB
    gen time = 1m30s

with show_contexts:

    HMTL size = 1.5GiB
    gen time = 2m40s

The bulk of the size in generated HTML is occupied by:

    <span class="ctxs">
      <span>python_fname::test_name|run</span>
      ...
    </span>

**Expected behavior / Proposed solution**
Maybe some kind of ad-hoc compression could work?
Replace the span with:
```
<span class="ctxs">
  4,8,15,16,23,42
  ...
</span>
```

Then embed a JS block with a dictionary that maps the numbers above to `python_fname::test_name|run`. On expanding the ctxts span, JS could populate it with content.

I've estimated that at 120 chars per pytest context name and all 26184 contexts included into the JS block, the block will be around 3.14MiB. Which is a lot less than the 192MiB HTML file that Firefox can't open. And that's just for the largest file.
The size of the smaller files could improve. One 3.5MiB file that I looked at has a context repeating 45 times. Compressing it to tar.gz has reduced the size to 292K.

A simple alternative is to add an option to render only the first 10 contexts in HTML. Feels a bit bad to generate a lossy report though.

abo-abo:
Fixes #1584.
The overall size of the generated HTML files has shrunk from 1.5G to 257M.
The size of the largest file down from 196M down to 9.2M.
All files are manageable now with <1s load time in the browser. And maybe 1s reveal time via JavaScript in case of 5k contexts.

### DIFF
--- a/coverage/html.py
+++ b/coverage/html.py
@@ -10,6 +10,8 @@ import json
 import os
 import re
 import shutil
+import json
+from collections import Counter
 
 from dataclasses import dataclass
 from typing import Any, Dict, Iterable, List, Optional, Tuple, TYPE_CHECKING, cast
@@ -367,6 +369,11 @@ class HtmlReporter:
 
         # Write the HTML page for this file.
         file_data = self.datagen.data_for_file(ftr.fr, ftr.analysis)
+
+        contexts=Counter(c for cline in file_data.lines for c in cline.contexts)
+        context_codes = {y: i for (i, y) in enumerate(x[0] for x in contexts.most_common())}
+        contexts_json = json.dumps({v: k for (k, v) in context_codes.items()}, indent=2)
+
         for ldata in file_data.lines:
             # Build the HTML for the line.
             html_parts = []
@@ -379,6 +386,8 @@ class HtmlReporter:
                         f'<span class="{tok_type}">{tok_html}</span>'
                     )
             ldata.html = ''.join(html_parts)
+
+            ldata.context_str = ",".join([str(context_codes[c_context]) for c_context in ldata.context_list])
 
             if ldata.short_annotations:
                 # 202F is NARROW NO-BREAK SPACE.
@@ -412,9 +421,14 @@ class HtmlReporter:
                 )
             ldata.css_class = ' '.join(css_classes) or "pln"
 
+        di = file_data.__dict__
+        if contexts_json == "{}":
+            di["contexts_json"] = None
+        else:
+            di["contexts_json"] = contexts_json
         html_path = os.path.join(self.directory, ftr.html_filename)
         html = self.source_tmpl.render({
-            **file_data.__dict__,
+            **di,
             'prev_html': prev_html,
             'next_html': next_html,
         })
--- a/coverage/html.py
+++ b/coverage/html.py
@@ -10,7 +10,6 @@ import json
 import os
 import re
 import shutil
-import json
 from collections import Counter
 
 from dataclasses import dataclass
@@ -86,6 +85,7 @@ class LineData:
     short_annotations: List[str]
     long_annotations: List[str]
     html: str = ""
+    context_str: Optional[str] = None
     annotate: Optional[str] = None
     annotate_long: Optional[str] = None
     css_class: str = ""
@@ -387,7 +387,8 @@ class HtmlReporter:
                     )
             ldata.html = ''.join(html_parts)
 
-            ldata.context_str = ",".join(str(context_codes[c_context]) for c_context in ldata.context_list)
+            ldata.context_str = ",".join(
+                str(context_codes[c_context]) for c_context in ldata.context_list)
 
             if ldata.short_annotations:
                 # 202F is NARROW NO-BREAK SPACE.
--- a/coverage/html.py
+++ b/coverage/html.py
@@ -370,7 +370,7 @@ class HtmlReporter:
         # Write the HTML page for this file.
         file_data = self.datagen.data_for_file(ftr.fr, ftr.analysis)
 
-        contexts=Counter(c for cline in file_data.lines for c in cline.contexts)
+        contexts = Counter(c for cline in file_data.lines for c in cline.contexts)
         context_codes = {y: i for (i, y) in enumerate(x[0] for x in contexts.most_common())}
         contexts_json = json.dumps({v: k for (k, v) in context_codes.items()}, indent=2)
 
@@ -387,7 +387,7 @@ class HtmlReporter:
                     )
             ldata.html = ''.join(html_parts)
 
-            ldata.context_str = ",".join([str(context_codes[c_context]) for c_context in ldata.context_list])
+            ldata.context_str = ",".join(str(context_codes[c_context]) for c_context in ldata.context_list)
 
             if ldata.short_annotations:
                 # 202F is NARROW NO-BREAK SPACE.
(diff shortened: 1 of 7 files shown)


coverage/html.py
```diff
@@ -367,6 +369,11 @@ def write_html_file(self, ftr: FileToReport, prev_html: str, next_html: str) ->
 
         # Write the HTML page for this file.
         file_data = self.datagen.data_for_file(ftr.fr, ftr.analysis)
+
+        contexts=Counter(c for cline in file_data.lines for c in cline.contexts)
```
nedbat: Suprisingly, my linters don't complain about the lack of spaces around the `=` here...

coverage/html.py
```diff
@@ -380,6 +387,8 @@ def write_html_file(self, ftr: FileToReport, prev_html: str, next_html: str) ->
                     )
             ldata.html = ''.join(html_parts)
 
+            ldata.context_str = ",".join([str(context_codes[c_context]) for c_context in ldata.context_list])
```
nedbat: nit: you don't need the list comprehension inside the `join`, you can just remove the square brackets.

coverage/html.py
```diff
@@ -412,9 +421,14 @@ def write_html_file(self, ftr: FileToReport, prev_html: str, next_html: str) ->
                 )
             ldata.css_class = ' '.join(css_classes) or "pln"
 
+        di = file_data.__dict__
+        if contexts_json == "{}":
+            di["contexts_json"] = None
```
nedbat: Wouldn't it just be easier to use:
```python
file_data["contexts_json"] = None
```
and then not have to create the `di` variable?

coverage/html.py
```diff
@@ -412,9 +421,14 @@ def write_html_file(self, ftr: FileToReport, prev_html: str, next_html: str) ->
                 )
             ldata.css_class = ' '.join(css_classes) or "pln"
 
+        di = file_data.__dict__
+        if contexts_json == "{}":
```
nedbat: It feels a little odd to check for no contexts like this.  Can we use `if not contexts:` instead?

coverage/htmlfiles/coverage_html.js
nedbat, line 1: Please note that I use four-space indents in .js and .html files.  I recommend an editorconfig plugin for your editor to adjust settings automatically.

coverage/htmlfiles/coverage_html.js
```diff
@@ -595,10 +595,31 @@ coverage.wire_up_sticky_header = function () {
...
 document.addEventListener("DOMContentLoaded", () => {
-    if (document.body.classList.contains("indexfile")) {
-        coverage.index_ready();
-    } else {
-        coverage.pyfile_ready();
-    }
+  cboxes = document.querySelectorAll('[id^=ctxs]')
+  cboxes.forEach(function(cbox) {
+    cbox.addEventListener("click", coverage.showContexts)
+  });
```
nedbat: Maybe this code should be inside `pyfile_ready`, since it only applies to those files?

coverage/htmlfiles/coverage_html.js
```diff
@@ -595,10 +595,31 @@ coverage.wire_up_sticky_header = function () {
...
+  span = e.target.nextElementSibling.nextElementSibling;
+  span_text = span.textContent;
+
+  if (/^[0-9,]+$/.test(span_text))
+  {
+    span.textContent = "";
+    span_text.split(",").forEach(function(s) {
+      ctx = contexts[s];
+      span.appendChild(document.createTextNode(ctx));
+      span.appendChild(document.createElement("br"));
```
nedbat: This creates slightly different HTML than the current code, so the styling needs to be adjusted to keep the right-alignment.

nedbat (changes requested):
Thanks so much for doing this.  There are other linting errors to address also.

nedbat:
Now I'm wondering about fancier over-engineered more compact encodings of the contexts, but that can wait (perhaps until never).

abo-abo:
Hi. Thanks for the review. I've addressed the review comments. Even the CSS one somehow. Please check if it's fine: I'm not a frontend guy so I just googled.

coverage/htmlfiles/style.css
```diff
@@ -258,7 +258,7 @@ kbd { border: 1px solid black; border-color: #888 #333 #333 #888; padding: .1em
 
 @media (prefers-color-scheme: dark) { #source p label.ctx { color: #777; } }
 
-#source p .ctxs { display: block; max-height: 0; overflow-y: hidden; transition: all .2s; padding: 0 .5em; font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Ubuntu, Cantarell, "Helvetica Neue", sans-serif; white-space: nowrap; background: #d0e8ff; border-radius: .25em; margin-right: 1.75em; }
+#source p .ctxs { display: block; max-height: 0; overflow-y: hidden; transition: all .2s; padding: 0 .5em; font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Ubuntu, Cantarell, "Helvetica Neue", sans-serif; white-space: nowrap; background: #d0e8ff; border-radius: .25em; margin-right: 1.75em; text-align: right; }
```
nedbat, line 261: This file is generated from a .scss file.  You want to add this setting in place of the `span` clause here: https://github.com/nedbat/coveragepy/blob/master/coverage/htmlfiles/style.scss#L625-L628

nedbat:
I'm surprised you didn't need to update any tests. That tells me that I had no tests of HTML output with contexts!? Yikes.  I can provide one later.

nedbat:
The test failures are because the CSS file has changed.  I can help get those tests passing once you move the setting to the .scss file.

abo-abo:
Hi. I've fixed the tests except `tox lint`. That seems to fail unrelated to the PR?

nedbat:
The lint failures are in your changes in html.py:
```
coverage/html.py:390 C: Line too long (107/100) (line-too-long)
coverage/html.py:13 W: Reimport 'json' (imported line 9) (reimported)
coverage/html.py:13 C: Imports from package json are not grouped (ungrouped-imports)
```
Also, the mypy failure:
```
coverage/html.py:390: error: "LineData" has no attribute "context_str"; maybe "context_list" or "contexts"?  [attr-defined]
```

abo-abo:
Ah, I see it now. Github actions has helpfully hidden the lint results like this:
```
> tox:lint
... weird exit code...
```
So I had to click on `> tox:lint` to see the lint erorrs. Thanks for pointing this out!

nedbat:
I added a test to the master branch that should fail here because the context HTML is different.  We can update the gold files on this branch to make them pass.

nedbat:
To fix the test failures now, you should be able to cd into `tests/gold/html`, and run `make update-gold`.

nedbat (approved):

nedbat:
BTW, in b8498bf9 I made the shortcodes more compact. It shaved another 5% off the size of my contextful HTML reports.

nedbat:
This is now released as part of [coverage 7.2.3](https://pypi.org/project/coverage/7.2.3).

PR #1588, opened 2023-03-20
build: Migrate pytest configuration to pyproject.toml.

neiljp:
This works locally for me, but not all of the tox suite would run since the `.pip` files won't reliably install (yet).

However, the pytest suite runs locally on 3.9 just fine, so pushing this to see if it runs the CI test suite OK.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,32 @@ exclude = """(?x)(
     | ^tests/balance_xdist_plugin\\.py$         # not part of our test suite.
     )"""
 
+[tool.pytest.ini_options]
+addopts = "-q -n auto -p no:legacypath --strict-markers --no-flaky-report -rfEX --failed-first"
+python_classes = "*Test"
+markers = [
+    "expensive: too slow to run during \"make smoke\"",
+]
+
+# How come these warnings are suppressed successfully here, but not in conftest.py??
+filterwarnings = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -28,6 +28,33 @@ exclude = """(?x)(
     | ^tests/balance_xdist_plugin\\.py$         # not part of our test suite.
     )"""
 
+[tool.pytest.ini_options]
+minversion = "6.0"
```
nedbat: What does this do? It wasn't in the setup.cfg.

neiljp: https://docs.pytest.org/en/7.1.x/reference/customize.html#pyproject-toml

neiljp:
Re the balanced_clumps section more generally, I wasn't sure how to know if the plugin is definitely working and configured, if that's what that section is about. Since I'm unsure if the test system would fail to run, or just not balance cleanly (using xdist?) if this config was not found in setup.cfg, but was present in  pyproject.toml.

pyproject.toml
nedbat: I'm not sure we need this setting, since we pin the version of pytest.  I think it would be better to leave it out.

neiljp: Yeah, when I migrated our version, I included it to be on the safe side, but I agree this is likely ok to leave out.

nedbat (approved):

PR #1600, opened 2023-04-02
Fix bug where coverage data is not saved on SIGTERM

### LINKED ISSUE #1599: Coverage data not saved on SIGTERM in 'concurrency = thread' mode
**Describe the bug**

If you have a long-running/daemon process that you're collecting coverage for, e.g. with `coverage run long_running.py`, then even with the `sigterm = true` setting the coverage data is not written if the process is killed with SIGTERM.

If the process is killed with SIGINT instead (either with Ctrl+C or `kill -2 <pid>`) then data is successfully saved, presumably because this manifests as a python `KeyboardInterrupt` exception that allows `finally` blocks to be executed, whereas when SIGTERM is captured by coverage the original SIGTERM handler is invoked without saving data (`finally` blocks do not get run).

**To Reproduce**
    - 3.8, 3.11 (not specific to any version)
1. What version of coverage.py shows the problem? The output of `coverage debug sys` is helpful.
    - latest (master branch)
    - n/a
1. What code shows the problem?  Give us a specific commit of a specific repo that we can check out. If you've already worked around the problem, please provide a commit before that fix.
    - I have a testcase reproducing the problem, will link via PR, also shown below
1. What commands did you run?

```
$cat handler.py
import os, signal

print("START", flush=True)
print("SIGTERM", flush=True)
os.kill(os.getpid(), signal.SIGTERM)
print("NOT HERE", flush=True)
$cat .coveragerc
[run]
sigterm = true
$coverage run handler.py
START
SIGTERM
Terminated
$ls -Al
-rw-r--r-- 1 legaul legaul   77 Apr  3 00:23 .coveragerc
-rw-r--r-- 1 legaul legaul  142 Apr  3 00:49 handler.py
```

**Expected behavior**
Expected coverage data file to be created.

**Additional context**
n/a

LewisGaul:
Fixes https://github.com/nedbat/coveragepy/issues/1599

Failing test prior to the fix:
```
$python -m pytest tests/test_concurrency.py -k Sigterm
bringing up nodes...
...                                                                                                                            [100%] [100%]F [100%]
============================================================== FAILURES ==============================================================
___________________________________________ SigtermTest.test_sigterm_threading_saves_data ____________________________________________
[gw2] linux -- Python 3.8.10 /mnt/c/Users/legaul/repos/coveragepy/venv/bin/python

self = <tests.test_concurrency.SigtermTest object at 0x7f7ffc8f75e0>

    def test_sigterm_threading_saves_data(self) -> None:
        # A terminated process should save its coverage data.
        self.make_file("handler.py", """\
            import os, signal

            print("START", flush=True)
            print("SIGTERM", flush=True)
            os.kill(os.getpid(), signal.SIGTERM)
            print("NOT HERE", flush=True)
            """)
        self.make_file(".coveragerc", """\
            [run]
            # The default concurrency option.
            concurrency = thread
            sigterm = true
            """)
        out = self.run_command("coverage run handler.py")
        assert out == "START\nSIGTERM\nTerminated\n"
        out = self.run_command("coverage report -m")
        expected = "handler.py 5 1 80% 6"
>       assert self.squeezed_lines(out)[2] == expected
E       IndexError: list index out of range

/mnt/c/Users/legaul/repos/coveragepy/tests/test_concurrency.py:774: IndexError
-------------------------------------------------------- Captured stdout call --------------------------------------------------------
START
SIGTERM
Terminated

No data to report.

-------------------------------------------------------- Captured stdout call --------------------------------------------------------
START
SIGTERM
Terminated

No data to report.

-------------------------------------------------------- Captured stdout call --------------------------------------------------------
START
SIGTERM
Terminated

No data to report.

====================================================== short test summary info =======================================================
FAILED tests/test_concurrency.py::SigtermTest::test_sigterm_threading_saves_data - IndexError: list index out of range
1 failed, 3 passed in 3.17s
```

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -653,7 +653,7 @@ class Coverage(TConfigurable):
             self._debug.write(f"{event}: pid: {os.getpid()}, instance: {self!r}")
         if self._started:
             self.stop()
-        if self._auto_save:
+        if self._auto_save or event == "sigterm":
             self.save()
 
     def _on_sigterm(self, signum_unused: int, frame_unused: Optional[FrameType]) -> None:
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -768,7 +768,10 @@ class SigtermTest(CoverageTest):
             sigterm = true
             """)
         out = self.run_command("coverage run handler.py")
-        assert out == "START\nSIGTERM\nTerminated\n"
+        if env.LINUX:
+            assert out == "START\nSIGTERM\nTerminated\n"
+        else:
+            assert out == "START\nSIGTERM\n"
         out = self.run_command("coverage report -m")
         expected = "handler.py 5 1 80% 6"
         assert self.squeezed_lines(out)[2] == expected
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -705,7 +705,7 @@ class SigtermTest(CoverageTest):
     """Tests of our handling of SIGTERM."""
 
     @pytest.mark.parametrize("sigterm", [False, True])
-    def test_sigterm_saves_data(self, sigterm: bool) -> None:
+    def test_sigterm_multiprocessing_saves_data(self, sigterm: bool) -> None:
         # A terminated process should save its coverage data.
         self.make_file("clobbered.py", """\
             import multiprocessing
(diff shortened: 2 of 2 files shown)


coverage/control.py
```diff
@@ -653,7 +653,7 @@ def _atexit(self, event: str = "atexit") -> None:
             self._debug.write(f"{event}: pid: {os.getpid()}, instance: {self!r}")
         if self._started:
             self.stop()
-        if self._auto_save:
+        if self._auto_save or event == "sigterm":
```
nedbat, line 656: I'm curious why `_auto_save` isn't true, but I've only just taken a quick look at this pull request.

LewisGaul: Yeah I wondered that at first - it's being set by the `auto_data` arg to the `Coverage` class, which defaults to `False` and is passed as `True` by `multiproc.ProcessWithCoverage`, i.e. only in the multiprocessing case.

LewisGaul: That said, I'm more than happy with solving this another way - more than anything this proposed change serves as an illustration for what's going wrong!

nedbat: I'm trying to think this through.  On the one hand it seems odd: if `_auto_save` is false, then we won't save the data if the process ends normally, but we will if it's terminated?  But on the other, `sigterm=true` is documented as "to capture data at unusual process ends," which this definitely is. 

Maybe the answer is that `_auto_save` is only needed for the subprocess case.  For main processes, something else will be handling the data saving at normal process termination, and this is fine to put here.

I like that you have a test demonstrating the issue.  `event == "sigterm"` will only happen if `sigterm=true` is in the settings, so clearly the user wanted sigterm handling. 

Let's merge it!

nedbat:
This is now released as part of [coverage 7.2.3](https://pypi.org/project/coverage/7.2.3).

Issue #1523, opened 2023-01-03
Add protocol for paths printed by xml/html subcommands


gaborbernat:
Currently, coverage `xml/html` report prints the path of the file written to stdout:

```
py311: commands[4]> coverage xml -o /home/bernat/git/virtualenv/.tox/coverage.py311.xml
Wrote XML report to /home/bernat/git/virtualenv/.tox/coverage.py311.xml
py311: commands[5]> coverage html -d /home/bernat/git/virtualenv/.tox/py311/tmp/htmlcov --show-contexts --title virtualenv-py311-coverage
Wrote HTML report to /home/bernat/git/virtualenv/.tox/py311/tmp/htmlcov/index.html
```

It would be great if we'd prepend the `file:/` protocol prefix to this path so these links are clickable from the modern terminal, like iTerm or wezterm. I'm ok if it requires a flag to enable this behavior, as I can add that flag to tox.

See:
- https://wezfurlong.org/wezterm/hyperlinks.html#implicit-hyperlinks
- https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

The file prefix enables integrated hyperlinks in all modern terminals.

Perhaps a better and more complicated alternative would be to hide it behind an extra and use rich to print integrated hyperlinks - see https://www.willmcgugan.com/blog/tech/post/real-working-hyperlinks-in-the-terminal-with-rich whenever `tty` is detected; this would allow to avoid the file prefix and also enable it for `coverage` report.

If you consider this feature out of scope for the core project, I'd request some way to achieve this via a plugin :blush:

newbery:
Picking this up

nedbat:
Are you sure you want .xml files to open this way?  And if so, why not .json files?

gaborbernat:
I'm only really interested in the HTML to be honest 😊

nedbat:
This is implemented in 5f31ff9a thanks to #1613.

nedbat:
This is now released as part of [coverage 7.2.4](https://pypi.org/project/coverage/7.2.4).

PR #1613, opened 2023-04-25
Add hyperlink to console output

newbery:
Adds hyperlink as requested in issue #1523


newbery:
Neglected to run tests after final change. Fixing...

nedbat:
@newbery Thanks! Bernat confirmed that this is only needed for HTML.  I'll merge this and then undo the report.py part.

nedbat:
This is now released as part of [coverage 7.2.4](https://pypi.org/project/coverage/7.2.4).

PR #1614, opened 2023-04-27
CONTRIBUTORS.txt: Add 29 names from the Git log

hugovk:
I compared `CONTRIBUTORS.txt` against `git log --pretty="%an" | sort -u --ignore-case` and added missing names.

I ignored some usernames where their name couldn't be easily found.


nedbat:
Thanks, I guess I wasn't as careful as I should have been.

Also: our first Dennis, John, Louis, and Tom, but still no Robert or Bob..!

PR #1641, opened 2023-06-11
Add CITATION.cff

### LINKED ISSUE #1639: Add CITATION.cff
I am unsure of how to properly cite coveragepy in a research article.

I would like to see a CITATION.cff file in the root of the repository.

[GitHub recommends](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files), and integrates support for, [CITATION.cff](https://citation-file-format.github.io/) files. Inclusion of this file would immediately have these benefits:

- Users will have the necessary metadata to properly cite mypy.
- GitHub can automatically generate both APA and BibTeX formatted citations. A "Cite this repository" option will automatically appear on the repo's sidebar when viewed on a desktop browser.
- Reference managers like Zotero will automatically detect the metadata so the user can save the citation to their library.

The other main options are to supply a suggested citation in the README or other documentation, or to have a BibTeX formatted citation file. However, GitHub has better support for CFF formatted citations.

**Additional context**
Examples of popular packages using this are [pandas](https://github.com/pandas-dev/pandas) and [tensorflow](https://github.com/tensorflow/tensorflow).

I would be happy to create an initial version of this and open a PR if there is interest in possibly merging it after refinement.

schackartk:
This pull request adds a CITATION.cff file to the root of the repository. If merged, it resolves #1639.

Some items suggested for further consideration:

- Are there changes you would like to make to the "author" field? We could add a few notable individuals followed by "Contributors to Coverage.py"
- Would you like to include a date? Perhaps the year of first release?
- A version number can be included, but that would require continued maintenance of the file. It is fairly common to leave this off since it adds friction.

Thank you for considering this pull request, and please let me know of any changes you would like to see.

-Ken

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,16 @@
+cff-version: 1.2.0
+title: "Coverage.py: The code coverage tool for Python"
+message: >-
+  If you use this software, please cite it using the metadata from this file.
+type: software
+authors:
+  - name: "Contributors to Coverage.py"
+repository-code: "https://github.com/nedbat/coveragepy"
+url: "https://coverage.readthedocs.io/"
+abstract: >-
+  Coverage.py is a tool for measuring code coverage of Python programs. It monitors your program,
+  noting which parts of the code have been executed, then analyzes the source to identify code
+  that could have been executed but was not.
+  Coverage measurement is typically used to gauge the effectiveness of tests. It can show which
+  parts of your code are being exercised by tests, and which are not.
+license: Apache-2.0
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -4,6 +4,9 @@ message: >-
   If you use this software, please cite it using the metadata from this file.
 type: software
 authors:
+  - family-names: Batchelder
+    given-names: Ned
+    orcid: https://orcid.org/0009-0006-2659-884X
   - name: "Contributors to Coverage.py"
 repository-code: "https://github.com/nedbat/coveragepy"
 url: "https://coverage.readthedocs.io/"
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -8,6 +8,7 @@
 # .tar.gz source distribution would give you everything needed to continue
 # developing the project.  "pip install" will not install many of these files.
 
+include CITATION.cff
 include CONTRIBUTORS.txt
 include CHANGES.rst
 include LICENSE.txt


nedbat:
If we're going to do it, let's do it right! :)   I can see the CFF format gets a bit involved.  If you could put me in as one more contributor, then the generic entry you have will be fine.   I can automate updating the version number if you add an entry for it with the current version number.

I don't know the convention for dates, let me know what the usual practice is.  Development has been 2001-present.

schackartk:
> If we're going to do it, let's do it right! :)

Great!

> If you could put me in as one more contributor, then the generic entry you have will be fine.

So just to clarify, add your name as an author, followed by "Contributors to Coverage.py"? If so, would you also like to include an ORCiD?

>  I can automate updating the version number if you add an entry for it with the current version number.

That's exciting. I will add a field for that.

> I don't know the convention for dates, let me know what the usual practice is. Development has been 2001-present.

I am still learning as well. Of those CFF files that I have found, [tensorflow](https://github.com/tensorflow/tensorflow/blob/master/CITATION.cff) is the only one I see using the `date-released` field. They use the date of first release. However, they do not include the `version` field. So if the `version` is included, it's unclear if the release date should refer to the specific version or the software as a whole.

The CFF documentation states:

> The date the software or data set has been released. Format is 4-digit year, 2-digit month, 2-digit day of month, separated by dashes.

Thinking about it from the perspective of a reader of a research article, I would likely assume the date in the reference list item is the date the software was originally released. 2001 in this case.

nedbat:
> So just to clarify, add your name as an author, followed by "Contributors to Coverage.py"? 

Yes.

> If so, would you also like to include an ORCiD?

Might as well: https://orcid.org/0009-0006-2659-884X

Dates and versions: it feels like the CFF format isn't really expecting an ongoing series of releases, so I think it should either be 1) no version and no date, or 2) latest version and its release date.

schackartk:
That makes sense. So do you have a preference between the two options? Option 1 is certainly more common as far as I have seen.

nedbat:
Let's go with option 1. It's the simplest, and doesn't require automation.

nedbat:
Thanks!

nedbat:
This is now released as part of [coverage 7.3.0](https://pypi.org/project/coverage/7.3.0).

Issue #1407, opened 2022-06-17
--omit="./*.py" omits ALL python files, not only files at the top level


Kache:
**Describe the bug**
My intent is to trace all python files under project root, excepting a couple directories and python files in the top level.

This is close to what I want, but contains python files at the top level:
```sh
coverage run --source=. --omit="tests/*","tests_integ/*" -m pytest
```

but these will omit ALL python files (`CoverageWarning: No data was collected. (no-data-collected)`):
```
coverage run --source=. --omit="tests/*,tests_integ/*,*.py" -m pytest
coverage run --source=. --omit="tests/*,tests_integ/*,./*.py" -m pytest
coverage run --source=. --omit='tests/*,tests_integ/*,*.py' -m pytest
```

**To Reproduce**
How can we reproduce the problem?  Please *be specific*. Don't link to a failing CI job.

Minimal repro:

```
❯ tree -I "__*"
.
├── bar
│   └── bar.py
├── foo.py
└── tests
    └── test_baz.py

2 directories, 3 files
```

```
❯ bat foo.py bar/bar.py tests/test_baz.py
───────┬────────────────────────────────────
       │ File: foo.py   <EMPTY>
───────┴────────────────────────────────────
───────┬────────────────────────────────────
       │ File: bar/bar.py   <EMPTY>
───────┴────────────────────────────────────
───────┬────────────────────────────────────
       │ File: tests/test_baz.py
───────┼────────────────────────────────────
   1   │ def test_foo():
   2   │     assert True
───────┴────────────────────────────────────
```

Then compare `coverage report` after running the two different commands:

```sh
coverage run --source=. -m pytest  # traces coverage for every file
coverage run --source=. --omit="./*.py" -m pytest  # no files traced
```

Answer the questions below:
1. What version of Python are you using? `3.7.6`
1. What version of coverage.py shows the problem? The output of `coverage debug sys` is helpful. `6.4.1`
1. What versions of what packages do you have installed? The output of `pip freeze` is helpful.
1. What code shows the problem?  Give us a specific commit of a specific repo that we can check out. If you've already worked around the problem, please provide a commit before that fix.
1. What commands did you run?

**Expected behavior**
I expected either `*.py` or `./*.py` to only match python files at the top level. In minimal repro above, that means `./foo.py` should have been omitted from coverage trace, but `./bar/bar.py` should not.

**Additional context**
I've also tried "omitting on read" via `coverage report --omit`, but that has the same problem.

[...]

PR #1650, opened 2023-07-11
🐛 Fix trailing single asterisk matching subdirs

webknjaz:
This patch demonstrates a corner case in the path glob matcher.
Specifically, it documents how a single trailing asterisk is supposed
to be treated as opposed to a double asterisk.
With [[1]], a trailing `/*` is interpreted as an equivalent of `/**`.
The commit adds a case that shows that `/*` shouldn't be greedy as
described in the docs [[2]][[3]].

It then goes on to prevent path patterns ending with `/*` from being
greedy and interpreted the same as `/**`. After applying it, that
trailing asterisk only matches one file or directory but not nested
ones.

See also the observations in the bug report ticket [[4]].

[1]: https://github.com/nedbat/coveragepy/commit/ec6205a8de972af6a09453235d02a7ebea6aea8e
[2]: https://coverage.rtfd.io/en/stable/source.html#file-patterns
[3]: https://coverage.rtfd.io/en/7.2.7/migrating.html#migrating-to-coverage-py-7-x
[4]: https://github.com/nedbat/coveragepy/issues/1407#issuecomment-1631085209

### DIFF
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -316,7 +316,7 @@ G2RX_TOKENS = [(re.compile(rx), sub) for rx, sub in [
     (r"\*\*+[^/]+", None),          # Can't have **x
     (r"\*\*/\*\*", None),           # Can't have **/**
     (r"^\*+/", r"(.*[/\\\\])?"),    # ^*/ matches any prefix-slash, or nothing.
-    (r"/\*+$", r"[/\\\\].*"),       # /*$ matches any slash-suffix.
+    (r"/\*\*+$", r"[/\\\\].*"),       # /**$ matches any slash-suffix.
     (r"\*\*/", r"(.*[/\\\\])?"),    # **/ matches any subdirs, including none
     (r"/", r"[/\\\\]"),             # / matches either slash or backslash
     (r"\*", r"[^/\\\\]*"),          # * matches any number of non slash-likes
--- a/tests/test_files.py
+++ b/tests/test_files.py
@@ -348,11 +348,14 @@ class MatcherTest(CoverageTest):
             (self.make_file("sub/file1.py"), True),
             (self.make_file("sub/file2.c"), False),
             (self.make_file("sub2/file3.h"), True),
+            (self.make_file("sub2/sub/file3.h"), False),
             (self.make_file("sub3/file4.py"), True),
             (self.make_file("sub3/file5.c"), False),
+            (self.make_file("sub4/file3.h"), True),
+            (self.make_file("sub4/sub/file3.h"), True),
         ]
-        fnm = GlobMatcher(["*.py", "*/sub2/*"])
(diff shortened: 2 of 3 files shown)


coverage/files.py
```diff
@@ -316,7 +316,7 @@ def sep(s: str) -> str:
     (r"\*\*+[^/]+", None),          # Can't have **x
     (r"\*\*/\*\*", None),           # Can't have **/**
     (r"^\*+/", r"(.*[/\\\\])?"),    # ^*/ matches any prefix-slash, or nothing.
-    (r"/\*+$", r"[/\\\\].*"),       # /*$ matches any slash-suffix.
+    (r"/\*\*+$", r"[/\\\\].*"),       # /**$ matches any slash-suffix.
```
webknjaz, line 319: @nedbat at first I thought I'd just send a test case demonstrating the problem, but then, I thought it should be easy to fix, so I've added this change as well. I don't know if there was a motivation for things to work as they do, but hopefully this can improve the behavior and match it with what's documented…

coverage/files.py
```diff
@@ -316,7 +316,7 @@ def sep(s: str) -> str:
     (r"\*\*+[^/]+", None),          # Can't have **x
     (r"\*\*/\*\*", None),           # Can't have **/**
     (r"^\*+/", r"(.*[/\\\\])?"),    # ^*/ matches any prefix-slash, or nothing.
```
webknjaz, line 318: @nedbat I wonder if this has the same problem with the prefix..

webknjaz:
Hey @nedbat, so do you have an opinion on this? I'm pretty sure it's a bug, but not sure to what extent…

nedbat:
Can you say more about why it's wrong for `/*` to match subdirectories also?

webknjaz:
Mostly because this doesn't match the docs. But also, this mechanism used in several different contexts. How do I omit files in a dir but not subdirs, for example?

webknjaz:
The leading `*/` is probably fine, though.

nedbat:
I see what you mean, but I also see `omit = */site-packages/*` in a few places (one you had to change).  I fear that people are already using the current meaning, and am reluctant to change it again.

If you needed to omit files in a directory, but not further down, you could use `directory/*.py`.

nedbat:
A GitHub search for people using this:  https://github.com/search?type=code&q=omit+site-packages+path%3A**%2F.coveragerc&query=omit+site-packages+path%3A**%2F.coveragerc

webknjaz:
Fair enough. Then, it's better to document how this works both in tests and the docs. I only stumbled across this case while evaluating the update of coveragepy in ansible-test as we pin such things through pip constraints for stability and any version bumps are reviewed manually.
In this case, there was a chance that we might need to generate a version-department config file for coverage as we still pin older versions in some environments.

nedbat:
We're updating tests and docs in #1675.

PR #1651, opened 2023-07-11
🎨 Expand the failed tests rerun by default

webknjaz:
Having `tox-gh` integrated is useful for collapsing bits of the test output in a successful test run. But the re-attempt is trying to include as many verbose details as possible. In this case, what this plugin does is rather unhelpful so the patch removes it.

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -82,9 +82,12 @@ jobs:
 
       - name: "Retry tox for ${{ matrix.python-version }}"
         if: failure()
-        run: |
-          # `exit 1` makes sure that the job remains red with flaky runs
-          python -m tox -- -rfsEX --lf -vvvvv && exit 1
+        # `exit 1` makes sure that the job remains red with flaky runs;
+        # Removing `tox-gh` makes all verbose output easily accessible.
+        run: >-
+          python -m pip uninstall --yes tox-gh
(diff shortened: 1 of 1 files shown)


webknjaz:
Re-run without this patch: https://github.com/nedbat/coveragepy/actions/runs/5524264323/jobs/10076324965#step:6:26.
Re-run with this patch: https://github.com/nedbat/coveragepy/actions/runs/5526039318/jobs/10080303492?pr=1651#step:6:26.

As you can see, the commands being run are immediately visible in the second case. You already need to expand that step if you want to inspect the rerun, and this eliminates the need to locate internal sections that need expanding.
And for cases when you don't need to look into rerun, it doesn't matter.

nedbat:
But removing the plugin means that the tests also run in more environments than they are meant to.  I think it's better to keep the plugin and have to open the twisty.  The latest version of tox-gh now collapses the initial steps, so the test runs are easier to find.

webknjaz:
@nedbat oh, I didn't take that into account because I don't typically use this plugin. Instead, I tend to have a `TOXENV=py` env var set that only runs the default env against the currently active interpreter… I wonder if adding `-vvvvv` to the tox invocation itself would be helpful.

I also noticed that there's a dist reinstall that's happening on the second invocation. The docs for gh-tox show how to do that.

PR #1658, opened 2023-07-22
Add some typing changes required by Mypyc

nickdrozd:
I tried compiling Coverage with Mypyc, the compiler that comes with Mypy. Mypyc is pretty stringent about what it expects, and addressing some of its objections will require some design thought. But these changes are pretty straightforward.

### DIFF
--- a/coverage/env.py
+++ b/coverage/env.py
@@ -53,10 +53,7 @@ class PYBEHAVIOR:
 
     # Is "if not __debug__" optimized away? The exact details have changed
     # across versions.
-    if pep626:
-        optimize_if_not_debug = 1
-    else:
-        optimize_if_not_debug = 2
+    optimize_if_not_debug = 1 if pep626 else 2
 
     # 3.7 changed how functions with only docstrings are numbered.
     docstring_only_function = (not PYPY) and (PYVERSION <= (3, 10))
--- a/coverage/env.py
+++ b/coverage/env.py
@@ -9,7 +9,7 @@ import os
 import platform
 import sys
 
-from typing import Any
+from typing import Any, Final
 from collections.abc import Iterable
 
 # debug_info() at the bottom wants to show all the globals, but not imports.
@@ -148,7 +148,7 @@ class PYBEHAVIOR:
     soft_keywords = (PYVERSION >= (3, 10))
 
     # PEP669 Low Impact Monitoring: https://peps.python.org/pep-0669/
-    pep669 = bool(getattr(sys, "monitoring", None))
+    pep669: Final[bool] = bool(getattr(sys, "monitoring", None))
 
     # Where does frame.f_lasti point when yielding from a generator?
     # It used to point at the YIELD, in 3.13 it points at the RESUME,
--- a/coverage/config.py
+++ b/coverage/config.py
@@ -13,7 +13,7 @@ import os.path
 import re
 
 from typing import (
-    Any, Callable, Union,
+    Any, Callable, Final, Union,
 )
 from collections.abc import Iterable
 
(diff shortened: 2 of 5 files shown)


nickdrozd:
https://github.com/nedbat/coveragepy/issues/1659

nickdrozd:
@nedbat Any interest in this one? If not, I can close it.

KRRT7:
Hi @nickdrozd,

It's been a while since you opened this PR, so I wanted to take a closer look.

Could you clarify the reasoning for integrating these changes directly into Coverage?

Are there tangible benefits to Coverage itself compiling its pure Python source code to C via mypyc?
Have you observed or measured any performance improvements (benchmarks are welcome) that would show that transpiling coverage's pure python code base into C via mypyc would help?

nickdrozd:
Hi @KRRT7. This set of changes makes it possible to successfully run Mypyc as follows:

```sh
mypyc coverage           \
      --exclude cmdline  \
      --exclude control  \
      --exclude env      \
      --exclude files    \
      --exclude parser   \
      --exclude plugin   \
      --exclude sysmon   \
      --exclude tracer
```

The changes are extremely minor, and IMO would be worth doing anyway.

I haven't done any measurements of any kind. Maybe Mypyc won't make much of a difference. On the other hand, Python is really really slow, so maybe it will make a difference. Coverage isn't exactly a performance-critical tool, but it would be nice if it were faster.

Two open questions:

1. How much benefit is there?
2. How much work would it take?

If the benefit is high or the work is low, I think it would be worth pursuing further.

nedbat:
I don't yet understand how mypyc might fit into the coverage.py world, but these typing changes are good, so I'll merge them.

coverage/env.py
```diff
@@ -53,10 +53,7 @@ class PYBEHAVIOR:
 
     # Is "if not __debug__" optimized away? The exact details have changed
     # across versions.
-    if pep626:
-        optimize_if_not_debug = 1
-    else:
-        optimize_if_not_debug = 2
+    optimize_if_not_debug = 1 if pep626 else 2
```
nedbat, line 56: Why was this necessary?

coverage/env.py
```diff
@@ -148,7 +145,7 @@ class PYBEHAVIOR:
     soft_keywords = (PYVERSION >= (3, 10))
 
     # PEP669 Low Impact Monitoring: https://peps.python.org/pep-0669/
-    pep669 = bool(getattr(sys, "monitoring", None))
+    pep669: Final[bool] = bool(getattr(sys, "monitoring", None))
```
nedbat, line 148: Conceptually, all of PYBEHAVIOR should be Final.  How come only this attribute needed Final?

nedbat:
I've also gone ahead and enabled strict mypy checking in commit ea8ae17b5a74d8925a336ae52688f5dfdcd70647.

coverage/env.py
nickdrozd:
```
coverage/env.py:56: error: Unsupported statement in class body
```

In general Mypyc doesn't things being defined in multiple places. Using `if`/`else` means the class var is defined in two places, while using conditional statement means it is defined in just one place.

A more dramatic example is in `files.py`, where the function `actual_path` is defined conditionally:

```python
if env.WINDOWS:

    _ACTUAL_PATH_CACHE: dict[str, str] = {}
    _ACTUAL_PATH_LIST_CACHE: dict[str, list[str]] = {}

    def actual_path(path: str) -> str:
        """Get the actual path of `path`, including the correct case."""
	...

else:
    def actual_path(path: str) -> str:
        """The actual path for non-Windows platforms."""
        return path
```

coverage/env.py
nickdrozd: That's a good question. I think it has to do with the file exclusions I mentioned above. Probably they would all require `Final` if everything were included.

PR #1661, opened 2023-07-25
Add missing MacOS ARM64 wheel for Python 3.12 to build matrix

edgarrmondragon:
Tested locally with

```shell
export CIBW_BUILD=cp312-*
export CIBW_ARCHS=arm64
export CIBW_ENVIRONMENT=PIP_DISABLE_PIP_VERSION_CHECK=1
export CIBW_PRERELEASE_PYTHONS=True
python -m cibuildwheel --output-dir wheelhouse --platform macos
```


nedbat:
Thanks for pointing this out.  The workflow needs to be edited slightly differently, which I've done in commit 6327b7d8.  Thanks.

edgarrmondragon:
oh, sorry about the noise then. should've gone with an issue 😅. Thanks for taking a look!

PR #1675, opened 2023-08-29
Clarify how stars work in file patterns

nedbat:
@webknjaz How does this look?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,8 +20,15 @@ development at the same time, such as 4.5.x and 5.0.
 Unreleased
 ----------
 
-Nothing yet.
+- The semantics of stars in file patterns has been clarified in the docs.  A
+  leading or trailing star matches any number of path components, like a double
+  star would.  This is different than the behavior of a star in the middle of a
+  pattern.  This discrepancy was `identified by Sviatoslav Sydorenko
+  <starbad_>`_, who `provided patient detailed diagnosis <pull 1650_>`_ and
+  graciously agreed to a pragmatic resolution.
 
+.. _starbad: https://github.com/nedbat/coveragepy/issues/1407#issuecomment-1631085209
+.. _pull 1650: https://github.com/nedbat/coveragepy/pull/1650
 
 .. scriv-start-here
 
@@ -32,7 +39,7 @@ Version 7.3.0 — 2023-08-12
 
 - Added a :meth:`.Coverage.collect` context manager to start and stop coverage
   data collection.
-  
+
 - Dropped support for Python 3.7.
 
 - Fix: in unusual circumstances, SQLite cannot be set to asynchronous mode.
--- a/doc/source.rst
+++ b/doc/source.rst
@@ -178,15 +178,67 @@ individual source lines.  See :ref:`excluding` for details.
 File patterns
 -------------
 
-File path patterns are used for include and omit, and for combining path
-remapping.  They follow common shell syntax:
-
-- ``*`` matches any number of file name characters, not including the directory
-  separator.
+File path patterns are used for :ref:`include <config_run_include>` and
+:ref:`omit <config_run_omit>`, and for :ref:`combining path remapping
+<cmd_combine_remapping>`.  They follow common shell syntax:
 
 - ``?`` matches a single file name character.
 
-- ``**`` matches any number of nested directory names, including none.
+- ``*`` matches any number of file name characters, not including the directory
+  separator.  As a special case, if a pattern starts with ``*/``, it is treated
+  as ``**/``, and if a pattern ends with ``/*``, it is treated as ``/**``.
+
+- ``**`` matches any number of nested directory names, including none. It must
+  be used as a full component of the path, not as part of a word: ``/**/`` is
+  allowed, but ``/a**/`` is not.
 
 - Both ``/`` and ``\`` will match either a slash or a backslash, to make
   cross-platform matching easier.
+
+- A pattern with no directory separators matches the file name in any
+  directory.
+
+Some examples:
+
+.. list-table::
+    :widths: 20 20 20
+    :header-rows: 1
+
+    * - Pattern
+      - Matches
+      - Doesn't Match
+    * - ``a*.py``
+      - | anything.py
+        | sub1/sub2/another.py
+      - | cat.py
+    * - ``sub/*/*.py``
+      - | sub/a/main.py
+        | sub/b/another.py
+      - | sub/foo.py
+        | sub/m1/m2/foo.py
+    * - ``sub/**/*.py``
+      - | sub/something.py
+        | sub/a/main.py
+        | sub/b/another.py
+        | sub/m1/m2/foo.py
+      - | sub1/anything.py
+        | sub1/more/code/main.py
+    * - ``*/sub/*``
+      - | some/where/sub/more/something.py
+        | sub/hello.py
+      - | sub1/anything.py
+    * - ``*/sub*/*``
+      - | some/where/sub/more/something.py
+        | sub/hello.py
+        | sub1/anything.py
+      - | some/more/something.py
+    * - ``*/*sub/test_*.py``
+      - | some/where/sub/test_everything.py
+        | moresub/test_things.py
+      - | some/where/sub/more/test_everything.py
+        | more/test_things.py
+    * - ``*/*sub/*sub/**``
+      - | sub/sub/something.py
+        | asub/bsub/more/thing.py
+        | code/sub/sub/code.py
+      - | sub/something.py
--- a/tests/test_files.py
+++ b/tests/test_files.py
@@ -348,6 +348,7 @@ class MatcherTest(CoverageTest):
             (self.make_file("sub/file1.py"), True),
             (self.make_file("sub/file2.c"), False),
             (self.make_file("sub2/file3.h"), True),
+            (self.make_file("sub2/sub/file3.h"), True),
             (self.make_file("sub3/file4.py"), True),
             (self.make_file("sub3/file5.c"), False),
         ]
--- a/tests/test_files.py
+++ b/tests/test_files.py
@@ -235,8 +235,23 @@ def globs_to_regex_params(
         ),
         globs_to_regex_params(
             ["*/foo"], case_insensitive=False, partial=True,
-            matches=["abc/foo/hi.py", "foo/hi.py"],
+            matches=["abc/foo/hi.py", "foo/hi.py", "abc/def/foo/hi.py"],
             nomatches=["abc/xfoo/hi.py"],
+        ),
+        globs_to_regex_params(
+            ["*c/foo"], case_insensitive=False, partial=True,
+            matches=["abc/foo/hi.py"],
+            nomatches=["abc/xfoo/hi.py", "foo/hi.py", "def/abc/foo/hi.py"],
+        ),
+        globs_to_regex_params(
+            ["foo/x*"], case_insensitive=False, partial=True,
+            matches=["foo/x", "foo/xhi.py", "foo/x/hi.py"],
+            nomatches=[],
+        ),
+        globs_to_regex_params(
+            ["foo/x*"], case_insensitive=False, partial=False,
+            matches=["foo/x", "foo/xhi.py"],
+            nomatches=["foo/x/hi.py"],
         ),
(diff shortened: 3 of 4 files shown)


CHANGES.rst
```diff
@@ -20,8 +20,15 @@ development at the same time, such as 4.5.x and 5.0.
...
-
+- The semantics of stars in file patterns has been clarified in the docs.  A
+  leading or trailing star component matches any number of path components,
+  like a double star would.  This is different than the behavior of a star in
+  the middle of a pattern.  This discrepancy was `identified by Sviatoslav
+  Sydorenko <starbad_>`_, who `provided patient detailed diagnosis
+  <pull 1650_>`_ and graciously agreed to a pragmatic resolution.
+
+.. _starbad: https://github.com/nedbat/coveragepy/issues/1407#issuecomment-1631085209
+.. _issue 1646: https://github.com/nedbat/coveragepy/issues/1646
```
webknjaz: This will fix the docs build crash @ https://github.com/nedbat/coveragepy/actions/runs/6018518753/job/16326834195?pr=1675#step:5:33
```suggestion
.. _pull 1650: https://github.com/nedbat/coveragepy/pull/1650
```

nedbat: D'oh!

nedbat: fixed.

tests/test_files.py
```diff
@@ -235,9 +235,24 @@ def globs_to_regex_params(
         ),
         globs_to_regex_params(
             ["*/foo"], case_insensitive=False, partial=True,
-            matches=["abc/foo/hi.py", "foo/hi.py"],
+            matches=["abc/foo/hi.py", "foo/hi.py", "abc/def/foo/hi.py"],
             nomatches=["abc/xfoo/hi.py"],
         ),
+        globs_to_regex_params(
+            ["*c/foo"], case_insensitive=False, partial=True,
+            matches=["abc/foo/hi.py"],
```
webknjaz, line 243: Looks like this will also work:
```suggestion
            matches=["abc/foo/hi.py", "stuff/abc/foo/hi.py"],
```

webknjaz: Oh wait, I see that a similar case is already in nomatches.. Let me reread the new doc explanation.

nedbat: It doesn't match. The next line includes `nomatches=["def/abc/foo/hi.py"]`

doc/source.rst
```diff
@@ -181,12 +181,16 @@ File patterns
...
-  separator.
-
 - ``?`` matches a single file name character.
 
-- ``**`` matches any number of nested directory names, including none.
+- ``*`` as part of a path component matches any number of file name characters,
+  not including the directory separator: one path component.
+
+- ``*`` as the first or last path component matches anything including
+  directory separators: any number of path components, including none, like
```
webknjaz: So this kinda implies that `"*thing/stuff/"` would match `"some/xthing/stuff/"` because it's unclear what a “path component” refers to.
I think, this is because of the ambiguity — the match expression with a wildcard looks like `"*thing/stuff/"` and so I read this as `"*thing"` is the first path component, and it matches things like `[a-zA-Z0-9/]` basically.
So while you must've been thinking of the path of a discovered file during the filtering process, the reader is likely to think of what's in front of them — a file path pattern.

Maybe, we could rephrase this by specifically pointing out the “if a path pattern starts with `*/`, this is equivalent to `**/`”? WDYT?

nedbat: Yeah, I had a few tries at how to explain this, and didn't love any of them.  I think your idea is a good one.

doc/source.rst
```diff
@@ -181,12 +181,16 @@ File patterns
 File path patterns are used for include and omit, and for combining path
 remapping.  They follow common shell syntax:
 
-- ``*`` matches any number of file name characters, not including the directory
-  separator.
-
 - ``?`` matches a single file name character.
 
-- ``**`` matches any number of nested directory names, including none.
+- ``*`` as part of a path component matches any number of file name characters,
```
webknjaz: Maybe, we could clarify the distinction between "part of" and "the entire match pattern component" through a concrete example?

tests/test_files.py
```diff
@@ -235,9 +235,24 @@ def globs_to_regex_params(
...
+        ),
+        globs_to_regex_params(
+            ["foo/x*"], case_insensitive=False, partial=True,
+            matches=["foo/x", "foo/xhi.py", "foo/x/hi.py"],
+            nomatches=[],
+        ),
+        globs_to_regex_params(
+            ["foo/x*"], case_insensitive=False, partial=False,
+            matches=["foo/x", "foo/xhi.py"],
+            nomatches=["foo/x/hi.py"],
```
webknjaz, line 254: Even though, it's there in the source code, I forgot that this wouldn't match since some time has passed since I first opened my PR. So when I read the new explanation, I assumed this would match too. I think that adding more specific examples to the docs would really help to disambiguate them.

CHANGES.rst
```diff
@@ -20,8 +20,15 @@ development at the same time, such as 4.5.x and 5.0.
 Unreleased
 ----------
 
-Nothing yet.
+- The semantics of stars in file patterns has been clarified in the docs.  A
+  leading or trailing star component matches any number of path components,
+  like a double star would.  This is different than the behavior of a star in
+  the middle of a pattern.  This discrepancy was `identified by Sviatoslav
+  Sydorenko <starbad_>`_, who `provided patient detailed diagnosis
```
webknjaz: On a somewhat unrelated note, I wanted to share one practice I came up with for crediting people in changelogs in projects using Towncrier fragments. I've added a `:user:` RST role to Sphinx (example: https://github.com/pypa/setuptools/blob/1ef36f2/docs/conf.py#L138) and encourage the contributors to add bylines with that role. It expands into a GitHub Sponsors link. When it's not activated, GH makes a redirect to the normal user page.
I borrowed part of the idea from Tox — the only link GH profiles. Here's what it looks like: https://setuptools.pypa.io/en/latest/history.html#v67-0-0.
I use it in some of the projects I maintain or just contribute to.

I figured, if this document is targeted at being consumed by Sphinx and not other tools/systems, you might want to implement something similar in the future.

doc/source.rst
```diff
@@ -181,11 +181,12 @@ File patterns
...
 remapping.  They follow common shell syntax:
 
-- ``*`` matches any number of file name characters, not including the directory
-  separator.
-
 - ``?`` matches a single file name character.
 
+- ``*`` matches any number of file name characters, not including the directory
+  separator.  As a special case, if a pattern starts with ``*/``, it is treated
+  as ``**/``, and if a pattern ends with ``/*``, it is treated as ``/**`.
```
webknjaz, line 188: Having reviewed other places, I'm starting to get a feeling that things like `x/**z/y` and `s/x**z/y` won't work the way it's implied here either. How about adding more examples to this as well?

CHANGES.rst
nedbat: That is a cool idea, I will look into it.  This changelog becomes part of the GitHub releases, so I have to be careful about unusual features.

doc/source.rst
```diff
@@ -181,11 +181,12 @@ File patterns
...
+
 - ``**`` matches any number of nested directory names, including none.
```
webknjaz: Having reviewed other places, I'm starting to get a feeling that things like `x/**z/y` and `s/x**z/y` won't work the way it's implied here either. How about adding more examples to this as well?

doc/source.rst
webknjaz:
```suggestion
  as ``**/``, and if a pattern ends with ``/*``, it is treated as ``/**``.
```

tests/test_files.py
webknjaz: Yeah, I saw that after writing the comment.

doc/source.rst
nedbat: Fixed.

nedbat (commented):
x

doc/source.rst
nedbat: `x/**z/y` and `s/x**z/y` are refused as invalid.

nedbat:
I hear what you are saying about examples in the docs.  I'll have to think about how to do that.

CHANGES.rst
webknjaz: Yeah, that's one pain point. You need to convert to GitHub-Flavoured Markdown for GH Releases. I used pandoc for this in some workflows. Though, since pandoc doesn't know of Sphinx roles (only about the docutils' ones), I had to arm with regexes and pre-process those files via `sed`.

doc/source.rst
webknjaz: Ah, sounds reasonable. Is there a test for this? Should this be mentioned in a "non-working example"?

webknjaz:
Thanks for picking this up! I've been meaning to add some docs myself, but you got to doing this first.
Short of the examples, I think the PR is on the right track towards clarity!

webknjaz:
@nedbat so I wanted to preview how the PR renders and realized that this repository doesn't have PR builds enabled on RTD.
Assuming, this isn't on purpose, could you enable this feature? It's just a single checkbox at the bottom of the first section on this page: https://readthedocs.org/dashboard/coverage/advanced/.
Ref: https://docs.readthedocs.io/en/stable/guides/pull-requests.html.

nedbat:
I've added some examples.  Here's the rendered page: https://coverage.readthedocs.io/en/nedbat-finish-1650/source.html#file-patterns

webknjaz:
@nedbat I see… Though, it's not the same as PR builds. This only works for your personal branches created upstream. PR builds work for PRs coming from forks submitted by anybody.

nedbat:
Right, I don't have pull requests configured to build docs on readthedocs, I created this one manually.

webknjaz:
Is that because you don't want to see those build statuses in PRs?

nedbat:
It's because I don't usually use pull requests myself, and I don't get enough pull requests that update docs in significant ways, so it's never been important to set up.

webknjaz:
Fair enough. Though, from the contributor point, it's definitely useful. I'm so used to having this available that it even confused me a little. I think nowadays, all the new projects have it enabled by default and the only ones that don't are old projects that were created before 2018.

doc/source.rst
```diff
@@ -181,12 +181,46 @@ File patterns
 File path patterns are used for include and omit, and for combining path
 remapping.  They follow common shell syntax:
```
webknjaz: This seems like a good place to reference the corresponding features:
```suggestion
File path patterns are used for :ref:`include <config_run_include>` and
:ref:`omit <config_run_omit>`, and for :ref:`combining path remapping
<cmd_combine_remapping>`.  They follow common shell syntax:
```

webknjaz: _Shameless plug:_ I looked up the references through my little static site generator for intersphinx-enabled docs: https://webknjaz.github.io/intersphinx-untangled/coverage.rtfd.io/

webknjaz:
> I've added some examples.

Can we also have examples for `*/sub*/*`, `*/*sub/test_*.py`, `*/*sub/*sub/**`. And maybe a separate one under an `.. attention::` admonition for all the things that are supposed to error out?

doc/source.rst
webknjaz: Should this also clarify something like “must not be combined with any other characters in the same path frament of the match pattern”?

webknjaz: Out of curiosity, what happens with `"***/**"`?

webknjaz:
Review round complete.

doc/source.rst
nedbat: done.

nedbat:
I've enabled PR builds on readthedocs.

nedbat:
Thanks for all the help!

doc/source.rst
webknjaz: Did you force-push on top after that?

nedbat: I don't know what you mean.  The links are now on master.

webknjaz: You're right, sorry about the noise! GitHub must've been showing me outdated diffs due to its infamous eventual consistency. And then, I've confused myself even more by clicking on the readme badge. Now I realize that the badge doesn't lead to the latest docs as it often happens in other projects.
Having understood that, I checked the correct place and I do indeed see the final change.

PR #1681, opened 2023-09-18
'data' is plural

hiramf:
The word "data" here should be treated as plural


nedbat:
I appreciate the pull request and the Latin origins, but to my ears, "is" sounds better.

Issue #1684, opened 2023-09-25
exclude_lines in .coveragerc not being respected


tingilee:
**Describe the bug**

exclude_lines in .coveragerc are not being respected for excluding `import` statements and `pragma: no cover`. 

**To Reproduce**
How can we reproduce the problem?  Please *be specific*. Don't link to a failing CI job. Answer the questions below:
1. What version of Python are you using? 3.8.17
1. What version of coverage.py shows the problem? 
```
coverage debug sys
-- sys -------------------------------------------------------
               coverage_version: 7.2.1
                coverage_module: /opt/homebrew/lib/python3.8/site-packages/coverage/__init__.py
                         tracer: -none-
                        CTracer: available
           plugins.file_tracers: -none-
            plugins.configurers: -none-
      plugins.context_switchers: -none-
              configs_attempted: .coveragerc
                   configs_read: /Users/jacqueline.lee/jkl/prototype/.coveragerc
                    config_file: /Users/jacqueline.lee/jkl/prototype/.coveragerc
                config_contents: b'[run]\nrelative_files = True\n# omit =\n#     # omit anything in a .local directory anywhere\n#     */.local/*\n#     # omit everything in /usr\n#     /usr/*\n#     /opt/**/*.py\n\nomit=*/site-packages/*,*/tests/*,*/.eggs/*\n\nsource =\n    /Users/jacqueline.lee/jkl/prototype/*\n\n\n[report]\n# number of digits after the decimal point to display for reported coverage percentages\nprecision = 2\n\nexclude_lines =\n    # skip any line with a `pass` (such as may be used for @abstractmethod or @suffixed_method)\n    pass\n    # skip any line with import\n    import\n\n    # standard pragma\n    pragma: no cover\n    #\\s*(pragma|PRAGMA)[:\\s]?\\s*(no|NO)\\s*(branch|BRANCH)\n\n    # skip blocks reserved for type checking\n    if TYPE_CHECKING:\n\n\n# Any line of your source code containing a match for one of these regexes is excluded from being reported as missing.\n# You can exclude lines introducing blocks, and the entire block is excluded. If you exclude a def line or decorator line, the entire function is excluded.\n# https://coverage.readthedocs.io/en/latest/excluding.html#excluding\n\nskip_empty = True'
                      data_file: -none-
                         python: 3.8.17 (default, Jun  6 2023, 17:41:59) [Clang 14.0.3 (clang-1403.0.22.14.1)]
[...]

PR #1685, opened 2023-09-27
[lcov] Exclude executed lines respected exclude configs

tingilee:
`coverage lcov` currently does not respect `exclude_lines`, this PR introduces changes so `exclude_lines` will not show up as hits or misses

Tested via following instructions in https://github.com/nedbat/coveragepy/issues/1684 and running the command:
```
python3.8 ~/robinhood/robinhood/coveragepy/__main__.py run --debug=config --rcfile=/Users/jacqueline.lee/jkl/prototype/.coveragerc --append --branch -m pytest ~/jkl/prototype/ && python3.8 ~/robinhood/robinhood/coveragepy/__main__.py lcov   --rcfile=/Users/jacqueline.lee/jkl/prototype/.coveragerc

### DIFF
--- a/coverage/lcovreport.py
+++ b/coverage/lcovreport.py
@@ -62,8 +62,11 @@ class LcovReporter:
         outfile.write("TN:\n")
         outfile.write(f"SF:{fr.relative_filename()}\n")
         source_lines = fr.source().splitlines()
-
+        sorted_excluded = sorted(analysis.excluded)
         for covered in sorted(analysis.executed):
+            if covered in sorted_excluded:
+                # Do not report excluded as executed
+                continue
             # Note: Coverage.py currently only supports checking *if* a line
             # has been executed, not how many times, so we set this to 1 for
             # nice output even if it's technically incorrect.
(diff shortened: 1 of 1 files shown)


nedbat:
Wow, thanks! I'm a bit surprised no one has mentioned this problem before... lcov support has been in for more than 1.5 years!

Would you be able to include some tests? If not, that's OK, I can add them after the merge.  Let me know.

You mentioned patching 6.4.1.  I don't do that; would you be able to make the pull request based on master instead?  Thanks.

tingilee:
I can certainly add some tests if you can provide some guidance, or I'm happy to leave the testing out of this diff and let you add the tests later. 

This PR is already created to merge into master. In addition, is it possible to get this commit into 6.4.5 release? Our monorepo needs this feature and upgrading to 7.x leads to some weird test errors that I have yet to root cause.

nedbat:
I don't usually put fixes on older versions.  Ideally, we'd figure out what's going wrong with your tests under the latest coverage.py.  Another option is for you to use this fix from your own fork.  Either way, I'd like to get the fix on master.

tingilee:
Sounds great, thank you!

nedbat:
@bradb423 Any opinions on this? Is it right to keep lines out of the .lcov result file when they are excluded in the settings?

bradb423:
@nedbat , Yeah this seems like a good idea to me. I think the results of the Lcov should match the settings :)

nedbat:
This is now released as part of [coverage 7.3.2](https://pypi.org/project/coverage/7.3.2).

nedbat:
This is now released as part of [coverage 7.3.2](https://pypi.org/project/coverage/7.3.2).

PR #1705, opened 2023-11-30
Fix handling of multi-line function declarations

### LINKED ISSUE #684: Line break within function declaration disables exclusion of the function body
`.coveragerc` file:

```
[report]
exclude_lines =
    long_function_name.+
```

Code:
```python
def long_function_name(some_long_argument,  another_long_argument):
    pass
```
This setup works and results in the following report:

```
Name      Stmts   Miss  Cover   Missing
---------------------------------------
file.py       0      0   100%
```
However, if a line break is introduced between arguments and the second line gets indented in compliance to the PEP-8:
```python

def long_function_name(some_long_argument,
                       another_long_argument):
    pass
```
Then the body isn't excluded anymore:

```
Name      Stmts   Miss  Cover   Missing
---------------------------------------
file.py       1      1     0%   4
```

I'm not really sure if this is a bug, but I haven't found a way to bypass this behaviour - I would like the whole function excluded, regardless of the formatting of function arguments. Is this possible?

kajakaj:
Resolves #684
  
The cause of this problem is the method of checking whether a particular clause should be excluded. In a file `parser.py` in function `_raw_parse` (line 161) only the line that ends with `:` is checked for excluded expressions, not all the lines in a statement that ends with a colon. The solution is to add checking if at least one line in the multi-line statement ended with `:` has to be excluded and if that is the case - to exclude that clause. Multi-line statements are added after a token `\n`, so after checking for excluded clauses in one-line statements. Because of that, checking if a clause after multi-line statement should be excluded was added starting from the line 193.
Tests with different use cases were added.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,11 +20,18 @@ development at the same time, such as 4.5.x and 5.0.
 Unreleased
 ----------
 
+- Fix: function definitions with multi-line signatures can now be excluded by
+  matching any of the lines, closing `issue 684`_.  Thanks, `Jan Rusak and
+  others <pull 1705_>`_.
+
 - Added new :ref:`debug options <cmd_run_debug>`:
 
   - ``pytest`` writes the pytest test name into the debug output.
 
   - ``dataop2`` writes the full data being added to CoverageData objects.
+
+.. _issue 684: https://github.com/nedbat/coveragepy/issues/684
+.. _pull 1705: https://github.com/nedbat/coveragepy/pull/1705
 
 
 .. scriv-start-here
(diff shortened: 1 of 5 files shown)


nedbat:
Thanks, this looks promising.  I will take a look soon.

nedbat:
I'm going to merge this. I've thanked Jan Rusak in the changelog and contributors file.  I'll gladly add other names, but I'm not sure what they are! :)

kajakaj:
Thanks - we did it together: Jan Rusak, Maciej Kowalczyk and Joanna Ejzel.

nedbat:
Thanks again, I added the names in 470e086f2961169e0998737157d99924c19bc8e4

nedbat:
This is now released as part of [coverage 7.3.3](https://pypi.org/project/coverage/7.3.3).

PR #1717, opened 2023-12-18
Make so line number not copied

enrh:
Within the htmlcov output, there are line numbers for the code. This update makes so when you highlight the code the line numbers are not included.

image

becomes

image

### DIFF
--- a/coverage/htmlfiles/style.css
+++ b/coverage/htmlfiles/style.css
@@ -148,7 +148,7 @@ kbd { border: 1px solid black; border-color: #888 #333 #333 #888; padding: .1em 
 
 #source p * { box-sizing: border-box; }
 
-#source p .n { float: left; text-align: right; width: 3.5rem; box-sizing: border-box; margin-left: -3.5rem; padding-right: 1em; color: #999; }
+#source p .n { float: left; text-align: right; width: 3.5rem; box-sizing: border-box; margin-left: -3.5rem; padding-right: 1em; color: #999; user-select: none; }
 
 @media (prefers-color-scheme: dark) { #source p .n { color: #777; } }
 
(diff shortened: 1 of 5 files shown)


enrh:
Hey Ned,

I saw in https://coverage.readthedocs.io/en/latest/contributing.html that you prefer if people run their idea by you first, but I suspect that is for more complicated changes. I hope that is okay!

enrh:
Dang it! I will work through the errors :)

nedbat:
Thanks for this.  Small changes can be very useful.  The .css files are generated from .scss files, so the edits should start there.  Maybe I need to add more to the .css files to alert people, or to the contributing docs.

enrh:
> Maybe I need to add more to the .css files to alert people, or to the contributing docs.

Actually I think it's reasonable that I should have noticed `style.scss` :) I have updated and run `make css` for good measure, and it did not change the files I had updated by hand.

enrh:
> > Maybe I need to add more to the .css files to alert people, or to the contributing docs.
> 
> Actually I think it's reasonable that I should have noticed `style.scss` :) I have updated and run `make css` for good measure, and it did not change the files I had updated by hand.

Also there is this
```
coverage/htmlfiles/style.css:/* Don't edit this .css file. Edit the .scss file instead! */
```

enrh:
I believe the checks failed because of a GitHub backend timeout or something unrelated to the code. Are you able to restart or would you prefer I make a faux commit and then undo the commit?

nedbat:
Thanks!

enrh:
You are welcome! Thank you!

nedbat:
This is now released as part of [coverage 7.3.4](https://pypi.org/project/coverage/7.3.4).

PR #1727, opened 2024-01-05
Remove redundant code for EOL Python 3.7

hugovk:
Python 3.7 support was dropped in August 2023 in https://github.com/nedbat/coveragepy/releases/tag/7.3.0 so we can remove some old compatibility checks.

### DIFF
--- a/coverage/env.py
+++ b/coverage/env.py
@@ -60,13 +60,12 @@ class PYBEHAVIOR:
         optimize_if_not_debug = 2
 
     # 3.7 changed how functions with only docstrings are numbered.
-    docstring_only_function = (not PYPY) and ((3, 7, 0, "beta", 5) <= PYVERSION <= (3, 10))
+    docstring_only_function = (not PYPY) and (PYVERSION <= (3, 10))
 
     # When a break/continue/return statement in a try block jumps to a finally
-    # block, does the finally block do the break/continue/return (pre-3.8), or
-    # does the finally jump back to the break/continue/return (3.8) to do the
-    # work?
-    finally_jumps_back = ((3, 8) <= PYVERSION < (3, 10))
+    # block, does the finally jump back to the break/continue/return (pre-3.10)
+    # to do the work?
+    finally_jumps_back = (PYVERSION < (3, 10))
 
     # CPython 3.11 now jumps to the decorator line again while executing
     # the decorator.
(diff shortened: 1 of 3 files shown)


nedbat:
Thanks.  Part of me wants to keep the old stuff in show_pyc as a possibly generally useful tool, but you are right, I shouldn't hold onto it.  How did you find these bits?

hugovk:
I have this hacky function in my dotfiles:

```sh
function grepv () {
if [ -z "$1" ]
  then
    # No argument supplied
    git grep -nrE "version_info|python_version_tuple|python_version|PY_MAJOR_VERSION|PY_VERSION_HEX|hexversion|sys\.version" . "${@:3}"
else
#   git grep -nrIE --colour=always "(py(thon)?( )?)?$1(\.?|(, ?))$2|0x0${1}0${2}0000" . "${@:3}"
  git grep -nrIE "(py(thon)?( )?)?$1(\.?|(, ?))$2|0x0${1}0${2}0000" . "${@:3}"
fi
git grep -n "universal *= *1"
git grep -n python_requires
git grep -n requires-python
}
```

Then run plain `grepv` without args to see lots of places where things like `sys.version_info` are used, or `grepv 3 7` for things that might look like a 3.7 version number. Lots of false positives, but that's okay, it's just for glancing over!

---

https://github.com/asottile/pyupgrade is also a useful too that can clean up some version checks too:

```
pip install pyupgrade
pyupgrade **/*.py --py38-plus
```

But looks like it wouldn't find these. It could be useful to upgrade lots of type annotations (`Tuple` -> `tuple` etc).

nedbat:
Very cool, thanks.  I've used pyupgrade, it's great:
```
% git log --grep=pyupgrade --format='%h %ad | %s' --date=short
  6205c9fe 2023-08-30 | style: pyupgrade --py38-plus --keep-runtime-typing
  df79a639 2021-05-03 | refactor: remove redundant Python 2 code (#1155)
  14449eed 2021-05-01 | refactor: pyupgrade --py36-plus *.py
  ddf5ba8c 2021-05-01 | refactor: pyupgrade --py36-plus tests/**.py
  4c4ba2e0 2021-05-01 | refactor: pyupgrade --py36-plus coverage/*.py
```

nedbat:
Oh, and one of those was you:
```
 df79a639 2021-05-03 | refactor: remove redundant Python 2 code (#1155)
```

hugovk:
I see you don't have pre-commit in this repo, but I have pyupgrade in there for most of my projects.

PR #1728, opened 2024-01-05
CI: Only run dependency review for upstream

hugovk:
This workflow fails when running on forks, which by default have it disabled.

> Run actions/dependency-review-action@v3
Error: Dependency review is not supported on this repository. Please ensure that Dependency graph is enabled along with GitHub Advanced Security on private repositories, see https://github.com/hugovk/coveragepy/settings/security_analysis

https://github.com/hugovk/coveragepy/actions/runs/7418996024

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -19,6 +19,7 @@ permissions:
 
 jobs:
   dependency-review:
+    if: github.repository_owner == 'nedbat'
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout Repository'


nedbat:
This is one of those GitHub-suggested actions.  I guess you could enable it in your fork, but no one would of course, so we should disable it to prevent the failure.

TBH, I'm not exactly sure what this action is doing for me.  Looking at a recent run of it, I'm not sure it's checking any dependencies because I use quirky dependency management.  But also, coverage.py has no run-time dependencies, so, whatever?

hugovk:
I don't think I've seen this used anywhere else either.

PR #1740, opened 2024-01-26
CI: Set PYTHON_COLORS=0 to disable coloured CPython tracebacks

hugovk:
Possible alternative to https://github.com/nedbat/coveragepy/commit/75b22f0e24559446ff551acb6ddbaa92bce8add3.

Instead of stripping colour escape sequences from the Python 3.13 tracebacks, set `PYTHON_COLORS=0` to disable only the CPython colour output on the CI.

https://docs.python.org/3.13/using/cmdline.html#envvar-PYTHON_COLORS

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -19,6 +19,7 @@ env:
   PIP_DISABLE_PIP_VERSION_CHECK: 1
   COVERAGE_IGOR_VERBOSE: 1
   FORCE_COLOR: 1    # Get colored pytest output
+  PYTHON_COLORS: 0  # But not colored CPython output
 
 permissions:
   contents: read
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -19,7 +19,6 @@ env:
   PIP_DISABLE_PIP_VERSION_CHECK: 1
   COVERAGE_IGOR_VERBOSE: 1
   FORCE_COLOR: 1    # Get colored pytest output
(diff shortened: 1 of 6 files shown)


nedbat:
I thought about this (after I made the "ignore color" commit).  Is there a reason to prefer one over the other?

hugovk:
Fix it at the source, rather than cleaning up the output. 

One line added in one file, vs. ~30 lines in four files.

I expect CPython will add colour in other places too at some point (I don't know if in 3.13), which could need `without_color` peppering elsewhere too.

nedbat:
An embarrassing admission: until I looked at this PR's diff, I didn't realize I had set FORCE_COLOR=1, I thought GitHub had done it!  You are probably right: less code is better code.

nedbat:
It might be better to set PYTHON_COLORS=0 in a more central place, like tox.ini, so that local test runs will also be immune to local settings.

hugovk:
> An embarrassing admission: until I looked at this PR's diff, I didn't realize I had set FORCE_COLOR=1, I thought GitHub had done it! You are probably right: less code is better code.

Yeah, GitHub Actions is [not a tty](https://github.com/actions/runner/issues/241) so tools' usual colour auto-detection doesn't work there (and we don't want escape codes when piping output), so FORCE_COLOR is needed. (I think colour worked back in the day on Travis CI.)


> It might be better to set PYTHON_COLORS=0 in a more central place, like tox.ini, so that local test runs will also be immune to local settings.

Good idea, updated!

PR #1748, opened 2024-02-10
CI: Only push to report repo and create badge for upstream

hugovk:
Avoid failures like https://github.com/hugovk/coveragepy/actions/runs/7854238034/job/21434885977 when forks update from `upstream/master`.


nedbat:
Thanks!

PR #1754, opened 2024-03-10
Update config.py

tanaydin:
Fixes report commands `format` config type. Which should be a `string`, but defined as `boolean`

https://github.com/nedbat/coveragepy/blob/fd4022ec6db7f4830c3d9b7f1c16b5fab8b14895/coverage/config.py#L220
https://github.com/nedbat/coveragepy/blob/fd4022ec6db7f4830c3d9b7f1c16b5fab8b14895/coverage/report.py#L31

### DIFF
--- a/coverage/config.py
+++ b/coverage/config.py
@@ -395,7 +395,7 @@ class CoverageConfig(TConfigurable, TPluginConfig):
         ("exclude_list", "report:exclude_lines", "regexlist"),
         ("exclude_also", "report:exclude_also", "regexlist"),
         ("fail_under", "report:fail_under", "float"),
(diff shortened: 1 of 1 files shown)


nedbat:
Thanks! I'm curious what drew your attention to this mistake?

nedbat:
A little bit more: 3d57a072ec6073988b4c775a9cffd431ef63ab5f

tanaydin:
I was just trying to configure it via pyproject file to have same output that we got by command line arguments.

nedbat:
This is now released as part of [coverage 7.4.4](https://pypi.org/project/coverage/7.4.4).

Issue #1752, opened 2024-02-22
Relative paths duplicates paths on Windows


henryiii:
**Describe the bug**

I've been working on issues with nox's test suite. We weren't correctly including runs from all three OS's, and we have just been fixing that. We seem to have a 50% chance of failure with `D:\a\nox\nox\__init__.py` showing up with "no source for file". I'm pretty sure this started showing up for the first time without this fix roughly 1-2 days ago, but the issue is more reliable if we properly make sure coverage files don't overwrite each other. We are using editable installs and relative paths, so I'm not sure why it's getting stored that way (even in the succeeding 50%). If I look at the report on Windows, every file is listed twice. If I look at the `.coverage.win32*` files, I see that every file is present twice in `files`, once as a relative path, and once as an absolute path. The other OS's are just fine, it's only Windows that is doing this.

This is with a hatchling editable install of nox, inside a nox runner. Using pytest-cov.

**To Reproduce**
How can we reproduce the problem?  Please *be specific*. Don't link to a failing CI job. Answer the questions below:
1. What version of Python are you using? _3.7-3.12_.
1. What version of coverage.py shows the problem? The output of `coverage debug sys` is helpful.  _7.4.2 and 7.2.7_
1. What versions of what packages do you have installed? The output of `pip freeze` is helpful. _all latest supported_ 
1. What code shows the problem?  Give us a specific commit of a specific repo that we can check out. If you've already worked around the problem, please provide a commit before that fix. https://github.com/wntrblm/nox/commit/3b8e0267b7c2c3631d1098f8a4d6e4f731582423 passed at one point, and now is failing. https://github.com/wntrblm/nox/pull/778/commits/1c998d114dbe31c79b7a78e4010fd2a2a48008f7 is probably better, though, since that's with the OS naming fix.
1. What commands should we run to reproduce the problem? *Be specific*. Include everything, even `git clone`, `pip install`, and so on. Explain like we're five! _I've only done it in CI, since I don't have easy access to a Windows machine right now, but it should be possible to run `pipx run `nox -s 'tests(python="3.12", tox_version="latest")'` (AFAICT it happens on all of them).

**Expected behavior**
There should not be two copies of files on Windows. `SELECT * FROM file;` reports:

```
1|D:\a\nox\nox\nox\__init__.py
2|D:\a\nox\nox\nox\_options.py
3|D:\a\nox\nox\nox\_option_set.py
4|D:\a\nox\nox\nox\tasks.py
[...]

PR #1755, opened 2024-03-14
Fix duplicating paths during remapping

BruRdgz:
Adds tests for #1752's bugs, along with a patch. Closes the issue.

### DIFF
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -489,6 +489,9 @@ class PathAliases:
 
         # If we get here, no pattern matched.
 
+        if self.relative:
+            path = relative_filename(path)
+
         if self.relative and not isabs_anywhere(path):
             # Auto-generate a pattern to implicitly match relative files
             parts = re.split(r"[/\\]", path)
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -500,18 +500,6 @@ class ApiTest(CoverageTest):
             filenames = {relative_filename(f).replace("\\", "/") for f in data.measured_files()}
             return filenames
 
-        # Case 1: get the order right.
-        make_files()
-        self.make_file(".coveragerc", """\
-            [paths]
-            plugins =
-                plugins/
-                ci/girder/plugins/
-            girder =
-                girder/
-                ci/girder/
-            """)
-        assert get_combined_filenames() == {'girder/g1.py', 'plugins/p1.py'}
 
         # Case 2: get the order "wrong".
         make_files()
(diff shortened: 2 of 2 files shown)


tests/test_api.py
```diff
@@ -526,6 +514,51 @@ def get_combined_filenames() -> set[str]:
...
+        self.start_import_stop(cov_bar, "bar")
+        cov_bar.save()
+
+        # Combine the data files in the opposite order.
+        cov2 = coverage.Coverage()
+        cov2.combine(data_paths=[".coverage.bar_cov.data", ".coverage.foo_cov.data"])
+        files2 = cov2.get_data().measured_files()
+
+        # The order of combining should not affect the resulting files.
+        assert set(files1) == set(files2)
```
BruRdgz, line 559: In hindsight, this is pretty obvious; so that's a stupid check to take away.

tests/test_api.py
```diff
@@ -500,18 +500,6 @@ def get_combined_filenames() -> set[str]:
             filenames = {relative_filename(f).replace("\\", "/") for f in data.measured_files()}
             return filenames
 
-        # Case 1: get the order right.
-        make_files()
-        self.make_file(".coveragerc", """\
-            [paths]
-            plugins =
-                plugins/
-                ci/girder/plugins/
```
BruRdgz, line 509: This is weird; I didn't touch this, and these files remain on my local repository. Go figure, I'll check it out later.

BruRdgz (commented):
Have to check those out.

coverage/files.py
```diff
@@ -489,6 +489,9 @@ def map(self, path: str, exists:Callable[[str], bool] = source_exists) -> str:
 
         # If we get here, no pattern matched.
 
+        if self.relative:
```
nedbat, line 492: A key indication of whether the test is doing what it should:  comment out "the fix" and see if the test fails.  When I comment out these two lines, all the tests still pass.

tests/test_api.py
```diff
@@ -500,18 +500,6 @@ def get_combined_filenames() -> set[str]:
             filenames = {relative_filename(f).replace("\\", "/") for f in data.measured_files()}
             return filenames
 
-        # Case 1: get the order right.
```
nedbat, line 503: I don't understand why you removed this test code.

nedbat:
Thanks, I fixed it like this: 1b19799edfbfc65dea254e8fcaa0dea23e9709f9

PR #1771, opened 2024-04-28
Support scrubbing absolute temporary paths in test_html.py on Windows

devdanzin:
Some gold files, like `tests/gold/html/other/index.html` record an absolute path that includes the temporary directory used while testing, like:
```
/private/var/folders/6j/khn0mcrj35d1k3yylpl8zl080000gn/T/pytest-of-ned/pytest-293/t40/othersrc/other.py
```

There's explicit support for the directory format above in the regex:
```regex
/private/var/[\w/]+/pytest-of-\w+/pytest-\d+/(popen-gw\d+/)?t\d+
```

However, when creating the gold files on Windows, the absolute path takes a rather different format, for example:
```
C:\Users\ddini\AppData\Local\Temp\pytest-of-ddini\pytest-9\popen-gw6\t1\othersrc\other.py
```

This PR adds a new regex to match and scrub directories in the format above:
```regex
[A-Z]:\\Users\\[\w\\]+\\pytest-of-\w+\\pytest-\d+\\(popen-gw\d+\\)?t\d+
```

### DIFF
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -712,6 +712,7 @@ def compare_html(
     ]
     if env.WINDOWS:
         # For file paths...
+        scrubs += [(r'[A-Z]:\\Users\\[\w\\]+\\pytest-of-\w+\\pytest-\d+\\(popen-gw\d+\\)?t\d+', 'TEST_TMPDIR')]
         scrubs += [(r"\\", "/")]
     if extra_scrubs:
         scrubs += extra_scrubs
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -712,7 +712,10 @@ def compare_html(
     ]
     if env.WINDOWS:
         # For file paths...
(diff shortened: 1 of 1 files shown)


nedbat:
This looks reasonable on the face of it, but we run the test suite on Windows on GitHub Actions all the time, and this hasn't been needed there.  Do you know why?

devdanzin:
Yes, it's because running the tests on Windows works due to `compare_html` successfully scrubbing the path of the `TEST_TMPDIR` of the generated HTML file. It's handled by the `filepath_to_regex(abs_file(os.getcwd()))` scrub rule.

However, the `TEST_TMPDIR` that gets recorded in a gold file doesn't get handled by the above scrub rule. That's why there is a scrub rule for recorded paths in test_html:
https://github.com/nedbat/coveragepy/blob/fb52e47d41cb47099d1c0020b35d99ace1dd55c8/tests/test_html.py#L711

It handles the current recorded value, but won't work for paths recorded on Windows. So merely running the tests on Windows works, updating the gold files on Windows doesn't.

PR #1772, opened 2024-04-30
Scrub absolute Windows temporary paths on all OSes when running test_html.py

devdanzin:
Sorry, #1771 has a flaw: it only applies the scrubbing when tests run on Windows, so the tests fail on other OSes. 

This PR fixes that.


nedbat:
I was also looking at this now.  The last scrub also needs to run when not on Windows.  You should make this change on your other PR instead of as a new PR, so that we can be sure to get all the cases.

It feels strange to use `scrubs += [(r"\\", "/")]` everywhere, and I was wondering if there was a way to limit it to only filenames, but I don't see a way without expanding how the `scrub()` function works.

nedbat:
See this suggestion: https://github.com/nedbat/coveragepy/pull/1768#issuecomment-2085007139

PR #1773, opened 2024-05-01
Support '# pragma: no branch' in multi-line if statements (fix #754)

### LINKED ISSUE #754: Pragmas on continued long lines don't work
(From https://stackoverflow.com/questions/54111889/coverage-pragma-no-branch-with-multiline-statement)

A pragma comment on the continuation lines of a multi-line if-statement won't have an effect:

For the python coverage package, a missing else can be ignored for the resulting coverage by using # pragma: no branch.

It seems, that this is not working when breaking a long if statement in multiple lines like this:
```
if this_is_a_verylong_boolean_expression == True and another_long_expression \
    and here_another_expression:  # pragma: no branch
    do_something()
```

devdanzin:
`PythonParser.lines_matching()` records a set of lines matching a given regex. However, it doesn't take multi-line statements into account. That causes statements that should be marked as "# pragma: no branch" to be counted as partial branches. IOW, "a pragma comment on the continuation lines of a multi-line if-statement won't have an effect".

This PR fixes that by mapping line numbers through `parser._multiline`, so that the number of the first line in a multi-line statement gets recorded instead.

Fixes #754.

### DIFF
--- a/coverage/parser.py
+++ b/coverage/parser.py
@@ -115,7 +115,7 @@ class PythonParser:
         matches = set()
         for i, ltext in enumerate(self.lines, start=1):
             if regex_c.search(ltext):
-                matches.add(i)
+                matches.add(self._multiline.get(i, i))
         return matches
 
     def _raw_parse(self) -> None:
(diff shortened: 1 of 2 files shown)


nedbat:
Thanks! A simple fix, I should have thought of it.

nedbat:
This is now released as part of [coverage 7.5.1](https://pypi.org/project/coverage/7.5.1).

PR #1776, opened 2024-05-05
Remember the values for 'filter' and 'hide 100% covered' in HTML report (fix #1725)

### LINKED ISSUE #1725: HTML report filter not working properly
**Describe the bug**
The filter on the coverage resets its state when I go to the previous page.

**To Reproduce**
Python 3.8.10  
coverage.py v7.0.1 in my code, and the live example as linked below in the v7.4.0

1. Go to the [example page](https://nedbatchelder.com/files/sample_coverage_html/index.html), as [indicated here](https://coverage.readthedocs.io/en/7.4.0/cmd.html#html-reporting-coverage-html:~:text=Here%E2%80%99s%20a%20sample%20report.): 
2. See the number of lines, 10 lines (counting the "total").
3. Type something in the filter on the top right end of the page, e.g., `utils`. The results will be filtered and there are now only 4 lines (counting the "total").
4. Click in a file, e.g., `cogapp/utils.py`
5. Go back to the previous page (click ALT + left)
6. See that the text is still on the filter, on the top right of the page, but the results is not filtered.

**Expected behavior**
When I go back to the previous page, the results must be filtered.

**Additional context**

When the code below runs, the filter text is still empty.
```js
```
[Location](https://github.com/nedbat/coveragepy/blob/aa55dbaddfce0ec3a053882f690e81b91c010371/coverage/htmlfiles/coverage_html.js#L167C8-L169C73)

### Hotfix
Whithout thinking too much, this is a solution that I came up with: appending a new keyboard event after a timeout period. It's not the ideal, but it's working.

This is my script to see the coverage results: 

```bash
coverage html  && \
    echo 'window.addEventListener("load",(function(){setTimeout((function(){const t=document.getElementById("filter");t&&t.dispatchEvent(new Event("input"))}),5)}));' >> htmlcov/coverage_html.js && \
    xdg-open 'http://localhost:8000' && \
    python -m http.server -d htmlcov
```

devdanzin:
This PR adds support for remembering the "hide covered" checkbox and filter text in HTML report pages.

It adds two new keys to localStorage, but this could be done with a single one if that's preferable. 

Fixes #1725.

### DIFF
--- a/coverage/htmlfiles/coverage_html.js
+++ b/coverage/htmlfiles/coverage_html.js
@@ -125,6 +125,16 @@ coverage.assign_shortkeys = function () {
 
 // Create the events for the filter box.
 coverage.wire_up_filter = function () {
+    // Populate the filter and hide inputs if there are saved values for them
+    const saved_filter_value = localStorage.getItem(coverage.FILTER_STORAGE);
+    if (saved_filter_value) {
+        document.getElementById("filter").value = saved_filter_value;
+    }
+    const saved_hide100_value = localStorage.getItem(coverage.HIDE_STORAGE);
+    if (saved_hide100_value) {
+        document.getElementById("hide100").checked = JSON.parse(saved_hide100_value);
+    }
+
     // Cache elements.
     const table = document.querySelector("table.index");
     const table_body_rows = table.querySelectorAll("tbody tr");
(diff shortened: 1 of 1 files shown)


nedbat:
Thanks! Since you are still working on these things, I think something is happening way too frequently when sorting.  When I get to the function page with about 500 functions, there's a noticeable lag before the sort is ready.

nedbat:
This is now released as part of [coverage 7.5.2](https://pypi.org/project/coverage/7.5.2).

PR #1778, opened 2024-05-05
Try some changes to see whether HTML report performance gets better

devdanzin:
TL;DR: I couldn't improve HTML report performance, only submitting a PR to record investigations somewhere.

HTML reports are lagging noticeably for sorting and page load with long lists of functions.

I've done my best to investigate but couldn't find anything that actually improved the situation. The changes in this PR seemed like good ideas:
 - Batching DOM updates.
 - Simplifying `getCellValue` to the code that actually runs (maybe the deleted code is necessary for other browsers?).
 - Only filter rows on page load if we have a saved filter value.
 - Some micro optimizations of dubious value.

I can't detect any improvements to performance from these.

But I'm also fairly sure now that we aren't doing something too obviously wrong, like quadratic loops or firing expensive events too frequently. It seems to be something related to DOM and JS performance, above my competence level.

The Chrome developer tools sometimes pops up `[Violation] Forced reflow while executing JavaScript took XXms`, but I couldn't get rid of the reflow by reordering attribute accesses.

Maybe we can record sort order and use it instead of sorting every time,  but even that isn't a clear win AFAICS.

### DIFF
--- a/coverage/htmlfiles/coverage_html.js
+++ b/coverage/htmlfiles/coverage_html.js
@@ -35,22 +35,13 @@ function on_click(sel, fn) {
 // Helpers for table sorting
 function getCellValue(row, column = 0) {
     const cell = row.cells[column]  // nosemgrep: eslint.detect-object-injection
-    if (cell.childElementCount == 1) {
-        var child = cell.firstElementChild;
-        if (child.tagName === "A") {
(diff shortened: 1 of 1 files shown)


nedbat:
Thanks for trying a bunch of things.  We can live with it as it is unless someone else finds an improvement.

PR #1780, opened 2024-05-09
Use os.device_encoding(0) to figure out stdin encoding in run_command() (Fix #1777)

### LINKED ISSUE #1777: On Windows, `test_wrong_alias_doesnt_work` fails if the error message contains bytes that UTF-8 cannot decode
**Describe the bug**
Running the tests on Windows, `tests/test_process.py::AliasedCommandTest::test_wrong_alias_doesnt_work` fails if the error message for the wrong alias contains bytes that UTF-8 cannot decode.

**To Reproduce**
How can we reproduce the problem?  Please *be specific*. Don't link to a failing CI job. Answer the questions below:
1. Python version: Python 3.12.3
1. coverage.py version: 7.5.2a0.dev1
1. Packages: pytest 8.1.1
1. Code that shows the problem and commands to run:

The issue happens on this line:
https://github.com/nedbat/coveragepy/blob/695b8bcc6d4a5da6f3bd320c3b7787cda0fc1467/tests/helpers.py#L65

From this code:
```python
def run_command(cmd: str) -> tuple[int, str]:
    """Run a command in a sub-process.

    Returns the exit status code and the combined stdout and stderr.

    """
    # Subprocesses are expensive, but convenient, and so may be over-used in
    # the test suite.  Use these lines to get a list of the tests using them:
    if 0:  # pragma: debugging
        with open("/tmp/processes.txt", "a") as proctxt:  # type: ignore[unreachable]
            print(os.getenv("PYTEST_CURRENT_TEST", "unknown"), file=proctxt, flush=True)

    # In some strange cases (PyPy3 in a virtualenv!?) the stdout encoding of
    # the subprocess is set incorrectly to ascii.  Use an environment variable
    # to force the encoding to be the same as ours.
    sub_env = dict(os.environ)
    sub_env['PYTHONIOENCODING'] = output_encoding()

    proc = subprocess.Popen(
        cmd,
        shell=True,
        env=sub_env,
        stdin=subprocess.PIPE,
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
    )
    output, _ = proc.communicate()
    status = proc.returncode

    # Get the output, and canonicalize it to strings with newlines.
    output_str = output.decode(output_encoding()).replace("\r", "")  # <= Here
    return status, output_str
```
The issue is that on Windows `output` can have cp1252 as encoding and `output_encoding()` will return UTF-8, leading to e.g.
```
 UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc6 in position 13: invalid continuation byte
```

It can be reproduced by running the following code in Windows Terminal on Windows 11:
```python
from tests.helpers import run_command
print(run_command('echo 1'))  # Prints "(0, '1\n')"
print(run_command('echo é'))  # Error
```

**Expected behavior**
`run_command` should be able to handle output from `subprocess.Popen` correctly.

**Additional context**
I see some ways to fix this, not sure what is best. I'd prefer calling `locale.getpreferredencoding()` in `run_commands`, as it's about the encoding used to generate the error message. Another way would be to change the order in which `output_encoding` tries to get the right encoding. And the way I like the least, try decoding as it's currently done and use `locale.getpreferredencoding()` if that fails.

I'll send a PR once we decide what's the best solution.

devdanzin:
The output of `subprocess.Popen` on Windows might include bytes that cannot be decoded by UTF-8. 

This PR uses `os.device_encoding(0)` to figure out the real encoding of stdin (fd 0) so `run_command` can use that to decode the output. Since `os.device_encoding` may return None, we fallback to `locale.getpreferredencoding`.

Fixes #1777.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -9,6 +9,7 @@ import collections
 import contextlib
 import dis
 import io
+import locale
 import os
 import os.path
 import re
@@ -62,7 +63,8 @@ def run_command(cmd: str) -> tuple[int, str]:
     status = proc.returncode
 
     # Get the output, and canonicalize it to strings with newlines.
-    output_str = output.decode(output_encoding()).replace("\r", "")
+    encoding = os.device_encoding(0) or locale.getpreferredencoding()
+    output_str = output.decode(encoding).replace("\r", "")
     return status, output_str
 
 
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -29,7 +29,6 @@ import pytest
 from coverage import env
 from coverage.debug import DebugControl
 from coverage.exceptions import CoverageWarning
-from coverage.misc import output_encoding
 from coverage.types import TArc, TLineNo
 
 
@@ -45,11 +44,13 @@ def run_command(cmd: str) -> tuple[int, str]:
         with open("/tmp/processes.txt", "a") as proctxt:  # type: ignore[unreachable]
             print(os.getenv("PYTEST_CURRENT_TEST", "unknown"), file=proctxt, flush=True)
 
+    encoding = os.device_encoding(1) or locale.getpreferredencoding()
+
     # In some strange cases (PyPy3 in a virtualenv!?) the stdout encoding of
     # the subprocess is set incorrectly to ascii.  Use an environment variable
     # to force the encoding to be the same as ours.
     sub_env = dict(os.environ)
-    sub_env['PYTHONIOENCODING'] = output_encoding()
+    sub_env['PYTHONIOENCODING'] = encoding
 
     proc = subprocess.Popen(
         cmd,
@@ -63,7 +64,6 @@ def run_command(cmd: str) -> tuple[int, str]:
     status = proc.returncode
 
     # Get the output, and canonicalize it to strings with newlines.
-    encoding = os.device_encoding(1) or locale.getpreferredencoding()
     output_str = output.decode(encoding).replace("\r", "")
     return status, output_str
 
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -63,7 +63,7 @@ def run_command(cmd: str) -> tuple[int, str]:
     status = proc.returncode
 
     # Get the output, and canonicalize it to strings with newlines.
-    encoding = os.device_encoding(0) or locale.getpreferredencoding()
+    encoding = os.device_encoding(1) or locale.getpreferredencoding()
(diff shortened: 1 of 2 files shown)


nedbat:
It wasn't until you mentioned stdin here that I remembered that 0 is stdin.  Wouldn't it be better to check stdout (1)?  I doubt they'd be different, but we will be processing the output of a command.

tests/helpers.py
```diff
@@ -62,7 +63,8 @@ def run_command(cmd: str) -> tuple[int, str]:
     status = proc.returncode
 
     # Get the output, and canonicalize it to strings with newlines.
-    output_str = output.decode(output_encoding()).replace("\r", "")
+    encoding = os.device_encoding(0) or locale.getpreferredencoding()
```
nedbat:
```suggestion
    encoding = os.device_encoding(1) or locale.getpreferredencoding()
```

devdanzin:
> Wouldn't it be better to check stdout (1)? I doubt they'd be different, but we will be processing the output of a command.

Indeed, it makes sense that it would be better to check fd 1. I tried a little experiment and changing the encoding of stdout in a terminal makes a difference, but that of stdin doesn't. Here's some notes for (my own) future reference.

Here's the test script I used:

```python
import os, subprocess

proc = subprocess.Popen('echo é𐍈', shell=True, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
out, _ = proc.communicate()
stdin_enc, stdout_enc = os.device_encoding(0), os.device_encoding(1)

print(out)
print(stdin_enc, stdout_enc)
# print(out.decode("utf-8"))
print(out.decode(stdin_enc))
print(out.decode(stdout_enc))
```

Running it in a default terminal will yield:
```
b'\x82??\r\n'
cp850 cp850
é??

é??

```

To only change the stdout encoding in PowerShell, run:
```powershell
[Console]::OutputEncoding = New-Object System.Text.UTF8Encoding
```

Running it in a terminal with utf-8 stdout results in:

```
b'\xc3\xa9\xf0\x90\x8d\x88\r\n'
cp850 cp65001
├®­Éìê

é𐍈

```

I think this means the important encoding is that of stdout.

nedbat:
Sorry to drag this out, but `output_encoding()` is only used in this function, so let's move the improved encoding choice code into output_encoding.  That way we're decoding with the same encoding we've set in the environment.

nedbat:
I've pushed a commit to remove the helper and use the same encoding in both places.

PR #1782, opened 2024-05-13
Report data file errors in more detail

### LINKED ISSUE #1781: Add a hint that `coverage combine` might be needed
From the Python IRC channel on Libera:
```
Person: coverage.py's `coverage report` gives me "No data to report.", and it looks like that is because the output file isn't .coverage, but dot-suffixed with my machine name etc. Now, this appears to be done *not* because of `parallel = true` in my config, but because `concurrency = multiprocessing`. However, `coverage report` does not pick up on this, or even when I do explicitly add `parallel = true`
Person: If I specify the output file explicitly when calling `coverage report`, I get the desired output (but that's not a feasible solution as these filenames are not consistent)
Person: fml.... I type out my problem and the solution appears: yes, I have to call `coverage combine` explicitly
```

The "No data to report" message could check if there are dot-suffixed data files, and add, "Perhaps you need to run `coverage combine`" or something similar.  This would guide people to a solution.

bartfeenstra:
This PR fixes https://github.com/nedbat/coveragepy/issues/1781

## Changes
- Introduce `NoDataError` subclasses for specific purposes
  - Factory methods to construct helpful messages, because it keeps all the default behavior around assumptions of exceptions just having a single string argument working (it gets funky when pickling and what not, so I tend to take this approach)
- When possible, suggest that the user runs the combine command
- All critical (albeit sometimes seemingly trivial) bits are now centralized, e.g. no more duplication of messages. This makes it easier to keep everything in sync and improve on.

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -13,6 +13,7 @@ import shlex
 import sys
 import textwrap
 import traceback
+from contextlib import suppress
 
 from typing import cast, Any, NoReturn
 
@@ -24,7 +25,8 @@ from coverage.config import CoverageConfig
 from coverage.control import DEFAULT_DATAFILE
 from coverage.data import combinable_files, debug_data_file
 from coverage.debug import info_header, short_stack, write_formatted_info
-from coverage.exceptions import _BaseCoverageException, _ExceptionDuringRun, NoSource
+from coverage.exceptions import _BaseCoverageException, _ExceptionDuringRun, NoSource, \
+    NoDataFilesFoundError
 from coverage.execfile import PyRunner
 from coverage.results import display_covered, should_fail_under
 from coverage.version import __url__
@@ -882,9 +884,10 @@ class CoverageScript:
             print(info_header("data"))
             data_file = self.coverage.config.data_file
             debug_data_file(data_file)
-            for filename in combinable_files(data_file):
-                print("-----")
-                debug_data_file(filename)
+            with suppress(NoDataFilesFoundError):
+                for filename in combinable_files(data_file):
+                    print("-----")
+                    debug_data_file(filename)
         elif args[0] == "config":
             write_formatted_info(print, "config", self.coverage.config.debug_info())
         elif args[0] == "premain":
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -25,8 +25,9 @@ from coverage.config import CoverageConfig
 from coverage.control import DEFAULT_DATAFILE
 from coverage.data import combinable_files, debug_data_file
 from coverage.debug import info_header, short_stack, write_formatted_info
-from coverage.exceptions import _BaseCoverageException, _ExceptionDuringRun, NoSource, \
-    NoDataFilesFoundError
+from coverage.exceptions import (
+    _BaseCoverageException, _ExceptionDuringRun, NoSource, NoDataFilesFoundError,
+)
 from coverage.execfile import PyRunner
 from coverage.results import display_covered, should_fail_under
 from coverage.version import __url__
--- a/coverage/data.py
+++ b/coverage/data.py
@@ -82,7 +82,7 @@ def combinable_files(data_file: str, data_paths: Iterable[str] | None = None) ->
             pattern = glob.escape(os.path.join(os.path.abspath(p), local)) +".*"
             files_to_combine.extend(glob.glob(pattern))
         else:
-            raise DataFileOrDirectoryNotFoundError.new_for_data_file_or_directory(
+            raise DataFileOrDirectoryNotFoundError.new(
                 p, is_combining=True
             )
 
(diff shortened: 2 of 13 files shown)


nedbat:
Thanks for doing this. I would like to include the "coverage combine" message only if there are combinable data files.  I've included other comments in the diff.

coverage/cmdline.py
```diff
@@ -24,7 +25,8 @@
 from coverage.control import DEFAULT_DATAFILE
 from coverage.data import combinable_files, debug_data_file
 from coverage.debug import info_header, short_stack, write_formatted_info
-from coverage.exceptions import _BaseCoverageException, _ExceptionDuringRun, NoSource
+from coverage.exceptions import _BaseCoverageException, _ExceptionDuringRun, NoSource, \
+    NoDataFilesFoundError
```
nedbat: Please use parentheses instead of backslashes:
```suggestion
from coverage.exceptions import (
    _BaseCoverageException, _ExceptionDuringRun, NoSource, NoDataFilesFoundError,
)
```

coverage/cmdline.py
```diff
@@ -882,9 +884,10 @@ def do_debug(self, args: list[str]) -> int:
             print(info_header("data"))
             data_file = self.coverage.config.data_file
             debug_data_file(data_file)
-            for filename in combinable_files(data_file):
-                print("-----")
-                debug_data_file(filename)
+            with suppress(NoDataFilesFoundError):
```
nedbat, line 888: Why do we need to suppress this error here?

coverage/data.py
```diff
@@ -82,12 +82,17 @@ def combinable_files(data_file: str, data_paths: Iterable[str] | None = None) ->
             pattern = glob.escape(os.path.join(os.path.abspath(p), local)) +".*"
             files_to_combine.extend(glob.glob(pattern))
         else:
-            raise NoDataError(f"Couldn't combine from non-existent path '{p}'")
+            raise DataFileOrDirectoryNotFoundError.new_for_data_file_or_directory(
```
nedbat: I think I understand why you are using classmethods instead of the constructor, but let's just call it `new`.  The method name pretty much duplicates the class name, and there is only one classmethod.

coverage/exceptions.py
```diff
@@ -24,11 +34,61 @@ class DataError(CoverageException):
...
     """We didn't have data to work with."""
     pass
 
 
+class DataFileOrDirectoryNotFoundError(NoDataError):
+    """A data file or data directory could be found."""
+    @classmethod
+    def new_for_data_file_or_directory(
+        cls, data_file_or_directory_path: str, *, is_combining: bool = False
+    ) -> 'DataFileOrDirectoryNotFoundError':
```
nedbat: No need to quote the return type, we have the `__future__` import at the top of the file.

coverage/exceptions.py
```diff
@@ -24,11 +34,61 @@ class DataError(CoverageException):
...
+        """
+        Create a new instance.
+        """
+        message = (
+            f"The data file or directory `{os.path.abspath(data_file_or_directory_path)}` could not"
```
nedbat: You've used backticks here and elsewhere, as if these are Markdown, but they are not.  Better to use simple single quotes.

coverage/exceptions.py
```diff
@@ -24,11 +34,61 @@ class DataError(CoverageException):
...
+            " be found."
```
nedbat: Please include an explicit `+` instead of relying on string concatenation.

coverage/html.py
```diff
@@ -317,7 +317,9 @@ def report(self, morfs: Iterable[TMorf] | None) -> float:
                 file_be_gone(os.path.join(self.directory, ftr.html_filename))
 
         if not have_data:
-            raise NoDataError("No data to report.")
+            raise DataFileOrDirectoryNotFoundError.new_for_data_file_or_directory(
+                os.path.dirname(self.coverage.get_data().base_filename())
```
nedbat, line 321: I have to think through whether the message here is correct.  Is it always that the file doesn't exist?

nedbat: It's good that the exceptions centralize the messages, but everywhere this exception is created it uses the same fairly complex expression.  Can we centralize that as well?

tests/test_api.py
```diff
@@ -300,7 +300,13 @@ def test_empty_reporting(self) -> None:
         # empty summary reports raise exception, just like the xml report
         cov = coverage.Coverage()
         cov.erase()
-        with pytest.raises(NoDataError, match="No data to report."):
+        with pytest.raises(
+            NoDataError,
+            match=(
+                r"^The data file or directory `(.+?)` could not be found\. Perhaps `coverage "
```
nedbat: "The data file or directory" seems odd to me.  Don't we know whether we are expecting a file or directory?

nedbat (changes requested):

tests/test_api.py
bartfeenstra: These results depend on the paths entered in the CLI. As far as I understand the code these are allowed to be data file paths or paths to directories containing data files. If the path does not exist, I'm not sure we would know if the user meant a file or a directory.

coverage/cmdline.py
bartfeenstra: We do not, but I tried to limit how many outputs I changed. Would you like this to display a friendly error instead?

nedbat:
Thanks for the work, but coming back to this, I'm not a fan of this style of exception creation.

bartfeenstra:
No worries! To learn from this: what would your approach be?

PR #1798, opened 2024-06-19
Contributing partial work towards pytest branch coverage

jaltmayerpizzorno:
Hi Ned, I started working on this test to ensure that Coverage.py would provide coverage within pytest modules, but ran out of time before my trip tomorrow. I am also including a `wrap_pytest` function that I think should be sufficient to restore AST-based instrumentation on pytest modules despite assertion rewriting... but I couldn't get the tests to work, so never integrated it.

### DIFF
--- a/coverage/sysmon.py
+++ b/coverage/sysmon.py
@@ -383,6 +383,10 @@ class SysMonitor(TracerCore):
             last_line = self.last_lines.get(frame)
             if last_line is not None:
                 arc = (last_line, -code.co_firstlineno)
+                import contextlib
+                with open("/tmp/foo.out", "a") as f:
+                    with contextlib.redirect_stdout(f):
+                        print(f"sysmon adding {arc = } (return)")
                 # log(f"adding {arc=}")
                 cast(Set[TArc], code_info.file_data).add(arc)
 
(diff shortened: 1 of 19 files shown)


coverage/sysmon.py
```diff
@@ -382,6 +383,10 @@ def sysmon_py_return_arcs(
             last_line = self.last_lines.get(frame)
             if last_line is not None:
                 arc = (last_line, -code.co_firstlineno)
+                import contextlib
+                with open("/tmp/foo.out", "a") as f:
+                    with contextlib.redirect_stdout(f):
+                        print(f"sysmon adding {arc = } (return)")
```
ferdnyc, line 389: I'm trying to understand the purpose of this code.

I get why `contextlib.redirect_stdout` is useful for capturing the output of _other_ code that directly `print()`s data we'd prefer to send somewhere else. Its docs give the example of capturing a `help()` call in the REPL. Smart use of redirection, very clever.

But what's the point of using it just to surround an explicit `print()`? `print` _**already has**_ support for redirecting its output somewhere other than `sys.stdout`.

What's the advantage of doing this, over...
```python
with open("/tmp/foo.out", "a") as f:
    print(f"sysmon adding {arc=} (return)", file=f)
```
or even...
```python
with open("/tmp/foo.out", "a") as f:
    f.write(f"sysmon adding {arc=} (return)\n")
```
?

nedbat: This code is temporary debugging code. It's written to be quick to insert, and then will be removed later.  I have a snippet in my editor to add this:
```py
import contextlib
with open("/tmp/foo.out", "a") as f:
    with contextlib.redirect_stdout(f):
        print(f"") 
```
This makes it easy for me to add print statements without having to remember `file=f`, or that `f.write` only accepts strings.

Issue #1803, opened 2024-06-24
Provide the ability to exclude a code block from coverage


xqt:
**Excluding a code block**
Having a greater code block it would be great to have aq change to disable a code block from coverage with a ``start`` end ``end`` token. Currently only a single line can be excluded or code blocks within a clause. For example either the whole debug clause can be  excluded but not the last two statements

This excludes the the whole `if debug` clause
```
a = my_function1()
if debug:   # pragma: no cover
    msg = "blah blah"
    log_message(msg, a)
    b = my_function2()
```
but I want to exclude the latest two statements (in practice these can be a lot of statemens). This can be done by excluding every line like
```
a = my_function1()
if debug: 
    msg = "blah blah"
    log_message(msg, a)   # pragma: no cover
    b = my_function2()   # pragma: no cover
```
but this may lead to a lot of such ``pragma: no cover`` command or a senseless clause must be used like
```
a = my_function1()
if debug: 
    msg = "blah blah"
    if True:   # pragma: no cover
        log_message(msg, a)
        b = my_function2()
```

**start/end command as proposal**
A start/end command can be introduced to solve this issue like `pragma: no cover start` and  `pragma: no cover stop` or shorter  `no cover: start` and  `no cover: stop` or something like that. Instead of `no cover: stop` command a `no cover: end` command can be implemented as stop command. The code can then be excluded with

```
a = my_function1()
if debug: 
    msg = "blah blah"
    # no cover: start
    log_message(msg, a)
    b = my_function2()
    # no cover: stop
```

**Other implementations as a sample**
Coverage for Dart language support such block exclusion: (https://github.com/dart-lang/coverage)

nedbat:
Help me understand your situation.  In this code:
```
a = my_function1()
if debug: 
    msg = "blah blah"
    # no cover: start
    log_message(msg, a)
    b = my_function2()
    # no cover: stop
```
How would the `msg = ` line be executed, but the `log_message()` line would not?  Why do we need to exclude a block of lines adjacent to other executed lines?

devdanzin:
Maybe in cases like:
 ```python
a = my_function1()
if debug: 
    msg = "blah blah"
    if a is None:  # common path
      return my_function3()
    # no cover: start
    log_message(msg, a)
    b = my_function2()
    return b # no cover: stop
```

Or even:
```python
def main():
  if not DEBUG:
    sys.exit()

if __name__ == "__main__":
  main()
  from debug import show_info  # no cover: start
[...]

Issue #1741, opened 2024-01-30
Question: Can `no-cover` "cover" the rest of the function?


stdedos:
**Is your feature request related to a problem? Please describe.**

I have the end of the `main()` function like so:

```py
    ...

    if args.command == Commands.CMD.value:
        return cmd_handler(j, args)
    
    print(f"Either command '{args.command}' was not handled, or the end of the world is coming!\n", file=sys.stderr)
    parser.print_help(file=sys.stderr)

    return os.EX_USAGE
```

I'd like to ignore it with a `no-cover`.

Basically, the rest of the function would be hard to hit (somehow argparse does not "match" command with its function, or dev forgets to `return cmd_handler()`.

I'd like to avoid the `if return else` pattern since, on top of that, I agree with the https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html check aesthetically

Any solutions I missed?

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
Add any other context about the feature request here.

nedbat:
Currently there is not a pragma comment that means "ignore the rest of this." The simplest approach that you haven't mentioned is to put that tail of the function into a new function, then you can pragma the call and the function definition with one comment each.  It will still look a bit odd though.

stdedos:
Oh yeah - something like `def handle_unexpected_main_fallthrough(cmd): # pragma: no-cover`

... ofc that will still need the `return handle_unexpected_main_fallthrough(cmd)`, I guess.
But better than nothing, or throwing some "known ignored exception" I guess.

I can go with that. Thank you for the idea 🙃

xqt:
See proposal #1803

nedbat:
As of commit 1fe897d740ebc67bd1f6493bcc889846f16abf10, you can do this with a multi-line exclusion regex if you are careful.  You need to come up with a regex that matches the region you want to exclude, perhaps this:
```ini
[report]
exclude_also =
    # no cover: to return(?s:.)*?return
```
which will work exclude everything from `# no cover: to return` to the next `return` in the function.

nedbat:
This is now released as part of [coverage 7.6.0](https://pypi.org/project/coverage/7.6.0).

Issue #118, opened 2011-02-05
Allow excluding entire modules/packages


nedbat:
*Originally reported by* **Ram Rachum (Bitbucket: [coolrr](https://bitbucket.org/coolrr), GitHub: [coolrr](https://github.com/coolrr))**

----------------------------------------

Allow excluding entire modules/packages from coverage measurement by including some identifiable string in their source, similarly to "pragma: no cover".

Here's a suggestion on how to do this, feel free to suggest alternatives: A line "pragma: no module cover" in a module will cause it not to be measured. A "pragma: no package cover" in an `__init__.py` file will cause the entire package, (including any submodules, subpackages, etc.) not to be measured.

----------------------------------------
- Bitbucket: https://bitbucket.org/ned/coveragepy/issue/118

nedbat:
Issue #300 was marked as a duplicate of this issue.

nedbat:
*Original comment by* **René Fleschenberg (Bitbucket: [rfleschenberg](https://bitbucket.org/rfleschenberg), GitHub: [rfleschenberg](https://github.com/rfleschenberg))**

----------------------------------------

Is there a reason why we cannot just use `pragma: no cover` at module level to exclude the whole module from coverage?

nedbat:
I haven't looked into what it would take to implement this.  Ram's original proposal included a new pragma for `__init__.py` to exclude the whole package.  That would need to be different than the pragma for a single module, I think.

nedbat:
As of commit 1fe897d740ebc67bd1f6493bcc889846f16abf10, you can do this with a multi-line exclusion regex:
```ini
[report]
exclude_also =
    (?s)\A.*# pragma: exclude file.*\Z
```
and then put the comment `# pragma: exclude file` in the file somewhere.

nedbat:
This is now released as part of [coverage 7.6.0](https://pypi.org/project/coverage/7.6.0).

PR #1807, opened 2024-07-02
Support excluding multiline regexes (fix #996)

### LINKED ISSUE #996: Excluding multiple lines with exclude_lines
As far as I can tell, it isn't posible to exclude multiple lines with exclude_lines. If I am reading the code correctly, it uses the `lines_matching` function, which checks the regexes line-by-line

https://github.com/nedbat/coveragepy/blob/eb5d81a302693abc08508e6431392056cc5043c9/coverage/parser.py#L99-L115

My use-case is I want to exclude lines like

```py
except ValueError: 
    assume(False)
```

where `assume` is from hypothesis. Simply excluding `assume(False)` doesn't work because it doesn't exclude the `except`. I don't want to exclude other `except ValueError` blocks in the codebase. I know I can rewrite this as `except ValueError: assume(False)` on a single line, but I would prefer not to as I don't like code that does that.

What I would like to do is to write something like

```ini
[report]
exclude_lines = 
    except ValueError:\n *assume\(False\)
```

Or if that can't work

```ini
[report]
exclude_lines = 
    except ValueError:(?=\n *assume\(False\))
```

Neither of these work presently. As far as I can tell this can only be excluded by manually putting `# pragma: no cover` on each instance.

devdanzin:
This PR allows using multiline regexes in `exclude_lines`. It does so by matching regexes against the whole source code, instead of line by line as was done before.

The behavior is equivalent to adding a `# pragma: no cover` comment to each line matched by the multiline regex. Because of that, if a suite (e.g. `if True:`) is matched, all statements inside it will also be excluded, just like if there was a no-cover pragma on the suite line.

This is made possible because @nedbat figured out a O(N) algorithm for finding line numbers for the matches. With it, there's a small performance increase for excluding lines compared to the previous code.

Fixes #996.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -922,6 +922,42 @@ class ExclusionParserTest(PythonParserTestBase):
         assert parser.raw_statements == {1, 2, 3, 5, 6, 8, 9}
         assert parser.statements == {1, 2, 3, 9}
 
+    def test_multiline_exclusion_whole_source(self) -> None:
+        # https://github.com/nedbat/coveragepy/issues/118
+        regex = r"(?s:.)*# pragma: exclude file(?s:.)*\Z"
+        parser = self.parse_text("""\
+            import coverage
+            # pragma: exclude file
+            def the_void() -> None:
+                if "py" not in __file__:
+                    print("Not a Python file.")
+                print("Everything here is excluded.")
+
+                return
+            print("Excluded too")
+            """, regex)
+        assert parser.lines_matching(regex) == {1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
+        assert parser.raw_statements == {1, 3, 4, 5, 6, 8, 9}
+        assert parser.statements == set()
+
+    def test_multiline_exclusion_from_marker(self) -> None:
+        # https://github.com/nedbat/coveragepy/issues/118
+        regex = r"# pragma: rest of file(?s:.)*\Z"
+        parser = self.parse_text("""\
+            import coverage
+            # pragma: rest of file
+            def the_void() -> None:
+                if "py" not in __file__:
+                    print("Not a Python file.")
+                print("Everything here is excluded.")
+
+                return
+            print("Excluded too")
+            """, regex)
+        assert parser.lines_matching(regex) == {2, 3, 4, 5, 6, 7, 8, 9, 10}
+        assert parser.raw_statements == {1, 3, 4, 5, 6, 8, 9}
+        assert parser.statements == {1}
+
 
 class ParserMissingArcDescriptionTest(PythonParserTestBase):
     """Tests for PythonParser.missing_arc_description."""
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -823,16 +823,18 @@ class ExclusionParserTest(PythonParserTestBase):
         assert parser.statements == set()
 
     def test_multiline_exclusion_multiple_matches(self) -> None:
-        regex = r"print\('.*'\)"
+        regex = r"print\('.*'\)\n\s+. = \d"
         parser = self.parse_text("""\
             def foo():
                 print('Hello, world!')
+                a = 5
             def bar():
                 print('Hello again!')
+                b = 6
             """, regex)
-        assert parser.lines_matching(regex) == {2, 4}
-        assert parser.raw_statements == {1, 2, 3, 4}
-        assert parser.statements == {1, 3}
+        assert parser.lines_matching(regex) == {2, 3, 5, 6}
+        assert parser.raw_statements == {1, 2, 3, 4, 5, 6}
+        assert parser.statements == {1, 4}
 
     def test_multiline_exclusion_suite2(self) -> None:
         regex = r"print\('Hello, world!'\)\n\s+if T:"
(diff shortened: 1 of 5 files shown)


tests/test_parser.py
```diff
@@ -781,6 +781,75 @@ def function() -> int:
         assert parser.raw_statements == {1, 3, 4, 5, 6, 8, 9}
         assert parser.statements == {1, 8, 9}
 
+    def test_multiline_exclusion(self) -> None:
```
nedbat: This test could be broken into a number of separate tests since it's so long.

devdanzin: Done, let me know if it should be done in a better way.

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
         assert parser.raw_statements == {1, 3, 4, 5, 6, 8, 9}
         assert parser.statements == {1, 8, 9}
 
+    def test_multiline_exclusion_single_line(self) -> None:
+        regex = r"print\('.*'\)"
+        parser = self.parse_text("""\
+        def foo():
+            print('Hello, world!')
+        """, regex)
```
nedbat: Please indent the code strings like this:
```suggestion
        parser = self.parse_text("""\
            def foo():
                print('Hello, world!')
            """, regex)
```
This should match the way other code is included in tests.

devdanzin: Thank you! Done.

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
         assert parser.raw_statements == {1, 3, 4, 5, 6, 8, 9}
         assert parser.statements == {1, 8, 9}
 
+    def test_multiline_exclusion_single_line(self) -> None:
+        regex = r"print\('.*'\)"
+        parser = self.parse_text("""\
+            def foo():
+                print('Hello, world!')
+        """, regex)
```
nedbat:
```suggestion
            """, regex)
```

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
...
+        assert parser.statements == {1}
+
+    def test_multiline_exclusion_suite(self) -> None:
+        regex = r"if T:\n\s+print\('Hello, world!'\)"
+        parser = self.parse_text("""\
+            def foo():
+                if T:
+                    print('Hello, world!')
+                    print('This is a multiline regex test.')
+        """, regex)
```
nedbat:
```suggestion
            """, regex)
```

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
...
+        assert parser.lines_matching(regex) == {2, 3}
+        assert parser.raw_statements == {1, 2, 3, 4}
+        assert parser.statements == {1}
+
+    def test_multiline_exclusion_no_match(self) -> None:
+        regex = r"nonexistent"
+        parser = self.parse_text("""\
+            def foo():
+                print('Hello, world!')
+        """, regex)
```
nedbat:
```suggestion
            """, regex)
```

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
...
+        assert parser.statements == set()
+
+    def test_multiline_exclusion_multiple_matches(self) -> None:
+        regex = r"print\('.*'\)"
+        parser = self.parse_text("""\
+            def foo():
+                print('Hello, world!')
+            def bar():
+                print('Hello again!')
+        """, regex)
```
nedbat:
```suggestion
            """, regex)
```

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
...
+        assert parser.statements == {1, 3}
+
+    def test_multiline_exclusion_suite2(self) -> None:
+        regex = r"print\('Hello, world!'\)\n\s+if T:"
+        parser = self.parse_text("""\
+            def foo():
+                print('Hello, world!')
+                if T:
+                    print('This is a test.')
+        """, regex)
```
nedbat:
```suggestion
            """, regex)
```

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
...
+
+    def test_multiline_exclusion_match_all(self) -> None:
+        regex = (r"def foo\(\):\n\s+print\('Hello, world!'\)\n"
+                 r"\s+if T:\n\s+print\('This is a test\.'\)")
+        parser = self.parse_text("""\
+            def foo():
+                print('Hello, world!')
+                if T:
+                    print('This is a test.')
+        """, regex)
```
nedbat:
```suggestion
            """, regex)
```

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
...
+        regex = r"print\('.*'\)"
+        parser = self.parse_text("""\
+            def foo():
+                print('Hello, world!')
+        """, regex)
+        assert parser.lines_matching(regex) == {2}
+        assert parser.raw_statements == {1, 2}
+        assert parser.statements == {1}
+
+    def test_multiline_exclusion_suite(self) -> None:
```
nedbat, line 794:
```suggestion
    def test_multiline_exclusion_suite(self) -> None:
        # A multi-line exclusion that matches a colon line still excludes the entire block.
```

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
...
+        regex = r"if T:\n\s+print\('Hello, world!'\)"
+        parser = self.parse_text("""\
+            def foo():
+                if T:
+                    print('Hello, world!')
+                    print('This is a multiline regex test.')
```
nedbat, line 801:
```suggestion
                    print('This is a multiline regex test.')
            a = 5
```

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
...
+        """, regex)
+        assert parser.lines_matching(regex) == {2, 3}
+        assert parser.raw_statements == {1, 2, 3, 4}
+        assert parser.statements == {1}
```
nedbat:
```suggestion
        assert parser.raw_statements == {1, 2, 3, 4, 5}
        assert parser.statements == {1, 5}
```

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
...
+
+    def test_multiline_exclusion_no_source(self) -> None:
+        regex = r"anything"
+        parser = PythonParser(text="", filename="dummy.py", exclude=regex)
+        assert parser.lines_matching(regex) == set()
+        assert parser.raw_statements == set()
+        assert parser.statements == set()
+
+    def test_multiline_exclusion_multiple_matches(self) -> None:
+        regex = r"print\('.*'\)"
```
nedbat: This isn't a multi-line regex?

tests/test_parser.py
```diff
@@ -781,6 +781,82 @@ def function() -> int:
...
+                if T:
+                    print('This is a test.')
+        """, regex)
+        assert parser.lines_matching(regex) == {2, 3}
+        assert parser.raw_statements == {1, 2, 3, 4}
+        assert parser.statements == {1}
+
+    def test_multiline_exclusion_match_all(self) -> None:
+        regex = (r"def foo\(\):\n\s+print\('Hello, world!'\)\n"
+                 r"\s+if T:\n\s+print\('This is a test\.'\)")
```
nedbat:
```suggestion
        regex = (
            r"def foo\(\):\n\s+print\('Hello, world!'\)\n"
            + r"\s+if T:\n\s+print\('This is a test\.'\)"
        )
```

nedbat:
We should also add a test for the #1803 example.

nedbat:
Do you want to update the docs, or leave that to me?

tests/test_parser.py
devdanzin: Right, that was just a sanity check from when I was sketching this that slipped through. I'll rewrite the test so it covers multiple multi-line matches.

nedbat:
When this pull request is merged, we should update issues #118, #996, #1741, and #1803.

devdanzin:
> Do you want to update the docs, or leave that to me?

No, I'd rather let you do your thing with the docs. But thanks for asking :)

nedbat:
This is now released as part of [coverage 7.6.0](https://pypi.org/project/coverage/7.6.0).

PR #1809, opened 2024-07-08
Add region data to JSON report (fix #1793)

### LINKED ISSUE #1793: A way to get report data in detail
The report method only returns the total coverage amount. However, sometimes we need this information in detail. For example, I need to know the function-by-function coverage of a file, and I can't do this using the report method. I know it's possible to do this using the html report method, but having the value inside a variable would make it much easier for a more detailed analysis of where the coverage is bad.

So if the report method returned a dictionary, for example, with the names of the functions and their coverage, it would be very useful.

devdanzin:
This PR addresses #1793 by adding data regarding class and function/method coverage to the JSON report.

### DIFF
--- a/coverage/jsonreport.py
+++ b/coverage/jsonreport.py
@@ -19,6 +19,7 @@ from coverage.types import TMorf, TLineNo
 if TYPE_CHECKING:
     from coverage import Coverage
     from coverage.data import CoverageData
+    from coverage.plugin import FileReporter
 
 
 # "Version 1" had no format number at all.
@@ -60,6 +61,7 @@ class JsonReporter:
             measured_files[file_reporter.relative_filename()] = self.report_one_file(
                 coverage_data,
                 analysis,
+                file_reporter,
             )
 
         self.report_data["files"] = measured_files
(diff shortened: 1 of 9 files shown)


coverage/jsonreport.py
```diff
@@ -122,6 +126,45 @@ def report_one_file(self, coverage_data: CoverageData, analysis: Analysis) -> di
...
+            }
+            if self.config.json_show_contexts:
+                contexts = coverage_data.contexts_by_lineno(narrowed_analysis.filename)
+                reported_file[region.kind][region.name]["contexts"] = contexts
+            if coverage_data.has_arcs():
+                narrowed_summary.update({
+                    "num_branches": narrowed_nums.n_branches,
+                    "num_partial_branches": narrowed_nums.n_partial_branches,
+                    "covered_branches": narrowed_nums.n_executed_branches,
+                    "missing_branches": narrowed_nums.n_missing_branches,
```
nedbat, line 160: When I change this to `"xmissing_branches"`, no tests fail, so it seems like we're missing a test of regions with branches?

nedbat: I can take over finishing this.

devdanzin: Feel free to, I probably won't have time to work on it today and would love to see it land.

nedbat:
This work is finished in commit ab609ef0fb235454050cf8383427ce5f1b0ec8e9.

nedbat:
This is now released as part of [coverage 7.6.0](https://pypi.org/project/coverage/7.6.0).

PR #1823, opened 2024-07-27
Remove usage of distutils in setup.py (fix #1820)

### LINKED ISSUE #1820: Setup.py attempts to import obsolete 'distutils'
The Python 'distutils' library has been [deprecated for a long time](https://peps.python.org/pep-0632/), is [prominently documented as obsolescent](https://docs.python.org/3.11/distutils/index.html), and [as of Python 3.12, the 'distutils' import is no longer available](https://docs.python.org/3/whatsnew/3.12.html#distutils).

> Note
> The entire distutils package has been deprecated and will be removed in Python 3.12. This documentation is retained as a reference only, and will be removed with the package. See the [What’s New](https://docs.python.org/3.11/whatsnew/3.10.html#distutils-deprecated) entry for more information.

The Coverage.py `setup.py` program still imports from `distutils`; this relies on a temporary magic work-around in the `setuptools` library.

The `setup.py` should be improved to no longer attempt to import `distutils` at all.

devdanzin:
This PR substitutes import and usage of distutils in setup.py by importing and using setuptools, due to distutils being deprecated/removed.

Fixes #1820.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,17 +3,14 @@
 
 """Code coverage measurement for Python"""
 
-# Distutils setup for coverage.py
+# Setuptools setup for coverage.py
 # This file is used unchanged under all versions of Python.
 
 import os
 import sys
 
-# Setuptools has to be imported before distutils, or things break.
-from setuptools import setup
-from distutils.core import Extension  # pylint: disable=wrong-import-order
+from setuptools import Extension, errors, setup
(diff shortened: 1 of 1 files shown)


nedbat:
Thanks! It looks like this still makes our distributions successfully (https://github.com/nedbat/coveragepy/actions/runs/10123548027), but I'm wondering how to test some of the error handling in setup.py now?

PR #1824, opened 2024-07-28
Test error handling in setup.py

devdanzin:
This PR adds some tests for exception handling in setup.py.

It currently uses direct monkey-patching, please let me know if using `mock,patch` or some other method would be better/cleaner.

I'd be happy to add or remove some of the testing done, as well as reworking or otherwise polishing this PR.

### DIFF
--- a/tests/test_setup.py
+++ b/tests/test_setup.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 
 import sys
 
-from typing import List, cast
+from typing import Any, Callable, List, cast
 
 import pytest
 import setuptools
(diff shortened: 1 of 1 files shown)


tests/test_setup.py
```diff
@@ -27,6 +29,12 @@ def setUp(self) -> None:
         # Force the most restrictive interpretation.
         self.set_environ('LC_ALL', 'C')
 
+    def raise_error(self, error: type[Exception]) -> Callable[[Any], Any]:
```
nedbat, line 32: This seems like something to use unittest.mock or pytest's monkeypatch for.  But see below.

tests/test_setup.py
```diff
@@ -58,3 +66,47 @@ def test_more_metadata(self) -> None:
         assert len(long_description) > 7
         assert long_description[0].strip() != ""
         assert long_description[-1].strip() != ""
+
+    def test_build_extension(self) -> None:
+        # Do we handle all expected errors?
+        for ext_error in setup.ext_errors:
+            with pytest.raises(setup.BuildFailed):
+                setup.build_ext.build_extension = self.raise_error(ext_error)
```
nedbat, line 74: These tests are mostly checking that we catch certain errors.  I'd be more interested in tests that demonstrated actual failure scenarios and that we handled them well, but that seems harder to do.  I'm not sure we need these tests.

devdanzin: Agreed, closing. If I figure out a way to construct actual failure scenarios I'll propose another PR with realistic tests. Thanks for taking a look!

PR #1828, opened 2024-07-30
Fix incorrect backslash before multi-lines string

LiuYinCarl:
The PR is used to fix incorrect backslash before multi-lines string which will result in incorrect lines alignment.

## Source code for test
```python
#################### case 1
a = ["aaa",\
     "bbb \
     ccc"]

#################### case 2
b = ("a",\
     "b \
     c")

########## case 3
c = "xxx"\
    "yyy \
    zzz"

#################### case 4
def f(a, b):
    pass

f(100,\
  "hello \
  world")

#################### case 5(good case)
d = """\
hello
"""


def show_incorrect_line_align():
    # line 1
    # line 2
    # line 3
    # line 4
    # line 5
    pass
```

## Run commands
```bash
coverage run main.py
coverage html
cd htmlcov
python3 -m http.server 9000
```

image

### DIFF
--- a/coverage/phystokens.py
+++ b/coverage/phystokens.py
@@ -57,7 +57,15 @@ def _phys_tokens(toks: TokenInfos) -> TokenInfos:
                 if last_ttext.endswith("\\"):
                     inject_backslash = False
                 elif ttype == token.STRING:
-                    if "\n" in ttext and ttext.split("\n", 1)[0][-1] == "\\":
+                    if last_line.endswith(last_ttext+"\\\n"):
+                        # Deal with special cases like such code::
+                        #
+                        #   a = ["aaa",\
+                        #        "bbb \
+                        #        ccc"]
+                        #
+                        pass
+                    elif "\n" in ttext and ttext.split("\n", 1)[0][-1] == "\\":
                         # It's a multi-line string and the first line ends with
                         # a backslash, so we don't need to inject another.
                         inject_backslash = False


LiuYinCarl:
@nedbat Hi,do you have time to take a kook at this PR? Thanks.

nedbat:
I will take a look, but it might take me some time.

nedbat:
BTW, what name should I use for the Changelog and Contributors files?

nedbat:
Tests added in dc819ff4
Changelog updated in da1682fd
Thanks!

nedbat:
This is now released as part of [coverage 7.6.1](https://pypi.org/project/coverage/7.6.1).

the-13th-letter:
> Tests added in dc819ff4d2333dfbf01c37406bb73ff1aceb8b3a

I stumbled across this while reviewing #1838. I think the tests (and this fix) are incomplete. 

https://github.com/nedbat/coveragepy/blob/dc819ff4d2333dfbf01c37406bb73ff1aceb8b3a/tests/test_html.py#L1134-L1154

In lines 1146–1154, you are testing whether the HTML output correctly reproduces the backslash in the `"bbb ccc"` string. However, the rendered text does not contain the backslash and linebreak within line 1137. This leads to a missing linebreak, and therefore to misaligned coverage information similar to #1836, just in the other direction.

https://github.com/nedbat/coveragepy/blob/dc819ff4d2333dfbf01c37406bb73ff1aceb8b3a/tests/test_phystokens.py#L101-L117

Testing the same input, just with the tokenizer instead of the HTML report renderer. You can see in line 112 that there is a `ws` token with many spaces (which doesn't occur in the tokenizer input) instead of two tokens with a backspace continuation (which does occur). Again, this leads to a missing linebreak, and to misaligned coverage info for all following lines.

---

Concerning this pull request in general:

- Python 3.11 (Linux) with coverage 7.6.1 (plus commit 01779db1fea8d98acb0a15ccf229768c5851b64a from #1838 applied manually): I see the "before fix"/incorrect version for the `a` variable and the "after fix"/correct version for the `b`, `c`, `d` and `f` variables. The same happens whether I use single-quoted strings, triple-quoted strings, or triple-quoted f-strings.
- Python 3.12 (Linux) with coverage 7.6.1 (plus commit 01779db1fea8d98acb0a15ccf229768c5851b64a): I see the incorrect version for the `a` variable if and only if I'm *not* using triple-quoted f-strings, and the correct version in all other cases/for all other variables.

the-13th-letter:
> I stumbled across this while reviewing #1838. I think the tests (and this fix) are incomplete.

My apologies, I forgot to ping @nedbat, as the author of these tests.

See also https://github.com/nedbat/coveragepy/pull/1838#issuecomment-2311174029 for a sample file-plus-coverage-output of a run which is *correct* according to (my reading of) the aforementioned test functions, despite misaligning coverage info later on in the file. Screenshot:

(screenshot)

nedbat:
I appreciate all the work you are doing. I'm trying to keep up!  You said:
> In lines 1146–1154, you are testing whether the HTML output correctly reproduces the backslash in the "bbb ccc" string. However, the rendered text does not contain the backslash and linebreak within line 1137.

When I print the HTML created by this test, I see:
```html
<main id="source">
    <p class="run"><span class="n"><a id="t1" href="#t1">1</a></span><span class="t"><span class="nam">a</span> <span class="op">=</span> <span class="op">[</span><span class="str">"aaa"</span><span class="op">,</span><span class="xx">\</span>&nbsp;</span><span class="r"></span></p>
    <p class="pln"><span class="n"><a id="t2" href="#t2">2</a></span><span class="t">     <span class="str">"bbb \</span>&nbsp;</span><span class="r"></span></p>
    <p class="pln"><span class="n"><a id="t3" href="#t3">3</a></span><span class="t"><span class="str">     ccc"</span><span class="op">]</span>&nbsp;</span><span class="r"></span></p>
</main>
```
Line 1 ends with backslash, and is separate from line 2.  Can you show me more specifically what you mean?

LiuYinCarl:
@nedbat maybe it's because the code of @the-13th-letter contain a  blank between "," and "\\", I fix it in https://github.com/nedbat/coveragepy/pull/1838/commits/a1aa6b8b79ec9c3bd639c2bf20f3148327d229a7

nedbat:
To help with these kinds of tests, I added a new helper to extract the rendered test from an HTML report in 9494ba76.

the-13th-letter:
@netbat:
> When I print the HTML created by this test, I see:
> 
> ```
> <main id="source">
>     <p class="run"><span class="n"><a id="t1" href="#t1">1</a></span><span class="t"><span class="nam">a</span> <span class="op">=</span> <span class="op">[</span><span class="str">"aaa"</span><span class="op">,</span><span class="xx">\</span>&nbsp;</span><span class="r"></span></p>
>     <p class="pln"><span class="n"><a id="t2" href="#t2">2</a></span><span class="t">     <span class="str">"bbb \</span>&nbsp;</span><span class="r"></span></p>
>     <p class="pln"><span class="n"><a id="t3" href="#t3">3</a></span><span class="t"><span class="str">     ccc"</span><span class="op">]</span>&nbsp;</span><span class="r"></span></p>
> </main>
> ```
> 
> Line 1 ends with backslash, and is separate from line 2. Can you show me more specifically what you mean?

Currently, the `test_bug_1828` function does not inspect or assert anything about the first line of the output; it only checks that the `"bbb \<newline>ccc"` part of the output is as expected. So both your rendition *and* my rendition would pass this test, even though *my* rendition has the "missing linebreak/coverage info misalignment" issue. That's why I'm arguing that the tests are incomplete, because methinks they should be explicitly asserting the comma-(space-)backslash in the first line too.

---

I'm going to be busy at `$WORK` for the next couple of days, and will only be able to return to debugging this issue on Saturday (Aug 31st).

> @nedbat maybe it's because the code of @the-13th-letter contain a blank between "," and "\\", I fix it in a1aa6b8b79ec9c3bd639c2bf20f3148327d229a7

Sounds plausible. I'll try that out too on Saturday.

nedbat:
> Currently, the test_bug_1828 function does not inspect or assert anything about the first line of the output;

I changed that this morning: now it makes an assertion about all of the rendered text.  I hope we can write more tests using the same technique.

the-13th-letter:
> > Currently, the test_bug_1828 function does not inspect or assert anything about the first line of the output;
> 
> I changed that this morning: now it makes an assertion about all of the rendered text. I hope we can write more tests using the same technique.

Wonderful. I've taken the liberty of extending `test_bug_1828` to cover more cases of whitespace-then-continuation-backslash.  On current HEAD (28d22a3e9f4bd72c87f18c9a4952e642fef6ce3f), I cannot get this test to break anymore, and it checks *all* the rendered text(-contents) against our expected output.  Please see the attached patch, which passes all linting and all pytest tests (though perhaps not all code style checks) on Python 3.11 and 3.12.

(Installing 3.8, 3.9, 3.10 or PyPy on my system – *with* development headers – is quite a pain, so I only checked 3.11 and 3.12.)

(Yes, I know, GitHub has proper pull requests, but this would be a pull request on a pull request, and it feels a little overkill for something like this.)
[issue_1828_more_whitespace_tests.diff.txt](https://github.com/user-attachments/files/16830521/issue_1828_more_whitespace_tests.diff.txt)

PR #1838, opened 2024-08-26
fix backslash in f-string

### LINKED ISSUE #1836: multiline f-strings with backslash continuation cause extra lines and misaligned coverage info in HTML report
**Describe the bug**

When a line of code ends on a multiline f-string that contains backslash continuation in the first line, the coverage HTML report inserts a synthetic line containing the backslash continuation's backslash at the same position (i.e. a line with a lot of whitespace, and then a backslash), following that line. This also increases the total line count. All following lines with coverage info are then misaligned, because the annotation gets assigned to the original line numbers, not the shifted ones.

**To Reproduce**

(Running Python 3.12.4 on Linux, in a fresh venv, using coverage 7.6.1 installed via `pip install 'coverage === 7.6.1'`.)

**bug.py**
~~~~ py
prog_name = 'bug.py'
err_msg = f"""\
{prog_name}: ERROR: This is the first line of the error.
{prog_name}: ERROR: This is the second line of the error.
\
{prog_name}: ERROR: This is the third line of the error.
"""
if (lambda: None)():  # just some non-constant always-false predicate
    'This line intentionally is not covered.'
'This line is covered again.'
~~~~

**Shell commands**
~~~~ shell-session
$ coverage run --branch python3.12 bug.py && coverage html
Wrote HTML report to htmlcov/index.html
~~~~

Screenshot of the output:

(screenshot)

Note the synthetic line 3, the misaligned partial/missing coloration on lines 8 and 9 (belonging to the code on lines 9 and 10), and the completely absent coloring of line 11. Note further that line 5 (6 in the screenshot) does *not* get its own synthetic line. Note finally that the syntax highlighting does not mark lines 2–8 as a string.

**Expected behavior**

I would expect the HTML rendering of the code `bug.py` to match the actual code of `bug.py`.

**Additional context**

- I could not reproduce this behavior if a b-string, u-string or r-string is used, but I *could* reproduce it with fr- and rf-strings.
- The other report types (JSON, XML, LCOV, text) are fine. In particular, the text-annotated version contains no synthetic lines, and all markers are at the right place.
- I tried switching to the Python trace function (`run --timid`) instead of the C trace function, and also tried the 3.12 `sysmon` core, but that didn't change anything. Which strengthens my belief that the collected data is fine, just the HTML reporter is not.

LiuYinCarl:
Fix https://github.com/nedbat/coveragepy/issues/1836

### DIFF
--- a/coverage/phystokens.py
+++ b/coverage/phystokens.py
@@ -57,10 +57,10 @@ def _phys_tokens(toks: TokenInfos) -> TokenInfos:
                 if last_ttext.endswith("\\"):
                     inject_backslash = False
                 elif ttype == token.STRING:
-                    if last_line.endswith(last_ttext + "\\\n"):
+                    if last_line.endswith("\\\n") and last_line.rstrip(" \\\n").endswith(last_text):
                         # Deal with special cases like such code::
                         #
-                        #   a = ["aaa",\
+                        #   a = ["aaa",\ # there may be zero or more blanks between "," and "\".
                         #        "bbb \
                         #        ccc"]
                         #
--- a/coverage/phystokens.py
+++ b/coverage/phystokens.py
@@ -57,7 +57,8 @@ def _phys_tokens(toks: TokenInfos) -> TokenInfos:
                 if last_ttext.endswith("\\"):
                     inject_backslash = False
                 elif ttype == token.STRING:
-                    if last_line.endswith("\\\n") and last_line.rstrip(" \\\n").endswith(last_ttext):
+                    if (last_line.endswith("\\\n") and
+                        last_line.rstrip(" \\\n").endswith(last_ttext)):
                         # Deal with special cases like such code::
                         #
                         #   a = ["aaa",\ # there may be zero or more blanks between "," and "\".
--- a/coverage/phystokens.py
+++ b/coverage/phystokens.py
@@ -57,7 +57,7 @@ def _phys_tokens(toks: TokenInfos) -> TokenInfos:
                 if last_ttext.endswith("\\"):
                     inject_backslash = False
                 elif ttype == token.STRING:
-                    if last_line.endswith("\\\n") and last_line.rstrip(" \\\n").endswith(last_text):
+                    if last_line.endswith("\\\n") and last_line.rstrip(" \\\n").endswith(last_ttext):
                         # Deal with special cases like such code::
                         #
                         #   a = ["aaa",\ # there may be zero or more blanks between "," and "\".
--- a/coverage/phystokens.py
+++ b/coverage/phystokens.py
@@ -69,6 +69,9 @@ def _phys_tokens(toks: TokenInfos) -> TokenInfos:
                         # It's a multi-line string and the first line ends with
                         # a backslash, so we don't need to inject another.
                         inject_backslash = False
+                elif sys.version_info >= (3, 12) and ttype == token.FSTRING_MIDDLE:
+                    if ttext.split("\n", 1)[0][-1] == "\\":
+                        inject_backslash = False
                 if inject_backslash:
                     # Figure out what column the backslash is in.
                     ccol = len(last_line.split("\n")[-2]) - 1
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -1207,7 +1207,7 @@ assert len(math) == 18
         "leader", ["", "f", "r", "fr", "rf"],
         ids=["string", "f-string", "raw_string", "f-raw_string", "raw_f-string"]
     )
-    def test_bug_1836(self, leader) -> None:
+    def test_bug_1836(self, leader: str) -> None:
         # https://github.com/nedbat/coveragepy/issues/1836
         self.make_file("py312_fstrings.py", f"""\
             prog_name = 'bug.py'
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -1203,6 +1203,36 @@ assert len(math) == 18
             '3     ccc"]',
             ]
 
+    @pytest.mark.parametrize(
+        "leader", ["", "f", "r", "fr", "rf"],
+        ids=["string", "f-string", "raw_string", "f-raw_string", "raw_f-string"]
+    )
+    def test_bug_1836(self, leader) -> None:
+        # https://github.com/nedbat/coveragepy/issues/1836
+        self.make_file("py312_fstrings.py", f"""\
+            prog_name = 'bug.py'
+            err_msg = {leader}'''\\
+            {{prog_name}}: ERROR: This is the first line of the error.
+            {{prog_name}}: ERROR: This is the second line of the error.
+            \\
+            {{prog_name}}: ERROR: This is the third line of the error.
+            '''
+            """)
+
+        cov = coverage.Coverage()
+        py312_fstrings = self.start_import_stop(cov, "py312_fstrings")
+        cov.html_report(py312_fstrings)
+
+        assert self.get_html_report_text_lines("py312_fstrings.py") == [
+            "1" + "prog_name = 'bug.py'",
+            "2" + f"err_msg = {leader}'''\\",
+            "3" + "{prog_name}: ERROR: This is the first line of the error.",
+            "4" + "{prog_name}: ERROR: This is the second line of the error.",
+            "5" + "\\",
+            "6" + "{prog_name}: ERROR: This is the third line of the error.",
+            "7" + "'''",
+            ]
+
     def test_unicode(self) -> None:
         surrogate = "\U000e0100"
 


the-13th-letter (commented):
With these changes (applied manually to 7.6.1), the synthetic extra lines in #1836 are now gone. (Tested with Python 3.11 and 3.12.)

LiuYinCarl:
@the-13th-letter Thanks for your careful tests in https://github.com/nedbat/coveragepy/pull/1828#issuecomment-2310266578 !

But I don't understand very much because I am not a native English speaker.

In https://github.com/nedbat/coveragepy/pull/1828#issuecomment-2310266578, do you mean the same as the table below ?

| Python Version | coverage version | use PR1838 | Still has bug |
|--- |--- |--- | --- | 
| 3.11 | 7.6.1 | No | No |
| 3.11 | 7.6.1 | Yes | No |
| 3.12 | 7.6.1 | No | Yes |
| 3.12 | 7.6.1 | Yes | No |

I tested these four scenarios below and It seems like this PR(#1838) can fix your problem in both Python3.11 and Python3.12.

image

image

image

image


So can I understand you like this:
1. This PR(#1838) can fix your problem now.
2. You think coveragepy need more test to include your bug report about f-string.

Please let me know if I'm wrong.

the-13th-letter:
> @the-13th-letter Thanks for your careful tests in [#1828 (comment)](https://github.com/nedbat/coveragepy/pull/1828#issuecomment-2310266578) !
> 
> But I don't understand very much because I am not a native English speaker.
> 
> In [#1828 (comment)](https://github.com/nedbat/coveragepy/pull/1828#issuecomment-2310266578), do you mean the same as the table below ?
> 
> [...]

@LiuYinCarl, no, that's not what I mean. In #1828, I can trigger the opposite effect: backslash continuation sometimes disappears in the HTML report, so a line break is missing, and coverage info gets "shifted upwards".

Specifically, in the following pytest-style file `test_bug.py`, the HTML report contains synthetic *joined* lines (e.g. the first two lines of the `x` variable are one line in the HTML report) by removing a backslash continuation from the input.

<details><summary><code>test_bug.py</code></summary>

~~~~ py
def test_bug():
    prog_name = 'bug.py'
    err_msg = f"""\
{prog_name}: ERROR: This is the first line of the error.
{prog_name}: ERROR: This is the second line of the error.
\
{prog_name}: ERROR: This is the third line of the error.
"""
    if (lambda: None)():  # just some non-constant always-false predicate
        'This line intentionally is not covered.'
    'This line is covered again.'

    # test cases from https://github.com/nedbat/coveragepy/pull/1828
    def f(a, b):
        pass
    a = ["aaa", \
         "bbb \
         ccc"]
    b = ("a",\
         "b \
         c")
    c = "xxx"\
        "yyy \
        zzz"
    f(100,\
      "hello \
      world")
    d = """\
    hello
    """

    # work-alikes
    x = ["""aaa""", \
         """bbb \
         ccc"""]
    y = ("""a""",\
         """b \
         c""")
    z = """xxx"""\
        """yyy \
        zzz"""
    f(100,\
      """hello \
      world""")
    A = [f"""aaa""", \
         f"""bbb \
         ccc"""]
    B = (f"""a""",\
         f"""b \
         c""")
    C = f"""xxx"""\
        f"""yyy \
        zzz"""
    f(100,\
      f"""hello \
      world""")
    D = f"""\
    hello
    """
    AA = [f"""\
          aaa""", \
          f"""\
          bbb \
          ccc"""]
    BA = [f"""\
          {x}""", \
          f"""\
          {y} \
          {z}"""]

    if (lambda: None)():
        'This is not covered.'
    'This is covered.'

if __name__ == '__main__':  # if not running under pytest
    test_bug()







~~~~

</details>

(screenshot)

(See the attached ZIP file for full coverage output: Py3.11 unpatched, Py3.11 patched with 01779db1fea8d98acb0a15ccf229768c5851b64a, Py3.12 unpatched, and Py3.12 patched with 01779db1fea8d98acb0a15ccf229768c5851b64a. I do not know why the same behavior does not also happen with your `cov.py` file contents.)

All of the above has nothing to do with this PR, which is why I commented on #1828 instead of here. The only connection is that while reviewing this PR, I was reading up on #1828 to understand the surrounding code and comments. And then I realized that the tests added in dc819ff4d2333dfbf01c37406bb73ff1aceb8b3a looked weird, so I commented on that.

> So can I understand you like this:
> 
> 1. This PR([fix backslash in f-string #1838](https://github.com/nedbat/coveragepy/pull/1838)) can fix your problem now.

Yes.

> 2. You think coveragepy need more test to include your bug report about f-string.

I was trying to make a different point: I think that the tests in dc819ff4d2333dfbf01c37406bb73ff1aceb8b3a are wrong. They seem to expect HTML-rendered output or tokenizer output that is clearly different from the input.

the-13th-letter:
The attached patch contains a pytest test function to check for #1836.
[issue_1836_tests.diff.txt](https://github.com/user-attachments/files/16830534/issue_1836_tests.diff.txt)

the-13th-letter (approved):
This code passes [my recent test file](https://github.com/nedbat/coveragepy/pull/1838#issuecomment-2323499750) on Python 3.11 and 3.12.

LGTM.

nedbat:
@LiuYinCarl @the-13th-letter Does one of you want to apply that new test to this pull request?  Then we'd be able to merge it.

@the-13th-letter Why use `{{prog_name}}` in the test?  It seems like you intended it to be substituted, but it's not, and now you have literal `{prog_name}` in the results.   You can make a pull request into this pull request to apply the test to it.

nedbat:
@the-13th-letter also, maybe we can do without `{quote}` by using triple-single-quotes?  The escaping is a lot to puzzle through.

nedbat:
Sorry, I went ahead and made a pull request with the test: https://github.com/LiuYinCarl/coveragepy/pull/1

LiuYinCarl:
@nedbat, your PR was merged and I add a annotation in this line to make it pass the tox checks.  https://github.com/nedbat/coveragepy/pull/1838/files#diff-29c93c7c1984a5716eb3a992a5e15cdcf5b7471539f0cc2831f141cc780a0885R1210

nedbat:
I cleaned up the commits and merged this as commit 515d99c6 and a77bba7b.  Thanks!

the-13th-letter:
Just to follow up:

> @the-13th-letter Why use `{{prog_name}}` in the test? It seems like you intended it to be substituted, but it's not, and now you have literal `{prog_name}` in the results.

It's a transcription of my original `bug.py` sample file in the original post of #1836, which included a "true" f-string. Since we're generating that as an output, and using an f-string to format the "f"/"fr"/whatever string leader, interpolations in the inner f-string like `{prog_name}` need to be double-escaped, as `{{prog_name}}`. It's very much on purpose, though admittedly born of the perhaps pointless desire of reusing the original demonstration file for this bug.

> @the-13th-letter also, maybe we can do without `{quote}` by using triple-single-quotes? The escaping is a lot to puzzle through.

Sure, go ahead. It'd be a style violation in my corner of the world to not use double quotes, but of course that doesn't necessarily apply here. (It does mean that this is the first solution that comes to mind, though.)

nedbat:
Looking at the [coverage reports for phystokens.py](https://htmlpreview.github.io/?https://raw.githubusercontent.com/nedbat/coverage-reports/main/reports/20240907_911678aa5f/htmlcov/z_56e48acd472f6ae7_phystokens_py.html#t59), there are some partial branches in the new checks:

image


Is that because it's impossible for those conditions to be false, or because we are missing some test cases in the test suite?  Either way, it would be great to resolve so that we don't have partial branches in the coverage report.

LiuYinCarl:
@nedbat It's my opinion below.

It is impossible for the condition on line 74 to be false because `if last_line and last_line.endswith("\\\n"):`
prevent other FSTRING_MIDDLE which is not end with `\\\n` into the condition.

I'm not sure about line 69, but I have tested the cases in https://github.com/nedbat/coveragepy/pull/1828 and https://github.com/nedbat/coveragepy/pull/1838, and it is impossible for any tests on line 69 to fail.

So, I think we can just replace the two conditions with `inject_backslash = False`

the-13th-letter:
@nedbat I too am mostly convinced that the condition on line 74 must always be true because of the `if last_line and last_line.endswith("\\\n")` guard on line 40. I would probably settle on leaving the condition inside the code, as an assertion.

Like @LiuYinCarl, I'm not so sure about line 69 either. I did however try to design (failing) test cases for that condition, without success. Thus, pragmatically, I would convert that condition into an assertion as well, and leave a note à la "We failed to design a test case that doesn't fulfill this condition.".

<details><summary>attempted test inputs that don't improve branch coverage</summary>

```py3
space = " "
string = f"""\
line{space}1\
 and still line{space}1
line 2 \
and still line 2\
 and yet still line 2
line{space}3\
{space}and still line 3
"""
```

```py3
space = " "
weird_formatting = {"empty_string": """\
""", "nonempty_string": """ \
\

""",\
"chunked_string": "hello"\
" "\
"world"}
weird_formatting_with_fstrings = {f"empty_string": f"""\
""", f"nonempty_string": f""" \
\

""",\
f"chunked_string": f"hello"\
f"{space}"\
f"world"}
```

</details>

nedbat:
Thanks to both of you for continuing to work on this.  I really appreciate it! <3

nedbat:
Updated in commit 595ea22bbebcc84884c1401628a6353d2f10e222, coverage changed from 95.256% to 95.266% :)

nedbat:

PR #1843, opened 2024-09-02
fix: null array indexing in CTracer_handle_return

### LINKED ISSUE #1835: null array indexing in coverage/ctracer/tracer.c
This issue was found by running [UndefinedBehaviourSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) in our internal builds at Google: coverage/ctracer/tracer.c's `CTracer_handle_return()` is sometimes called in situations where `self->pdata_stack->stack` is NULL after the call to `CTracer_set_pdata_stack()`, but `CTracer_handle_return()` still tries to index the stack to set `self->pcur_entry`. As far as I can tell `self->pdata_stack->depth` is always `-1` in that case, meaning the NULL array gets indexed with a non-0 index (that would also be out of bounds). I can't tell if this is intentional or not (i.e. whether the resulting value of `self->pcur_entry` actually matters.)

Here's a diff to reproduce the error without ubsan:
```
--- a/coverage/ctracer/tracer.c
+++ b/coverage/ctracer/tracer.c
@@ -722,6 +722,11 @@ CTracer_handle_return(CTracer *self, PyFrameObject *frame)
     if (CTracer_set_pdata_stack(self) < 0) {
         goto error;
     }
+    if (self->pdata_stack->stack == NULL) {
+        fprintf(stderr, "stack = NULL, depth = %d\n",
+                self->pdata_stack->depth);
+        abort();
+    }
     self->pcur_entry = &self->pdata_stack->stack[self->pdata_stack->depth];

     if (self->pdata_stack->depth >= 0) {
```

frigus02:
Fixes #1835 

`CTracer_set_pdata_stack` can initialize `self.pdata_stack` to an empty stack where `->stack` is `NULL` and `->deph` is -1.

Move index into `->stack` into `->depth>=0` check to avoid indexing into `NULL` array.

This issue was found running UndefinedBehaviourSanitizer. It's reproducible in regular test runs, e.g. `python3 -m tox -e py311`. Adding `if (self->pdata_stack->stack == NULL) { fprint(...) }` before the moved line shows the issue in several test cases.

---

The contributing guide mentions ever fix should include tests. I'd be happy to add some, but I'm not quite sure how to do that here. Do you have a preference/advice?

### DIFF
--- a/coverage/ctracer/tracer.c
+++ b/coverage/ctracer/tracer.c
@@ -722,9 +722,9 @@ CTracer_handle_return(CTracer *self, PyFrameObject *frame)
     if (CTracer_set_pdata_stack(self) < 0) {
         goto error;
     }
-    self->pcur_entry = &self->pdata_stack->stack[self->pdata_stack->depth];
 
     if (self->pdata_stack->depth >= 0) {
+        self->pcur_entry = &self->pdata_stack->stack[self->pdata_stack->depth];
         if (self->tracing_arcs && self->pcur_entry->file_data) {
             BOOL real_return = FALSE;
             pCode = MyCode_GetCode(MyFrame_GetCode(frame));


nedbat:

PR #1847, opened 2024-09-06
[WIP] Overhaul lcov reports

### LINKED ISSUE #1846: lcov reports should not use line number zero in BRDA: records
**Describe the bug**

`coverage lcov` can generate `BRDA:` records with line number zero.  Current versions of lcov's `genhtml` reject such lines:

```
genhtml: ERROR: line 0 of .../sourcefile.py has branchcov but no linecov data
```

**To Reproduce**

The bad BRDA records are coming from here: <https://github.com/nedbat/coveragepy/blob/5467e1f0c9e16ef5456ecccb56993b15c34b7a3e/coverage/lcovreport.py#L108-L112>

I do not understand how to hit this case.  The commentary, and the description of what negative numbers mean in arc records (<https://coverage.readthedocs.io/en/7.6.1/api_coveragedata.html#coverage.CoverageData.arc>) make it sound like unexecuted conditional return statements could cause the problem.  Running `coverage run --branch test.py` on this file

```
def f(x):
  if x < 0:
    return
  print("got here")

f(0)
f(1)
```

produces a `.coverage` database whose `arc` table does contain negative numbers, but running `coverage lcov` on that database produces 

```
BRDA:3,0,0,-
BRDA:4,0,1,1
BRF:2
BRH:1
```

which does not use line zero and looks correct to me (well, modulo the fact that a human would say there isn't a branch on line 4? but that's a separate issue).

Looking at the database for the real program that produced the actual bad lcov-format report that prompted this issue, I see lots more cases of negative numbers in the arc table, but the associated _positive_ numbers are too large to be actual line numbers in the associated file, which means I don't actually understand what the records in the arc table _mean_, so I'm stuck.

**Expected behavior**

`coverage lcov` should generate a .lcov file that `genhtml` does not reject.

zackw:
Fixes #1846 (per my understanding of the lcov file format) and addresses several other issues.  See individual commit logs for details.

Tested only with Python 3.11 and 3.12; I do not have convenient access to anything older.

I plan to come back to this branch next week and add a bunch more tests after comparing the reports generated on this branch to the troublesome reports I was getting that prompted #1846.  Please do not merge yet.  I'm filing this PR now so you know I'm working on it.

### DIFF
--- a/coverage/config.py
+++ b/coverage/config.py
@@ -250,6 +250,7 @@ class CoverageConfig(TConfigurable, TPluginConfig):
 
         # Defaults for [lcov]
         self.lcov_output = "coverage.lcov"
+        self.lcov_checksums = "off"
 
         # Defaults for [paths]
         self.paths: dict[str, list[str]] = {}
@@ -428,6 +429,7 @@ class CoverageConfig(TConfigurable, TPluginConfig):
 
         # [lcov]
         ("lcov_output", "lcov:output"),
+        ("lcov_checksums", "lcov:checksums")
     ]
 
     def _set_attr_from_config_option(
--- a/coverage/lcovreport.py
+++ b/coverage/lcovreport.py
@@ -36,6 +36,7 @@ class LcovReporter:
 
     def __init__(self, coverage: Coverage) -> None:
         self.coverage = coverage
+        self.config = coverage.config
         self.total = Numbers(self.coverage.config.precision)
 
     def report(self, morfs: Iterable[TMorf] | None, outfile: IO[str]) -> float:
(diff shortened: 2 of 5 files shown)


zackw:
I will, of course, also fix all the test failures :-/

nedbat:
Thanks for starting this.  As some first comments after a quick spin through it:
- Please try to follow the formatting already in the rest of the code: multi-line bracketed expressions should have a newline after the opening bracket, and don't indent to the opening bracket.  Indent once.
- There's a lot of discussion in the configuration page that should be moved the `lcov` command description.
- Do we really need the different styles of checksumming? I haven't seen that in lcov before.

zackw:
> * Do we really need the different styles of checksumming? I haven't seen that in lcov before.

I have the impression that the lcov team considers the per-line checksums that `coverage lcov` currently generates, to be deprecated in favor of the newer `--version-script` mechanism, which is how the lcov tools generate VER: lines.  This is mostly going by what it says about per-line checksums in the `geninfo` manpage, particularly "checksum generation is disabled by default" and "this option is somewhat subsumed by the --version-script option".

There _is_ a way to use the lcov tools to add VER: lines to a .lcov file that doesn't have them, but then you have to have both coverage.py and the lcov tools installed in the environment that's generating coverage data, which may be undesirable.

Instead of what I did, I'd be fine with either

- only adding an option to turn off generation of per-line checksums
- replacing the "file" mode with a mode in which `coverage lcov` runs an lcov-spec version script to generate VER: lines

Please let me know what you would prefer.

nedbat:
Can we use the new mechanism with no choice? I'd rather not add options where people don't strongly need an option.

zackw:
> Can we use the new mechanism with no choice?

We would have to at least offer choice of version script.  I read over the lcov docs some more and I'm now convinced that what I implemented - i.e. the hardcoded base64(sha256(file contents)) - was a mistake, the intention of the VER: feature is to allow people to pull in all sorts of contextual information (e.g. `git describe`) that _they_ feel is relevant to whether the lcov file matches the source code.

For purpose of _this_ pull request, I suggest that either turning off the per-line checksums unconditionally, or giving an option to turn them off, is the right thing, and generation of VER: lines can be a future thing if anyone wants them.  I'd prefer the option because that gives a smoother backcompat story.  I don't want to just leave them alone because (a) they appear to be semi-deprecated in lcov, and (b) they make it harder to compare the result of `coverage lcov` with the result of `coverage xml` followed by `xml2lcov`.

zackw:
Superseded by #1849.

PR #1849, opened 2024-09-09
Lcov report fixes v2

### LINKED ISSUE #1846: lcov reports should not use line number zero in BRDA: records

zackw:
Fixes #1846 (per my understanding of the lcov file format) and addresses several other issues. See individual commit logs for details. Should be ready to merge.

Changes from #1847:

- Verified that the lcov report for real partially covered code is sensible
- Adjusted handling of BRDA: based on what genhtml does with it on real code
- "File checksums" mode removed; now there's just a toggle for line checksums
- Passes `tox` locally with Python 3.8 through 3.12 and PyPy installed
- I *think* I have fixed the formatting issues you were concerned with; if I still didn't get it quite right please point me at specific sections you'd like formatted differently

### DIFF
--- a/coverage/config.py
+++ b/coverage/config.py
@@ -250,6 +250,7 @@ class CoverageConfig(TConfigurable, TPluginConfig):
 
         # Defaults for [lcov]
         self.lcov_output = "coverage.lcov"
+        self.lcov_line_checksums = False
 
         # Defaults for [paths]
         self.paths: dict[str, list[str]] = {}
@@ -428,6 +429,7 @@ class CoverageConfig(TConfigurable, TPluginConfig):
 
         # [lcov]
         ("lcov_output", "lcov:output"),
+        ("lcov_line_checksums", "lcov:line_checksums", "boolean")
     ]
 
     def _set_attr_from_config_option(
--- a/coverage/env.py
+++ b/coverage/env.py
@@ -99,10 +99,6 @@ class PYBEHAVIOR:
     # Some words are keywords in some places, identifiers in other places.
     soft_keywords = (PYVERSION >= (3, 10))
 
-    # Modules start with a line numbered zero. This means empty modules have
-    # only a 0-number line, which is ignored, giving a truly empty module.
-    empty_is_empty = (PYVERSION >= (3, 11, 0, "beta", 4))
-
     # PEP669 Low Impact Monitoring: https://peps.python.org/pep-0669/
     pep669 = bool(getattr(sys, "monitoring", None))
 
--- a/coverage/lcovreport.py
+++ b/coverage/lcovreport.py
@@ -53,13 +53,26 @@ class LcovReporter:
         self.coverage.get_data()
         outfile = outfile or sys.stdout
 
-        for fr, analysis in get_analysis_to_report(self.coverage, morfs):
+        # ensure file records are sorted by the _relative_ filename, not the full path
+        to_report = [
+            (fr.relative_filename(), fr, analysis)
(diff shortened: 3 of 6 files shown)


nedbat:
Thank you so much for doing all this!  I'm still curious about the checksums.  They used to always be present.  You want to disable them.  There's a new style which we don't support.  Help me understand the lcov community feelings about these checksums.  What if we simply removed them with no option?

zackw:
I don't have any special insight into the lcov community.  I'm just going by what I read in the lcov manpages and a bit of source archaeology.  That said:

- The lcov documentation gives me the impression that its maintainers think it's very important to be sure that the code being used to generate coverage reports matches the code that was used to measure coverage in the first place.  They spend hundreds of words on this in both the genhtml and lcov manpages.  This is the entire purpose of the DA: line checksums and of the newer VER: records.
- Having said that, there has been an option to turn off generation of DA: line checksums ever since they were introduced, in version 1.2 of lcov (released July 2003), and the default behavior changed to "don't generate per-line checksums" in version 1.6 (released July 2007).
- VER: records and the --version-script mechanism are quite new, with lcov 2.0, which came out in May of 2023.  The lcov documentation does not *quite* say these are considered a replacement for DA: line checksums, but it's strongly implied.
 
Putting those three things together, I am fairly confident in saying that DA: line checksums are considered obsolete by the lcov maintainers - but not obsolete enough to drop completely.  Whether `coverage lcov` should drop them completely is something I don't have a sense for.  _I_ would not miss them, but I don't know how many other people might.

I think generation of VER: records by `coverage lcov` can wait until someone comes around asking for it. For what I'm using `coverage lcov` for, I have a different way of ensuring generation-to-analysis consistency. And the implementation I had in #1847 is definitely no good. If `coverage lcov` is to generate VER: records at all, it needs to support _using lcov's version scripts to generate VER: records_, not just make up its own payloads.

nedbat:
Thanks, let's merge this and see if people have concerns.

Issue #1850, opened 2024-09-11
[internal] Make missing arc fragments useful to more report generators


zackw:
This is following on from the discussion near the end of #1846.  `parser.py` has a sophisticated algorithm for characterizing branch arc destinations - ordinary if statement, exiting the function, exiting a `with` clause, etc. - but the way it's presently implemented, it is only useful to the HTML report generator.  To make it useful to more reporters, I'd like to ask for the following:

- A new `FileReporter` API, analogous to the existing `missing_arc_description` function, that tersely describes the destination of an arc *without* any leading text about whether it was or wasn't executed.  For ordinary `if` destinations, it'd be most useful to get just "line {lineno}".  For function returns, "return from {self.name!r}".  And so on.
- Similarly, an exposed API that describes the circumstances where an arc would fail to have been taken - the existing "condition on line {lineno} was never {true/false}", "pattern on line {lineno} {always/never} matched" texts - *in isolation*, without the rest of the missing_arc_description verbiage, would be useful (not directly to the LCOV reporter, but I bet someone will want it eventually).
- ~~And finally, an exposed API that tells you whether an arc should be ignored, *whether or not* it was ever taken or not taken.  This is for things like `if any(pred(x) for x in xs)`, where the uninteresting exit arc from the generator expression gets filtered out of the HTML report by (it seems) always getting executed, but a reporter like LCOV that needs to talk about _all_ the arcs has no good way to tell.~~ (Split to #1852.)

nedbat:
I need to understand more about what you are requesting.  I've been de-emphasizing arcs to focus more on true branches.  I haven't checked the code, but I thought exit arcs from generator expressions were already removed.

zackw:
You know what, this should have been two separate issues.  I've split the exit-arcs-from-generator-expressions item to #1852 and I hope I've explained it better there. Let's focus this issue on the first two items.

The code at stake for the first two items is like this

```python
def fn(pred):
    if pred:
        print("yes")
fn(True)
fn(False)
```

The LCOV reporter (using the code in #1851) currently generates these BRDA: records for this code:

```none
BRDA:2,0,to line 3,1
BRDA:2,0,to exit,1
```

[...]

PR #1851, opened 2024-09-11
Lcov report improvements part 2

zackw:
- break up lcov_file to not be so gigantic
- improved branch destination reporting
- function coverage!
- addresses #1850 and #1852

Because of the refactor in the first commit, functional changes will be easier to see if you go commit by commit.

### DIFF
--- a/coverage/lcovreport.py
+++ b/coverage/lcovreport.py
@@ -63,6 +63,11 @@ def lcov_functions(
     # We continue to generate the old format because we don't know what
     # version of the lcov tools will be used to read this report.
 
+    # "and region.lines" below avoids a crash due to a bug in PyPy 3.8
+    # where, for whatever reason, when collecting data in --branch mode,
+    # top-level functions have an empty lines array.  Instead we just don't
+    # emit function records for those.
+
     # suppressions because of https://github.com/pylint-dev/pylint/issues/9923
     functions = [
         (min(region.start, min(region.lines)), #pylint: disable=nested-min-max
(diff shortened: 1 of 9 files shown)


zackw:
I think this is ready to be reviewed now.  Let me know if you want me to clean up the commit history or anything.

nedbat:
BTW, I'm about to drop Python 3.8, so you can simplify the tests by just skipping them on pypy3.8.

zackw:
Testing whether the existing `@xfail_pypy38` decorator does the trick now.

nedbat:
Looks good, thanks so much! I'll merge it.

nedbat:

coverage/lcovreport.py
```diff
@@ -32,6 +32,129 @@ def line_hash(line: str) -> str:
...
+        # and not-executed arcs to BRDA codes.  It seems to produce reasonable
+        # results when fed through genhtml.
+        _, taken = branch_stats[line]
+
+        if taken == 0:
+            # When _none_ of the out arcs from 'line' were executed,
+            # this probably means 'line' was never executed at all.
+            # Cross-check with the line stats.
+            assert len(executed_arcs[line]) == 0
+            assert line in analysis.missing
```
bgianfo, line 123: Our coverage workflow is now failing due to this line, when running with version `7.6.3`.

nedbat: Can you provide us with a way to reproduce it?  Failing that, can you add this line before the assert, and show us the output?
```py
print(f"{line=}, {analysis.missing=}")
```
(this should have been on the assert line, my bad)

rouge8: I'm seeing this same error too, but only in CI so far. I haven't been able to reproduce it locally so I can't put the `print()` line in either...

PR #1898, opened 2024-11-20
Make branch coverage accessible via the API

### LINKED ISSUE #1888: Adding Branch Coverage to API
Hi! Eleanor from the Python for VS Code extension following up on API updates I was hoping to get which the VS Code python extension needs to support branch coverage! As you know, right now there is no way to get branch coverage from the API so Im looking to implement this but wanted to start the discussion as an issue to make sure I have the right approach.

What I think is the easiest access point to the branch coverage values I am looking for is in the  [Analysis class](https://github.com/nedbat/coveragepy/blob/864ef601b9d7bc1aba07821daf6543081c85486e/coverage/results.py#L80). Here there is the function `_total_branches()` which would provide one value I need and then I would need either covered branches or missing branches. `n_missing_branches` is defined in the `Analysis` class but not accessible, so maybe a getter for this value would be sufficient. (it is my understanding that branches are different from arcs and therefore it is not sufficient to display arc values but if I am mistaken please let me know)

Next is getting the `Analysis` class instance for a given file. From what I can tell, it seems like this is returned in two places,[ _analyze](https://github.com/nedbat/coveragepy/blob/864ef601b9d7bc1aba07821daf6543081c85486e/coverage/control.py#L938-L939) or [analysis_from_file_reporter](https://github.com/nedbat/coveragepy/blob/864ef601b9d7bc1aba07821daf6543081c85486e/coverage/results.py#L23-L24), (which makes sense given that `_analyze` calls `analysis_from_file_reporter`). It seems like `_analyze` under the `Coverage` class is probably easiest, given the comment "Private for now." in its docstring this might have been something already considered.

To summarize, what I am suggesting is:
1. making the `Analysis` class part of the API 
2. adding the `_analyze` function in `Coverage` to the API
3. adding a function to Analysis to surface either covered branches or missing branches

These are quite a few asks so if you have a suggestion that might add less to the API (and less to maintain) please let me know. It is not my intent to bloat the API only get these branch numbers so we could also attempt edits to surface them without the Analysis data class but I am unsure.

For context, [here](https://github.com/eleanorjboyd/vscode-python/blob/main/python_files/unittestadapter/execution.py#L353) is where we collect and transform coverage in the extension. [Here](https://github.com/microsoft/vscode-python/pull/24180/files#diff-9a28b4b19fbb6331c15c76b4752852162785c0d521038cb637d8432901ebead2L456) was my previous idea on how to get branch coverage before I realized it was not part of the API.

Thank you!

eleanorjboyd:
fixes https://github.com/nedbat/coveragepy/issues/1888

one edit from above proposal is I exposed `branch_stats` from the `Analysis` class as it has the (total_exits, taken_exits) which is the info needed and is already public.

another note is Im not entirely sure how to test / view the changes to the `.rst` files for https://coverage.readthedocs.io/ so any advice here would be appreciated. Thanks!

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -926,7 +926,7 @@ class Coverage(TConfigurable):
         coverage data.
 
         """
-        analysis = self._analyze(morf)
+        analysis = self.analyze(morf)
         return (
             analysis.filename,
             sorted(analysis.statements),
@@ -935,7 +935,7 @@ class Coverage(TConfigurable):
             analysis.missing_formatted(),
         )
 
-    def _analyze(self, morf: TMorf) -> Analysis:
+    def analyze(self, morf: TMorf) -> Analysis:
         """Analyze a module or file.  Private for now."""
         self._init()
         self._post_init()
--- a/coverage/report_core.py
+++ b/coverage/report_core.py
@@ -98,7 +98,7 @@ def get_analysis_to_report(
 
     for fr, morf in sorted(fr_morfs):
         try:
-            analysis = coverage._analyze(morf)
+            analysis = coverage.analyze(morf)
         except NotPython:
             # Only report errors for .py files, and only if we didn't
             # explicitly suppress those errors.
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -55,4 +55,5 @@ only.  :ref:`dbschema` explains more.
     api_module
     api_plugin
     api_coveragedata
+    api_analysis
     dbschema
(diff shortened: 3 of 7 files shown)


nedbat:
@eleanorjboyd Sorry I haven't gotten to this.  I'm pondering whether there's anything in `Analysis` I wouldn't want to make public.  I can take care of checking the .rst changes.

nedbat:
Commit 33f12dfa365f3faa349d844e603b0432ba3476d3 has `Coverage.branch_stats()`.

eleanorjboyd:
Hi @nedbat- super excited thank you! I blocked off time this upcoming month to implement this- so hopefully I can get it in quick! Will keep you in the loop as I implement and release!

PR #1901, opened 2024-12-03
Use set instead of list for arcs_executed, fixes poor performance

turibe:
This fixes very poor performance in "coverage.xml" for a large-ish code base, which I traced to the `arcs_missing` procedure. (Could probably improve further using set difference.)

### DIFF
--- a/coverage/results.py
+++ b/coverage/results.py
@@ -178,7 +178,7 @@ class Analysis:
         """Returns a sorted list of the un-executed arcs in the code."""
         missing = (
             p for p in self.arc_possibilities
-                if p not in self.arcs_executed
+                if p not in self.arcs_executed_set
                     and p[0] not in self.no_branch
                     and p[1] not in self.excluded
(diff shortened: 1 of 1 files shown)


nedbat:
Nice, thanks! Do you have any stats on the size of the repo and the speed up?

turibe:
About 350K lines of code, 390K rows in the .coverage arc DB table.

Speedup went from "cant-wait-that-long-will-do-control-C",  to about under a minute, for `coverage xml`. (Will add actual numbers in a bit.)

turibe:
> Nice, thanks! Do you have any stats on the size of the repo and the speed up?

Speedup went from 730 seconds ("cant-wait-that-long-will-do-control-C"), to 13 seconds, for coverage xml.

nedbat:
This is now released as part of [coverage 7.6.9](https://pypi.org/project/coverage/7.6.9).

PR #1913, opened 2025-01-10
Add py3-none-any wheels to the distribution artefacts

### LINKED ISSUE #1912: Add py3-none-any wheel to distribution artefacts

Coverage contains a binary module to provide an optimised path for gathering coverage data. As a result, Coverage is published used platform-specific binary wheels.

If you're in an environment that *doesn't* have an officially published wheel, you need to install from source. 

However, cross-platform builds (such as mobile platforms like iOS and Android) can *require* the use of `--use-binary=:all:` when installing. These platforms need to use `--platform` to force the installation of wheels on a platform other than the one they're currently on; using `--platform` also requires the use of `--use-binary=:all:`, as pip can't guarantee it will be able to run the build entry point for a source distribution.

Coverage should publish a py3-none-any wheel as part of standard release artefacts. This would allow a wheel-based install of a non-optimized version of coverage on platforms that don't have an official wheel. pip will preferentially install platform-specific wheels over py3-none-any wheels, so the optimised binary wheel should take precedence over the non-optimized wheel wherever possible.

Publish optimised binary wheels for platforms affected by this problem. This is likely a better long-term solution; however, at present, cibuildwheel can't produce iOS wheels. We (BeeWare) are currently working on this, but there won't be a fix in place for (at best) a couple of months.

**Additional context**

This problem emerged recently as a result of a change in pip 24.3, which adds iOS support. 

Historically, the process for installing wheels on iOS has involved using a site-packages hack to fake an iOS environment inside a macOS virtual environment, tricking pip into installing iOS wheels with the `--prefer-binary` option. `--prefer-binary` will fall back to a source distributions if no binary can be found, and with the right hacks in `site-packages`, it's possible to force this environment to produce a non-optimized coverage wheel for iOS installation. 

As an immediate workaround, BeeWare is currently publishing iOS binary wheels in its [independent iOS wheel repo](https://anaconda.org/beeware/coverage). However, we'd rather not be an official distributor of coverage in the long term.

The one downside of publishing a py3-none-any wheel is that platforms that *don't* require the use of `--use-binary=:all:`, and don't have an official binary in PyPI, will no longer get a "compiled at time of installation" binary optimised install. However, those users would be able to compile and install their own binary wheels. This would only affect users on niche platforms (old Linuxes not covered by a published manylinux platform, other Unixes, etc).

freakboy3742:
Fixes #1912.

* Adds support for generating py3-none-any wheels. If the user has `COVERAGE_DISABLE_EXTENSION` defined in their environment, no attempt will be made to compile a binary wheel, even if the platform would otherwise support building the binary extension
* Adds a py3-none-any artefact to the build products of a normal build. A wheel was already being compiled as part of the sdist step; this ensures that (a) the wheel is a py3-none-any wheel, and (b) it's included in the release artefacts.

### DIFF
--- a/.github/workflows/kit.yml
+++ b/.github/workflows/kit.yml
@@ -162,7 +162,7 @@ jobs:
         run: |
           python -m pip install -r requirements/kit.pip
 
-      - name: "Build wheels"
+      - name: "Build binary wheels"
         env:
           CIBW_BUILD: ${{ matrix.py }}*-*
           CIBW_ARCHS: ${{ matrix.arch }}
@@ -173,7 +173,7 @@ jobs:
         run: |
           python -m cibuildwheel --output-dir wheelhouse
 
-      - name: "List wheels"
+      - name: "List binary wheels"
         run: |
           ls -al wheelhouse/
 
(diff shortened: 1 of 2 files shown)


freakboy3742:
Just realised that if I rename the branch, I can generate a test of the kit workflow; closing and re-opening.

PR #1914, opened 2025-01-10
Add py3-none-any wheels to the distribution artefacts

### LINKED ISSUE #1912: Add py3-none-any wheel to distribution artefacts

freakboy3742:
Fixes #1912.

* Adds support for generating py3-none-any wheels. If the user has `COVERAGE_DISABLE_EXTENSION` defined in their environment, no attempt will be made to compile a binary wheel, even if the platform would otherwise support building the binary extension
* Adds a py3-none-any artefact to the build products of a normal build. A wheel was already being compiled as part of the sdist step; this ensures that (a) the wheel is a py3-none-any wheel, and (b) it's included in the release artefacts.

### DIFF
--- a/.github/workflows/kit.yml
+++ b/.github/workflows/kit.yml
@@ -162,7 +162,7 @@ jobs:
         run: |
           python -m pip install -r requirements/kit.pip
 
-      - name: "Build wheels"
+      - name: "Build binary wheels"
         env:
           CIBW_BUILD: ${{ matrix.py }}*-*
           CIBW_ARCHS: ${{ matrix.arch }}
@@ -173,7 +173,7 @@ jobs:
         run: |
           python -m cibuildwheel --output-dir wheelhouse
 
-      - name: "List wheels"
+      - name: "List binary wheels"
         run: |
           ls -al wheelhouse/
 
@@ -181,15 +181,15 @@ jobs:
         run: |
           python -m twine check wheelhouse/*
 
-      - name: "Upload wheels"
+      - name: "Upload binary wheels"
         uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b # v4.5.0
         with:
           name: dist-${{ env.MATRIX_ID }}
           path: wheelhouse/*.whl
           retention-days: 7
 
-  sdist:
-    name: "Source distribution"
+  non-binary:
+    name: "Non-binary artifacts"
     runs-on: ubuntu-latest
     steps:
       - name: "Check out the repo"
@@ -208,23 +208,25 @@ jobs:
         run: |
           python -m pip install -r requirements/kit.pip
 
-      - name: "Build sdist"
+      - name: "Build non-binary artifacts"
+        env:
+          COVERAGE_DISABLE_EXTENSION: 1
         run: |
           python -m build
 
-      - name: "List sdist"
+      - name: "List non-binary artifacts"
         run: |
           ls -al dist/
 
-      - name: "Check sdist"
+      - name: "Check non-binary artifacts"
         run: |
           python -m twine check dist/*
 
-      - name: "Upload sdist"
+      - name: "Upload non-binary artifacts"
         uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b # v4.5.0
         with:
-          name: dist-sdist
-          path: dist/*.tar.gz
+          name: dist-non-binary
+          path: dist/*
           retention-days: 7
 
   pypy:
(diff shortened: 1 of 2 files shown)


freakboy3742:
The kit action was triggered in my own repo (where it was a `push` trigger, rather than a `pull_request`); [This is the run producing a py3-none-any wheel](https://github.com/freakboy3742/coveragepy/actions/runs/12701753339).

nedbat:
Thanks!

But in your action I see:
```
> List non-binary artifacts
-rw-r--r--  1 runner docker 200525 Jan 10 01:49 coverage-7.6.11a0.dev1-py3-none-any.whl
-rw-r--r--  1 runner docker 804612 Jan 10 01:49 coverage-7.6.11a0.dev1.tar.gz
```
and in [my last run of the action](https://github.com/nedbat/coveragepy/actions/runs/12506190364/job/34890738726) I see:
```
> List sdist
-rw-r--r--  1 runner docker 235413 Dec 26 16:45 coverage-7.6.10-cp39-cp39-linux_x86_64.whl
-rw-r--r--  1 runner docker 803868 Dec 26 16:45 coverage-7.6.10.tar.gz
```
Was my .whl there useless?  Is it ok that it has a different name?

freakboy3742:
The old `sdist` build rule ran `python -m build`, which will produce both a wheel (native to the local platform) and an sdist by default - that's what produced the `cp39-cp39-linux_x86_64` wheel. However, the action then [excluded all but the .tar.gz file from the uploaded artefacts](https://github.com/nedbat/coveragepy/pull/1914/files#diff-d7de28e615f067afef28ae6315bd62cbdfd0b4683b044d4d27b979cbe194ae16L227). 

This new config forces the "native" wheel to be py3-none-any, and *includes* it in the uploaded artefacts.

The old configuration could have been optimised slightly by invoking `python -m build -s` - that would have *only* produced the sdist. However, it's a moot point now, as we want the wheel as well.

nedbat:
Sorry for all the questions.  You say "The old sdist build rule ran `python -m build`", but that is still the command being run.  The new config seems to be the environment variable disabling the compilation of the extension.  That's enough to change the name of the result?

freakboy3742:
Invoking `python -m build` produces a wheel and an sdist. The command is running on Linux; so in the old configuration, a binary wheel is produced, because the extension module can be compiled. 

In the new configuration, the compilation of the binary extension is disabled with the environment variable, but a wheel is still being produced; so the wheel identifies as py3-none-any.

nedbat:
This is now released as part of [coverage 7.6.11](https://pypi.org/project/coverage/7.6.11).

PR #1919, opened 2025-01-20
Provide an API for passing plugins to the Coverage constructor

alex:


alex:
Hmm, what's the correct idiom for documenting `TCoverageInit`? I don't see the `types.py` types documented.

nedbat:
I'll have to figure out what to do about the docs.  It might be that we type it as `Iterable[Callable]` or something vague like that, with details in the English.

You made the passed list of plugins override the ones found in the config.  Is there a reason for that? I would have thought it would be additive.

alex:
Sure, I can switch to bare `Iterable[Callable]`, will do that tomorrow.

As for why override, I went with the first sentence in the `Coverage` docstring: "Many of these arguments duplicate and override values that can be provided in a configuration file."

If you have a preference for them being additive instead, happy to switch, I don't feel strongly.

alex:
@nedbat would you prefer I switch to adding, instead of replacing?

nedbat:
You make a good case: "read Ned his own docs to show that I was right" :D

alex:
:D Thanks for the review and merge!

coverage/plugin_support.py
```diff
@@ -51,19 +53,23 @@ def load_plugins(
         plugins = cls()
         plugins.debug = debug
 
-        for module in modules:
-            plugins.current_module = module
-            __import__(module)
-            mod = sys.modules[module]
+        if plugin_override is not None:
+            for override in plugin_override:
+                override(plugins, {})
```
nedbat, line 58: Of course, after merging, thinking about documenting this, I have a hard time explaining why config options are passed to this function but they are always empty.  I'm going to change the signature of the plugin function to leave it out.

alex: Ok! I went for consistency with the signature of plugin inits loaded from the config, but that did produce this slightly silly result.

nedbat:
This is now released as part of [coverage 7.7.0](https://pypi.org/project/coverage/7.7.0).

alex:
❤️ thanks!

PR #1932, opened 2025-03-04
Improve Warning Suppression Performance by Using `set`

allrob23:
## Description

This PR optimizes the way warning suppression is handled by changing `_no_warn_slugs` from a `list` to a `set`. This improves performance when checking if a warning should be suppressed, reducing lookup complexity from O(n) to O(1) on average. Additionally, using a `set` ensures uniqueness, preventing duplicate warning slugs from being added.

## Changes

- Replaced `list[str]` with `set[str]` for `_no_warn_slugs`.
- Updated initialization to use `set(self.config.disable_warnings)`.
- Replaced `.append(slug)` with `.add(slug)` to maintain set behavior.

## Why?

- Improves lookup efficiency.
- Ensures uniqueness of suppressed warning slugs.
- Reduces potential performance overhead in cases with many warnings.

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -258,7 +258,7 @@ class Coverage(TConfigurable):
         self._warn_no_data = True
         self._warn_unimported_source = True
         self._warn_preimported_source = check_preimported
-        self._no_warn_slugs: list[str] = []
+        self._no_warn_slugs: set[str] = []
         self._messages = messages
 
         # A record of all the warnings that have been issued.
@@ -438,7 +438,7 @@ class Coverage(TConfigurable):
 
         """
         if not self._no_warn_slugs:
-            self._no_warn_slugs = list(self.config.disable_warnings)
+            self._no_warn_slugs = set(self.config.disable_warnings)
 
         if slug in self._no_warn_slugs:
             # Don't issue the warning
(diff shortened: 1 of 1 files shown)


coverage/control.py
```diff
@@ -258,7 +258,7 @@ def __init__(                       # pylint: disable=too-many-arguments
         self._warn_no_data = True
         self._warn_unimported_source = True
         self._warn_preimported_source = check_preimported
-        self._no_warn_slugs: list[str] = []
+        self._no_warn_slugs: set[str] = []
```
devdanzin: The initialization is still done with an empty list, should use an empty set instead.

allrob23: sure, updated now, thanks for pointing that out

nedbat:
Thanks! I'm not sure we can measure the speed increase, but it is better philosophically.

nedbat:
@allrob23 How should I credit you in the contributors file?

allrob23:
This PR is part of an ongoing research project.  So may I respond to this question later, when and if our paper is published?

nedbat:
Now I'm curious about the topic of the paper!

Stephen0512:
Hi! Just wanted to follow up and mention that the paper related to this work has now been published: "PyMOP: A Runtime Verification Tool for Python", where we developed a runtime verification tool for Python to help developers check program properties during code execution.

Thanks again for the discussion and feedback on this PR earlier. It was really helpful for the project.

PR #1936, opened 2025-03-12
Add trailing newline to markdown report

rgryta:
Without trailing newline in a markdown report, the first additional information that comes from cmdline handler will end up in an additional row below the TOTALs row.

E.g. Coverage failure: https://github.com/nedbat/coveragepy/blob/master/coverage/cmdline.py#L761

### DIFF
--- a/coverage/report.py
+++ b/coverage/report.py
@@ -108,6 +108,8 @@ class SummaryReporter:
         for end_line in end_lines:
             self.write(end_line)
 
+        self.write("")
+
     def _report_markdown(
         self,
         header: list[str],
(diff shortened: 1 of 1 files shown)


nedbat:
Will you be fixing the failing tests?

rgryta:
> Will you be fixing the failing tests?

Hi, yes. Haven't had time to work on them this weekend. I'll change the last_line assertions to second_last_line.
Unless they'd be better implemented in some other way? Like adding a deletion of all last lines that are empty before performing TOTALS line assertions?

I can add those changes in any way tbh.

nedbat:
I think I'd change the `last_line_squeezed` function to return the last non-blank line.  Line count assertions can be incremented.

PR #1952, opened 2025-04-15
Support `.coveragerc.toml` for configuration

### LINKED ISSUE #1643: [FR] Tool-specific toml config

I maintain enough projects to have a need to synchronize the configs occasionally. Synchronizing configs where sections tend to get shuffled is rather tricky on scale, which is one of the reasons for keeping configs in tool-specific files (the other reason, for me, is that it also makes it obvious where to look for a config right away). Some projects also make use of git skeletons that employ `git merge` to synchronize configs that are primarily maintained in an upstream "skeleton" repository to downstreams that may occasionally diverge over time — in such case, having configs separate also helps with reducing merge conflicts a lot.

`.coveragerc` already solves this problem, but since the ecosystem seems to be moving away from plain INI files (because of some shortcomings) towards using TOML, it seems natural for coverage.py to also have a TOML version of such a config.

Why not `pyproject.toml`? It's suitable for people who don't have to maintain similar configs for different tools across many projects, which is not my case, hence the feature request...

I think there should be a `.coveragerc.toml` where people could put their configs that would take precedence over all other files (provided that either of `tomli` or `tomllib` is importable).
I expect that the structure (section names etc.) would be exactly the same as for `pyproject.toml` to allow for easy migration between the two (plus, it's easier to maintain as the parsing code could essentially be the same).

Keeping the config in `.coveragerc`.

**Additional context**

One example of using the Git skeleton approach is https://blog.jaraco.com/skeleton/.
Among other tools following this approach are Towncrier (supports `towncrier.toml` and `pyproject.toml` with the section name [`[tool.towncrier]`](https://towncrier.readthedocs.io/en/stable/configuration.html#tool-towncrier)) and pip-tools (supports `.pip-tools.toml` and `pyproject.toml` with the section name [`[tool.pip-tools]`](https://pip-tools--1863.org.readthedocs.build/en/1863/#configuration); release pending).

OlenaYefymenko:
This patch provides a new configuration — `.coveragerc.toml `. Considering that many projects have switched to toml configurations, this change offers a more flexible approach to manage coverage settings.

Resolves #1643

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,7 @@ Unreleased
 - New feature: coverage.py now supports :file:`.coveragerc.toml` configuration
   files. These files use TOML syntax and take priority over
   :file:`pyproject.toml` but lower priority than :file:`.coveragerc` files.
+  Closes `issue 1643`_ thanks to `Olena Yefymenko <pull 1952_>`_.
 
 - The .pth file we install was done incorrectly and didn't work when using
   the source wheel (`py3-none-any`).  This is now fixed. Thanks, `Henry
@@ -36,6 +37,8 @@ Unreleased
   3.10). The second and third of these are not needed and will eventually be
   removed.
 
+.. _issue 1643: https://github.com/coveragepy/coveragepy/issues/1643
+.. _pull 1952: https://github.com/coveragepy/coveragepy/pull/1952
 .. _pull 2100: https://github.com/coveragepy/coveragepy/pull/2100
 
 .. start-releases
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -181,9 +181,9 @@ Version 7.10.7 — 2025-09-21
   LCOV reports could take far too long due to some quadratic behavior when
   creating the function and class index pages.  This is now fixed, closing
   `issue 2048`_.  Thanks to Daniel Diniz for help diagnosing the problem.
-- New feature: coverage.py now supports `.coveragerc.toml` configuration files.
+- New feature: coverage.py now supports :file:`.coveragerc.toml` configuration files.
   These files use TOML syntax and take priority over `pyproject.toml` but lower
-  priority than `.coveragerc` files.
+  priority than :file:`.coveragerc` files.
 
 - Most warnings and a few errors now have links to a page in the docs
   explaining the specific message.  Closes `issue 1921`_.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -182,8 +182,8 @@ Version 7.10.7 — 2025-09-21
   creating the function and class index pages.  This is now fixed, closing
   `issue 2048`_.  Thanks to Daniel Diniz for help diagnosing the problem.
 - New feature: coverage.py now supports :file:`.coveragerc.toml` configuration
-  files. These files use TOML syntax and take priority over :file:`pyproject.toml`
-  but lower priority than :file:`.coveragerc` files.
+  files. These files use TOML syntax and take priority over
+  :file:`pyproject.toml` but lower priority than :file:`.coveragerc` files.
 
 - Most warnings and a few errors now have links to a page in the docs
   explaining the specific message.  Closes `issue 1921`_.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -181,6 +181,9 @@ Version 7.10.7 — 2025-09-21
   LCOV reports could take far too long due to some quadratic behavior when
   creating the function and class index pages.  This is now fixed, closing
   `issue 2048`_.  Thanks to Daniel Diniz for help diagnosing the problem.
+- New feature: coverage.py now supports `.coveragerc.toml` configuration files.
+  These files use TOML syntax and take priority over `pyproject.toml` but lower
+  priority than `.coveragerc` files.
 
 - Most warnings and a few errors now have links to a page in the docs
   explaining the specific message.  Closes `issue 1921`_.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,10 @@ upgrading your version of coverage.py.
 Unreleased
 ----------
 
+- New feature: coverage.py now supports :file:`.coveragerc.toml` configuration
+  files. These files use TOML syntax and take priority over
+  :file:`pyproject.toml` but lower priority than :file:`.coveragerc` files.
+
 - The .pth file we install was done incorrectly and didn't work when using
   the source wheel (`py3-none-any`).  This is now fixed. Thanks, `Henry
   Schreiner <pull 2100_>`_.
@@ -181,9 +185,6 @@ Version 7.10.7 — 2025-09-21
   LCOV reports could take far too long due to some quadratic behavior when
   creating the function and class index pages.  This is now fixed, closing
   `issue 2048`_.  Thanks to Daniel Diniz for help diagnosing the problem.
-- New feature: coverage.py now supports :file:`.coveragerc.toml` configuration
-  files. These files use TOML syntax and take priority over
-  :file:`pyproject.toml` but lower priority than :file:`.coveragerc` files.
 
 - Most warnings and a few errors now have links to a page in the docs
   explaining the specific message.  Closes `issue 1921`_.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -181,9 +181,9 @@ Version 7.10.7 — 2025-09-21
   LCOV reports could take far too long due to some quadratic behavior when
   creating the function and class index pages.  This is now fixed, closing
   `issue 2048`_.  Thanks to Daniel Diniz for help diagnosing the problem.
-- New feature: coverage.py now supports :file:`.coveragerc.toml` configuration files.
-  These files use TOML syntax and take priority over :file:`pyproject.toml` but lower
-  priority than :file:`.coveragerc` files.
+- New feature: coverage.py now supports :file:`.coveragerc.toml` configuration
+  files. These files use TOML syntax and take priority over :file:`pyproject.toml`
+  but lower priority than :file:`.coveragerc` files.
 
 - Most warnings and a few errors now have links to a page in the docs
   explaining the specific message.  Closes `issue 1921`_.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -182,7 +182,7 @@ Version 7.10.7 — 2025-09-21
   creating the function and class index pages.  This is now fixed, closing
   `issue 2048`_.  Thanks to Daniel Diniz for help diagnosing the problem.
 - New feature: coverage.py now supports :file:`.coveragerc.toml` configuration files.
-  These files use TOML syntax and take priority over `pyproject.toml` but lower
+  These files use TOML syntax and take priority over :file:`pyproject.toml` but lower
   priority than :file:`.coveragerc` files.
 
 - Most warnings and a few errors now have links to a page in the docs
--- a/coverage/config.py
+++ b/coverage/config.py
@@ -658,6 +658,7 @@ def config_files_to_try(config_file: bool | str) -> list[tuple[str, bool, bool]]
     assert isinstance(config_file, str)
     files_to_try = [
         (config_file, True, specified_file),
+        (".coveragerc.toml", True, False),
         ("setup.cfg", False, False),
         ("tox.ini", False, False),
         ("pyproject.toml", False, False),
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -39,8 +39,9 @@ environment variable.
 
 If ``.coveragerc`` doesn't exist and another file hasn't been specified, then
 coverage.py will look for settings in other common configuration files, in this
-order: :file:`.coveragerc.toml`, :file:`setup.cfg`, :file:`tox.ini`, or :file:`pyproject.toml`.  The first file found with
-coverage.py settings will be used and other files won't be consulted.
+order: :file:`.coveragerc.toml`, :file:`setup.cfg`, :file:`tox.ini`, or
+:file:`pyproject.toml`.  The first file found with coverage.py settings will be
+used and other files won't be consulted.
 
 Coverage.py will read from "pyproject.toml" if TOML support is available,
 either because you are running on Python 3.11 or later, or because you
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -39,7 +39,7 @@ environment variable.
 
 If ``.coveragerc`` doesn't exist and another file hasn't been specified, then
 coverage.py will look for settings in other common configuration files, in this
-order: .coveragerc.toml, setup.cfg, tox.ini, or pyproject.toml.  The first file found with
+order: :file:`.coveragerc.toml`, :file:`setup.cfg`, :file:`tox.ini`, or :file:`pyproject.toml`.  The first file found with
 coverage.py settings will be used and other files won't be consulted.
 
 Coverage.py will read from "pyproject.toml" if TOML support is available,
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -257,7 +257,7 @@ Here's a sample configuration file, in each syntax:
         [coverage:html]
         directory = coverage_html_report
 
-.. [[[end]]] (sum: HU1Z62mvRK)
+.. [[[end]]] (sum: FvgoIIyfXR)
 
 
 The specific configuration settings are described below.  Many sections and
@@ -624,7 +624,7 @@ equivalent when combining data from different machines:
             /jenkins/build/*/src
             c:\myproj\src
 
-.. [[[end]]] (sum: oHSl8SGiMT)
+.. [[[end]]] (sum: bV+PjKu+HS)
 
 
 The names of the entries ("source" in this example) are ignored, you may choose
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -39,7 +39,7 @@ environment variable.
 
 If ``.coveragerc`` doesn't exist and another file hasn't been specified, then
 coverage.py will look for settings in other common configuration files, in this
-order: setup.cfg, tox.ini, or pyproject.toml.  The first file found with
+order: .coveragerc.toml, setup.cfg, tox.ini, or pyproject.toml.  The first file found with
 coverage.py settings will be used and other files won't be consulted.
 
 Coverage.py will read from "pyproject.toml" if TOML support is available,
@@ -199,7 +199,7 @@ Here's a sample configuration file, in each syntax:
         directory = coverage_html_report
 
     .. code-tab:: toml
-        :caption: pyproject.toml
+        :caption: .coveragerc.toml or pyproject.toml
 
         [tool.coverage.run]
         branch = true
@@ -606,7 +606,7 @@ equivalent when combining data from different machines:
             c:\myproj\src
 
     .. code-tab:: toml
-        :caption: pyproject.toml
+        :caption: .coveragerc.toml or pyproject.toml
 
         [tool.coverage.paths]
         source = [
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -170,6 +170,36 @@ Here's a sample configuration file, in each syntax:
 
 .. tabs::
 
+    .. code-tab:: toml
+        :caption: pyproject.toml
+
+        [tool.coverage.run]
+        branch = true
+
+        [tool.coverage.report]
+        # Regexes for lines to exclude from consideration
+        exclude_also = [
+            # Don't complain about missing debug-only code:
+            "def __repr__",
+            "if self\\.debug",
+
+            # Don't complain if tests don't hit defensive assertion code:
+            "raise AssertionError",
+            "raise NotImplementedError",
+
+            # Don't complain if non-runnable code isn't run:
+            "if 0:",
+            "if __name__ == .__main__.:",
+
+            # Don't complain about abstract methods, they aren't run:
+            "@(abc\\.)?abstractmethod",
+            ]
+
+        ignore_errors = true
+
+        [tool.coverage.html]
+        directory = "coverage_html_report"
+
     .. code-tab:: ini
         :caption: .coveragerc
 
@@ -200,12 +230,13 @@ Here's a sample configuration file, in each syntax:
         directory = coverage_html_report
 
     .. code-tab:: toml
-        :caption: .coveragerc.toml or pyproject.toml
+        :caption: .coveragerc.toml
 
-        [tool.coverage.run]
+        # You can also use sections like [tool.coverage.run]
+        [run]
         branch = true
 
-        [tool.coverage.report]
+        [report]
         # Regexes for lines to exclude from consideration
         exclude_also = [
             # Don't complain about missing debug-only code:
@@ -226,7 +257,7 @@ Here's a sample configuration file, in each syntax:
 
         ignore_errors = true
 
-        [tool.coverage.html]
+        [html]
         directory = "coverage_html_report"
 
     .. code-tab:: ini
@@ -258,7 +289,7 @@ Here's a sample configuration file, in each syntax:
         [coverage:html]
         directory = coverage_html_report
 
-.. [[[end]]] (sum: FvgoIIyfXR)
+.. [[[end]]] (sum: qYCtIxMe+3)
 
 
 The specific configuration settings are described below.  Many sections and
@@ -597,6 +628,16 @@ equivalent when combining data from different machines:
 
 .. tabs::
 
+    .. code-tab:: toml
+        :caption: pyproject.toml
+
+        [tool.coverage.paths]
+        source = [
+            "src/",
+            "/jenkins/build/*/src",
+            "c:\\myproj\\src",
+            ]
+
     .. code-tab:: ini
         :caption: .coveragerc
 
@@ -607,9 +648,10 @@ equivalent when combining data from different machines:
             c:\myproj\src
 
     .. code-tab:: toml
-        :caption: .coveragerc.toml or pyproject.toml
+        :caption: .coveragerc.toml
 
-        [tool.coverage.paths]
+        # You can also use sections like [tool.coverage.run]
+        [paths]
         source = [
             "src/",
             "/jenkins/build/*/src",
@@ -625,7 +667,7 @@ equivalent when combining data from different machines:
             /jenkins/build/*/src
             c:\myproj\src
 
-.. [[[end]]] (sum: bV+PjKu+HS)
+.. [[[end]]] (sum: Aq4877/XV0)
 
 
 The names of the entries ("source" in this example) are ignored, you may choose
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1035,6 +1035,27 @@ class ConfigFileTest(UsingModulesMixin, CoverageTest):
         with pytest.raises(ConfigError, match="No option 'foo' in section: 'tool.coverage.run'"):
             config.get("run", "foo")
 
+    def test_coveragerc_toml_priority(self) -> None:
+        """Test that .coveragerc.toml has priority over pyproject.toml."""
+        self.make_file(".coveragerc.toml", """\
+            [tool.coverage.run]
+            timid = true
+            data_file = ".toml-data.dat"
+            branch = true
+            """)
+
+        self.make_file("pyproject.toml", """\
+            [tool.coverage.run]
+            timid = false
+            data_file = "pyproject-data.dat"
+            branch = false
+            """)
+        cov = coverage.Coverage()
+
+        assert cov.config.timid is True
+        assert cov.config.data_file == ".toml-data.dat"
+        assert cov.config.branch is True
+
 
 class SerializeConfigTest(CoverageTest):
     """Tests of serializing the configuration for subprocesses."""
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1017,7 +1017,7 @@ class ConfigFileTest(UsingModulesMixin, CoverageTest):
             assert cov.config.data_file == ".coverage"
 
     @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
-    def test_exceptions_from_missing_toml_things(self, filename) -> None:
+    def test_exceptions_from_missing_toml_things(self, filename: str) -> None:
         self.make_file(
             filename,
             """\
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1037,19 +1037,25 @@ class ConfigFileTest(UsingModulesMixin, CoverageTest):
 
     def test_coveragerc_toml_priority(self) -> None:
         """Test that .coveragerc.toml has priority over pyproject.toml."""
-        self.make_file(".coveragerc.toml", """\
+        self.make_file(
+            ".coveragerc.toml",
+            """\
             [tool.coverage.run]
             timid = true
             data_file = ".toml-data.dat"
             branch = true
-            """)
+            """,
+        )
 
-        self.make_file("pyproject.toml", """\
+        self.make_file(
+            "pyproject.toml",
+            """\
             [tool.coverage.run]
             timid = false
             data_file = "pyproject-data.dat"
             branch = false
-            """)
+            """,
+        )
         cov = coverage.Coverage()
 
         assert cov.config.timid is True
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -75,7 +75,7 @@ class ConfigTest(CoverageTest):
         assert cov.config.data_file == "delete.me"
 
     @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
-    def test_toml_config_file(self, filename) -> None:
+    def test_toml_config_file(self, filename: str) -> None:
         # A pyproject.toml and .coveragerc.toml file will be read into the configuration.
         self.make_file(
             filename,
@@ -109,7 +109,7 @@ class ConfigTest(CoverageTest):
         assert cov.config.get_plugin_options("plugins.a_plugin") == {"hello": "world"}
 
     @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
-    def test_toml_ints_can_be_floats(self, filename) -> None:
+    def test_toml_ints_can_be_floats(self, filename: str) -> None:
         # Test that our class doesn't reject integers when loading floats
         self.make_file(
             filename,
@@ -301,7 +301,7 @@ class ConfigTest(CoverageTest):
             ('[tool.coverage.report]\nfail_under="s"', "couldn't convert to a float"),
         ],
     )
-    def test_toml_parse_errors(self, filename, bad_config: str, msg: str) -> None:
+    def test_toml_parse_errors(self, filename: str, bad_config: str, msg: str) -> None:
         # Im-parsable values raise ConfigError, with details.
         self.make_file(filename, bad_config)
         with pytest.raises(ConfigError, match=msg):
@@ -333,7 +333,7 @@ class ConfigTest(CoverageTest):
         assert cov.config.exclude_list == ["the_$one", "anotherZZZ", "xZZZy", "xy", "huh${X}what"]
 
     @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
-    def test_environment_vars_in_toml_config(self, filename) -> None:
+    def test_environment_vars_in_toml_config(self, filename: str) -> None:
         # Config files can have $envvars in them.
         self.make_file(
             filename,
@@ -405,7 +405,7 @@ class ConfigTest(CoverageTest):
         self.assert_tilde_results()
 
     @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
-    def test_tilde_in_toml_config(self, filename) -> None:
+    def test_tilde_in_toml_config(self, filename: str) -> None:
         # Config entries that are file paths can be tilde-expanded.
         self.make_file(
             filename,
@@ -546,7 +546,7 @@ class ConfigTest(CoverageTest):
             _ = coverage.Coverage()
 
     @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
-    def test_unknown_option_toml(self, filename) -> None:
+    def test_unknown_option_toml(self, filename: str) -> None:
         self.make_file(
             filename,
             """\
@@ -614,7 +614,7 @@ class ConfigTest(CoverageTest):
             config.get("xyzzy", "foo")
 
     @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
-    def test_exclude_also(self, filename) -> None:
+    def test_exclude_also(self, filename: str) -> None:
         self.make_file(
             filename,
             """\
@@ -966,7 +966,7 @@ class ConfigFileTest(UsingModulesMixin, CoverageTest):
 
     @pytest.mark.skipif(env.PYVERSION >= (3, 11), reason="Python 3.11 has toml in stdlib")
     @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
-    def test_no_toml_installed_pyproject_toml(self, filename) -> None:
+    def test_no_toml_installed_pyproject_toml(self, filename: str) -> None:
         # Can't have coverage config in pyproject.toml/.coveragerc.toml without toml installed.
         self.make_file(
             filename,
@@ -983,7 +983,7 @@ class ConfigFileTest(UsingModulesMixin, CoverageTest):
 
     @pytest.mark.skipif(env.PYVERSION >= (3, 11), reason="Python 3.11 has toml in stdlib")
     @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
-    def test_no_toml_installed_pyproject_toml_shorter_syntax(self, filename) -> None:
+    def test_no_toml_installed_pyproject_toml_shorter_syntax(self, filename: str) -> None:
         # Can't have coverage config in pyproject.toml/.coveragerc.toml without toml installed.
         self.make_file(
             filename,
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -74,10 +74,11 @@ class ConfigTest(CoverageTest):
         assert not cov.config.branch
         assert cov.config.data_file == "delete.me"
 
-    def test_toml_config_file(self) -> None:
-        # A pyproject.toml file will be read into the configuration.
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_toml_config_file(self, filename) -> None:
+        # A pyproject.toml and .coveragerc.toml file will be read into the configuration.
         self.make_file(
-            "pyproject.toml",
+            filename,
             """\
             # This is just a bogus toml file for testing.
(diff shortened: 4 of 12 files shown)


coverage/config.py
```diff
@@ -571,6 +571,7 @@ def config_files_to_try(config_file: bool | str) -> list[tuple[str, bool, bool]]
     assert isinstance(config_file, str)
     files_to_try = [
         (config_file, True, specified_file),
+        (".coveragerc.toml", True, False),
         ("setup.cfg", False, False),
```
OlenaYefymenko, line 662: Is the priority of the new config properly determined in the `config_files_to_try` function?

tests/test_config.py
```diff
@@ -64,9 +64,10 @@ def test_named_config_file(self, file_class: FilePathType) -> None:
         assert not cov.config.branch
         assert cov.config.data_file == "delete.me"
 
-    def test_toml_config_file(self) -> None:
-        # A pyproject.toml file will be read into the configuration.
-        self.make_file("pyproject.toml", """\
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_toml_config_file(self, filename) -> None:
+        # A pyproject.toml and coveragerc.toml will be read into the configuration.
+        self.make_file(filename, """\
```
OlenaYefymenko: Is the parameterization implemented appropriately, similar to `pyproject.toml`?

tests/test_config.py
```diff
@@ -814,17 +817,39 @@ def test_no_toml_installed_pyproject_no_coverage(self) -> None:
...
+        config.read(filename)
         with pytest.raises(ConfigError, match="No section: 'xyzzy'"):
             config.options("xyzzy")
         with pytest.raises(ConfigError, match="No section: 'xyzzy'"):
             config.get("xyzzy", "foo")
         with pytest.raises(ConfigError, match="No option 'foo' in section: 'tool.coverage.run'"):
             config.get("run", "foo")
+            
+    def test_coveragerc_toml_priority(self) -> None:
+        """Test that .coveragerc.toml has priority over pyproject.toml."""
```
OlenaYefymenko, line 1038: Is testing priority only over `pyproject.tom`l in the `test_coveragerc_toml_priority` test sufficient?

OlenaYefymenko:
@nedbat,
I've opened PR with the improvements, but I have a few doubts about my implementation. I've added comments in the code. I'd really appreciate your feedback on whether this approach works well.
Thank you!

coverage/config.py
nedbat: Looks good.

tests/test_config.py
nedbat: yes.

tests/test_config.py
nedbat: yes.

tests/test_config.py
```diff
@@ -775,35 +774,39 @@ def test_no_toml_installed_explicit_toml(self) -> None:
...
         # Can't have coverage config in pyproject.toml without toml installed.
-        self.make_file("pyproject.toml", """\
+        self.make_file(filename, """\
             # A toml file!
             [tool.coverage]
             run.parallel = true
             """)
         with mock.patch.object(coverage.tomlconfig, "has_tomllib", False):
-            msg = "Can't read 'pyproject.toml' without TOML support"
+            msg = "Can't read '{filename}' without TOML support"
```
nedbat: Looks like tests are failing because this needs to be an f-string.

nedbat (changes requested):
One change needed.

nedbat:
Thanks for doing this.  There are some other changes needed.  Let me know which you want to do, and which I should do:
- Update the documentation
- Add an entry to CHANGES.rst
- Add your name to CONTRIBUTORS.txt

tests/test_config.py
OlenaYefymenko: Thank you. I will fix this error in the next commits. I run `tox` with version 3.12 locally, and I didn't see this error.

OlenaYefymenko:
> Thanks for doing this. There are some other changes needed. Let me know which you want to do, and which I should do:
> 
> * Update the documentation
> * Add an entry to CHANGES.rst
> * Add your name to CONTRIBUTORS.txt

Yes, I will add the changes

tests/test_config.py
```diff
@@ -775,35 +774,39 @@ def test_no_toml_installed_explicit_toml(self) -> None:
...
+    def test_no_toml_installed_pyproject_toml(self, filename) -> None:
+        # Can't have coverage config in pyproject.toml and .coveragerc.toml without toml installed.
+        self.make_file(filename, """\
             # A toml file!
             [tool.coverage.run]
             xyzzy = 17
             """)
         with mock.patch.object(coverage.tomlconfig, "has_tomllib", False):
-            msg = "Can't read 'pyproject.toml' without TOML support"
+            msg = "Can't read '{filename}' without TOML support"
```
webknjaz:
```suggestion
            msg = f"Can't read '{filename}' without TOML support"
```

tests/test_config.py
webknjaz:
```suggestion
            msg = f"Can't read '{filename}' without TOML support"
```

Elenore23 (approved):

doc/config.rst
```diff
@@ -39,12 +39,13 @@ environment variable.
 
 If ``.coveragerc`` doesn't exist and another file hasn't been specified, then
 coverage.py will look for settings in other common configuration files, in this
-order: setup.cfg, tox.ini, or pyproject.toml.  The first file found with
+order: .coveragerc.toml, setup.cfg, tox.ini, or pyproject.toml.  The first file found with
```
OlenaYefymenko:
```suggestion
order: :file:`.coveragerc.toml`, setup.cfg, tox.ini, or pyproject.toml.  The first file found with
```

doc/config.rst
```diff
@@ -39,12 +39,13 @@ environment variable.
...
 coverage.py settings will be used and other files won't be consulted.
 
-Coverage.py will read from "pyproject.toml" if TOML support is available,
+Coverage.py will read from ".coveragerc.toml" and "pyproject.toml" if TOML support is available,
```
webknjaz:
```suggestion
Coverage.py will read from ".coveragerc.toml" or "pyproject.toml" if TOML support is available,
```

doc/config.rst
webknjaz: ^

doc/config.rst
```diff
@@ -136,6 +137,7 @@ Here's a sample configuration file, in each syntax:
             [html]
             directory = coverage_html_report
             """,
+
```
webknjaz:
```suggestion
```

tests/test_config.py
```diff
@@ -66,9 +67,10 @@ def test_named_config_file(self, file_class: FilePathType) -> None:
         assert not cov.config.branch
         assert cov.config.data_file == "delete.me"
 
-    def test_toml_config_file(self) -> None:
-        # A pyproject.toml file will be read into the configuration.
-        self.make_file("pyproject.toml", """\
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_toml_config_file(self, filename: str) -> None:
+        # A pyproject.toml and coveragerc.toml will be read into the configuration.
```
webknjaz:
```suggestion
        # A pyproject.toml or coveragerc.toml will be read into the configuration.
```

tests/test_config.py
```diff
@@ -263,11 +268,13 @@ def test_environment_vars_in_config(self) -> None:
         cov = coverage.Coverage()
         assert cov.config.data_file == "hello-world.fooey"
         assert cov.config.branch is True
-        assert cov.config.exclude_list == ["the_$one", "anotherZZZ", "xZZZy", "xy", "huh${X}what"]
+        assert cov.config.exclude_list == [
+            "the_$one", "anotherZZZ", "xZZZy", "xy", "huh${X}what"]
```
webknjaz:
```suggestion
        assert cov.config.exclude_list == ["the_$one", "anotherZZZ", "xZZZy", "xy", "huh${X}what"]
```

tests/test_config.py
```diff
@@ -296,7 +303,8 @@ def test_environment_vars_in_toml_config(self) -> None:
         assert cov.config.branch is True
         assert cov.config.precision == 3
         assert cov.config.data_file == "hello-world.fooey"
-        assert cov.config.exclude_list == ["the_$one", "anotherZZZ", "xZZZy", "xy", "huh${X}what"]
+        assert cov.config.exclude_list == [
+            "the_$one", "anotherZZZ", "xZZZy", "xy", "huh${X}what"]
```
webknjaz:
```suggestion
        assert cov.config.exclude_list == ["the_$one", "anotherZZZ", "xZZZy", "xy", "huh${X}what"]
```

tests/test_config.py
```diff
@@ -384,7 +393,8 @@ def expanduser(s: str) -> str:
         assert cov.config.lcov_output == "/Users/me/lcov/~foo.lcov"
         assert cov.config.xml_output == "/Users/me/somewhere/xml.out"
         assert cov.config.exclude_list == ["~/data.file", "~joe/html_dir"]
-        assert cov.config.paths == {'mapping': ['/Users/me/src', '/Users/joe/source']}
+        assert cov.config.paths == {'mapping': [
+            '/Users/me/src', '/Users/joe/source']}
```
webknjaz:
```suggestion
        assert cov.config.paths == {'mapping': ['/Users/me/src', '/Users/joe/source']}
```

tests/test_config.py
```diff
@@ -440,7 +450,8 @@ def test_tweak_error_checking(self) -> None:
     def test_tweak_plugin_options(self) -> None:
         # Plugin options have a more flexible syntax.
         cov = coverage.Coverage()
-        cov.set_option("run:plugins", ["fooey.plugin", "xyzzy.coverage.plugin"])
+        cov.set_option("run:plugins", [
+                       "fooey.plugin", "xyzzy.coverage.plugin"])
```
webknjaz:
```suggestion
        cov.set_option("run:plugins", ["fooey.plugin", "xyzzy.coverage.plugin"])
```

tests/test_config.py
```diff
@@ -460,12 +471,13 @@ def test_unknown_option(self) -> None:
...
-    def test_unknown_option_toml(self) -> None:
-        self.make_file("pyproject.toml", """\
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_unknown_option_toml(self, filename: str) -> None:
+        self.make_file(filename, """\
             [tool.coverage.run]
             xyzzy = 17
             """)
-        msg = r"Unrecognized option '\[tool.coverage.run\] xyzzy=' in config file pyproject.toml"
+        msg = f"Unrecognized option '\\[tool.coverage.run\\] xyzzy=' in config file {filename}"
```
webknjaz:
```suggestion
        msg = fr"Unrecognized option '\[tool.coverage.run\] xyzzy=' in config file {filename}"
```

tests/test_config.py
```diff
@@ -732,10 +750,12 @@ def check_other_config_if_coveragerc_specified(self, fname: str, contents: str)
...
 
     def test_config_file_settings_in_setupcfg_if_coveragerc_specified(self) -> None:
-        self.check_other_config_if_coveragerc_specified("setup.cfg", self.SETUP_CFG)
+        self.check_other_config_if_coveragerc_specified(
+            "setup.cfg", self.SETUP_CFG)
 
     def test_config_file_settings_in_tox_if_coveragerc_specified(self) -> None:
-        self.check_other_config_if_coveragerc_specified("tox.ini", self.TOX_INI)
+        self.check_other_config_if_coveragerc_specified(
+            "tox.ini", self.TOX_INI)
```
webknjaz:
```suggestion
        self.check_other_config_if_coveragerc_specified("tox.ini", self.TOX_INI)
```

tests/test_config.py
```diff
@@ -732,10 +750,12 @@ def check_other_config_if_coveragerc_specified(self, fname: str, contents: str)
         self.assert_config_settings_are_correct(cov)
 
     def test_config_file_settings_in_setupcfg_if_coveragerc_specified(self) -> None:
-        self.check_other_config_if_coveragerc_specified("setup.cfg", self.SETUP_CFG)
+        self.check_other_config_if_coveragerc_specified(
+            "setup.cfg", self.SETUP_CFG)
```
webknjaz:
```suggestion
        self.check_other_config_if_coveragerc_specified("setup.cfg", self.SETUP_CFG)
```

tests/test_config.py
```diff
@@ -719,7 +736,8 @@ def check_config_file_settings_in_other_file(self, fname: str, contents: str) ->
         self.assert_config_settings_are_correct(cov)
 
     def test_config_file_settings_in_setupcfg(self) -> None:
-        self.check_config_file_settings_in_other_file("setup.cfg", self.SETUP_CFG)
+        self.check_config_file_settings_in_other_file(
+            "setup.cfg", self.SETUP_CFG)
```
webknjaz:
```suggestion
        self.check_config_file_settings_in_other_file("setup.cfg", self.SETUP_CFG)
```

tests/test_config.py
```diff
@@ -33,7 +33,8 @@ def test_default_config(self) -> None:
 
     def test_arguments(self) -> None:
         # Arguments to the constructor are applied to the configuration.
-        cov = coverage.Coverage(timid=True, data_file="fooey.dat", concurrency="multiprocessing")
+        cov = coverage.Coverage(
+            timid=True, data_file="fooey.dat", concurrency="multiprocessing")
```
webknjaz:
```suggestion
        cov = coverage.Coverage(timid=True, data_file="fooey.dat", concurrency="multiprocessing")
```

tests/test_config.py
```diff
@@ -529,7 +543,8 @@ def test_partial_also(self) -> None:
             """)
         cov = coverage.Coverage()
 
-        expected = coverage.config.DEFAULT_PARTIAL + ["foobar", "raise .*Error"]
+        expected = coverage.config.DEFAULT_PARTIAL + \
+            ["foobar", "raise .*Error"]
```
webknjaz:
```suggestion
        expected = coverage.config.DEFAULT_PARTIAL + ["foobar", "raise .*Error"]
```

webknjaz (commented):
Plz revert all the formatting changes. I marked some, find the rest.

CHANGES.rst
```diff
@@ -202,6 +202,16 @@ Version 7.10.0 — 2025-07-24
...
 
+- Feature: Added support for ``.coveragerc.toml`` configuration files. This
+  provides a more flexible approach to manage coverage settings as many
+  projects have switched to TOML configurations. Closes `issue 1643`_.
+
+.. _issue 1643: https://github.com/nedbat/coveragepy/issues/1643
+.. _pull 1952: https://github.com/nedbat/coveragepy/pull/1952
+
+
+.. start-releases
```
webknjaz: Invalid conflict resolution?

doc/config.rst
```diff
@@ -136,6 +139,7 @@ Here's a sample configuration file, in each syntax:
...
```
webknjaz: Stray edit?

tests/test_config.py
```diff
@@ -74,11 +74,10 @@ def test_named_config_file(self, file_class: FilePathType) -> None:
...
         assert cov.config.data_file == "delete.me"
 
-    def test_toml_config_file(self) -> None:
-        # A pyproject.toml file will be read into the configuration.
-        self.make_file(
-            "pyproject.toml",
-            """\
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_toml_config_file(self, filename: str) -> None:
+        # A pyproject.toml and coveragerc.toml will be read into the configuration.
```
webknjaz:
```suggestion
        # A pyproject.toml and .coveragerc.toml will be read into the configuration.
```

tests/test_config.py
```diff
@@ -262,45 +260,33 @@ def test_parse_errors(self, bad_config: str, msg: str) -> None:
...
-            ),
-            ('[tool.coverage.run]\nconcurrency="foo"', "not a list"),
-            ("[tool.coverage.report]\nprecision=1.23", "not an integer"),
-            ('[tool.coverage.report]\nfail_under="s"', "couldn't convert to a float"),
-        ],
-    )
-    def test_toml_parse_errors(self, bad_config: str, msg: str) -> None:
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    @pytest.mark.parametrize("bad_config, msg", [
+        ("[tool.coverage.run]\ntimid = \"maybe?\"\n", r"maybe[?]"),
```
webknjaz: A lot of reformatting?

tests/test_config.py
```diff
@@ -458,7 +443,7 @@ def expanduser(s: str) -> str:
         assert cov.config.lcov_output == "/Users/me/lcov/~foo.lcov"
         assert cov.config.xml_output == "/Users/me/somewhere/xml.out"
         assert cov.config.exclude_list == ["~/data.file", "~joe/html_dir"]
-        assert cov.config.paths == {"mapping": ["/Users/me/src", "/Users/joe/source"]}
+        assert cov.config.paths == {'mapping': ['/Users/me/src', '/Users/joe/source']}
```
webknjaz: ?

tests/test_config.py
```diff
@@ -976,9 +961,7 @@ def test_no_toml_installed_pyproject_toml(self) -> None:
     @pytest.mark.skipif(env.PYVERSION >= (3, 11), reason="Python 3.11 has toml in stdlib")
     def test_no_toml_installed_pyproject_toml_shorter_syntax(self) -> None:
         # Can't have coverage config in pyproject.toml without toml installed.
-        self.make_file(
-            "pyproject.toml",
-            """\
+        self.make_file("pyproject.toml", """\
```
webknjaz: ?

tests/test_config.py
```diff
@@ -1007,55 +988,62 @@ def test_no_toml_installed_pyproject_no_coverage(self) -> None:
...
+    def test_coveragerc_toml_priority(self) -> None:
+        """Test that .coveragerc.toml has priority over pyproject.toml."""
+        self.make_file(".coveragerc.toml", """\
+            [tool.coverage.run]
+            timid = true
+            data_file = ".toml-data.dat"
+            branch = true
+            """)
 
-class SerializeConfigTest(CoverageTest):
```
webknjaz, line 1028: ?

OlenaYefymenko:
I'm in progress -  some tests have failed, I'm going to fix them

doc/config.rst
```diff
@@ -39,7 +39,7 @@ environment variable.
 
 If ``.coveragerc`` doesn't exist and another file hasn't been specified, then
 coverage.py will look for settings in other common configuration files, in this
-order: setup.cfg, tox.ini, or pyproject.toml.  The first file found with
+order: .coveragerc.toml, setup.cfg, tox.ini, or pyproject.toml.  The first file found with
```
webknjaz: Is this actual priority order, though?

tests/test_config.py
```diff
@@ -74,10 +74,11 @@ def test_named_config_file(self, file_class: FilePathType) -> None:
         assert not cov.config.branch
         assert cov.config.data_file == "delete.me"
 
-    def test_toml_config_file(self) -> None:
-        # A pyproject.toml file will be read into the configuration.
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_toml_config_file(self, filename) -> None:
+        # A pyproject.toml and .coveragerc.toml file will be read into the configuration.
```
webknjaz: Well, not both, just one of them, right?
```suggestion
        # A pyproject.toml or .coveragerc.toml file will be read into the configuration.
```

tests/test_config.py
webknjaz: @OlenaYefymenko apply this suggestion to reduce the need to escape things too much.

CHANGES.rst
```diff
@@ -45,6 +45,9 @@ Version 7.10.7 — 2025-09-21
   LCOV reports could take far too long due to some quadratic behavior when
   creating the function and class index pages.  This is now fixed, closing
   `issue 2048`_.  Thanks to Daniel Diniz for help diagnosing the problem.
+- New feature: coverage.py now supports `.coveragerc.toml` configuration files.
```
webknjaz:
```suggestion
- New feature: coverage.py now supports :file:`.coveragerc.toml` configuration files.
```

CHANGES.rst
```diff
@@ -45,6 +45,9 @@ Version 7.10.7 — 2025-09-21
...
+  These files use TOML syntax and take priority over `pyproject.toml` but lower
+  priority than `.coveragerc` files.
```
webknjaz: this is not markdown
```suggestion
  priority than :file:`.coveragerc` files.
```

tests/test_config.py
```diff
@@ -74,10 +74,11 @@ def test_named_config_file(self, file_class: FilePathType) -> None:
         assert not cov.config.branch
         assert cov.config.data_file == "delete.me"
 
-    def test_toml_config_file(self) -> None:
-        # A pyproject.toml file will be read into the configuration.
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_toml_config_file(self, filename) -> None:
```
webknjaz:
```suggestion
    def test_toml_config_file(self, filename: str) -> None:
```

tests/test_config.py
```diff
@@ -107,10 +108,11 @@ def test_toml_config_file(self) -> None:
         assert cov.config.fail_under == 90.5
         assert cov.config.get_plugin_options("plugins.a_plugin") == {"hello": "world"}
 
-    def test_toml_ints_can_be_floats(self) -> None:
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_toml_ints_can_be_floats(self, filename) -> None:
```
webknjaz:
```suggestion
    def test_toml_ints_can_be_floats(self, filename: str) -> None:
```

tests/test_config.py
```diff
@@ -298,9 +301,9 @@ def test_parse_errors(self, bad_config: str, msg: str) -> None:
             ('[tool.coverage.report]\nfail_under="s"', "couldn't convert to a float"),
         ],
     )
-    def test_toml_parse_errors(self, bad_config: str, msg: str) -> None:
+    def test_toml_parse_errors(self, filename, bad_config: str, msg: str) -> None:
```
webknjaz:
```suggestion
    def test_toml_parse_errors(self, filename: str, bad_config: str, msg: str) -> None:
```

tests/test_config.py
```diff
@@ -329,10 +332,11 @@ def test_environment_vars_in_config(self) -> None:
         assert cov.config.branch is True
         assert cov.config.exclude_list == ["the_$one", "anotherZZZ", "xZZZy", "xy", "huh${X}what"]
 
-    def test_environment_vars_in_toml_config(self) -> None:
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_environment_vars_in_toml_config(self, filename) -> None:
```
webknjaz:
```suggestion
    def test_environment_vars_in_toml_config(self, filename: str) -> None:
```

tests/test_config.py
```diff
@@ -400,10 +404,11 @@ def test_tilde_in_config(self) -> None:
 
         self.assert_tilde_results()
 
-    def test_tilde_in_toml_config(self) -> None:
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_tilde_in_toml_config(self, filename) -> None:
```
webknjaz:
```suggestion
    def test_tilde_in_toml_config(self, filename: str) -> None:
```

tests/test_config.py
```diff
@@ -540,15 +545,16 @@ def test_unknown_option(self) -> None:
         with pytest.warns(CoverageWarning, match=msg):
             _ = coverage.Coverage()
 
-    def test_unknown_option_toml(self) -> None:
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_unknown_option_toml(self, filename) -> None:
```
webknjaz:
```suggestion
    def test_unknown_option_toml(self, filename: str) -> None:
```

tests/test_config.py
```diff
@@ -607,9 +613,10 @@ def test_exceptions_from_missing_things(self) -> None:
         with pytest.raises(ConfigError, match="No option 'foo' in section: 'xyzzy'"):
             config.get("xyzzy", "foo")
 
-    def test_exclude_also(self) -> None:
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_exclude_also(self, filename) -> None:
```
webknjaz:
```suggestion
    def test_exclude_also(self, filename: str) -> None:
```

tests/test_config.py
```diff
@@ -958,42 +965,45 @@ def test_no_toml_installed_explicit_toml(self) -> None:
                 coverage.Coverage(config_file="cov.toml")
 
     @pytest.mark.skipif(env.PYVERSION >= (3, 11), reason="Python 3.11 has toml in stdlib")
-    def test_no_toml_installed_pyproject_toml(self) -> None:
-        # Can't have coverage config in pyproject.toml without toml installed.
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_no_toml_installed_pyproject_toml(self, filename) -> None:
```
webknjaz:
```suggestion
    def test_no_toml_installed_pyproject_toml(self, filename: str) -> None:
```

tests/test_config.py
```diff
@@ -958,42 +965,45 @@ def test_no_toml_installed_explicit_toml(self) -> None:
...
-            msg = "Can't read 'pyproject.toml' without TOML support"
+            msg = f"Can't read '{filename}' without TOML support"
             with pytest.raises(ConfigError, match=msg):
                 coverage.Coverage()
 
     @pytest.mark.skipif(env.PYVERSION >= (3, 11), reason="Python 3.11 has toml in stdlib")
-    def test_no_toml_installed_pyproject_toml_shorter_syntax(self) -> None:
-        # Can't have coverage config in pyproject.toml without toml installed.
+    @pytest.mark.parametrize("filename", ["pyproject.toml", ".coveragerc.toml"])
+    def test_no_toml_installed_pyproject_toml_shorter_syntax(self, filename) -> None:
```
webknjaz:
```suggestion
    def test_no_toml_installed_pyproject_toml_shorter_syntax(self, filename: str) -> None:
```

doc/config.rst
webknjaz: Never mind, I got confused. Consider this matter resolved.

CHANGES.rst
```diff
@@ -45,6 +45,9 @@ Version 7.10.7 — 2025-09-21
   LCOV reports could take far too long due to some quadratic behavior when
   creating the function and class index pages.  This is now fixed, closing
   `issue 2048`_.  Thanks to Daniel Diniz for help diagnosing the problem.
+- New feature: coverage.py now supports :file:`.coveragerc.toml` configuration files.
+  These files use TOML syntax and take priority over `pyproject.toml` but lower
```
webknjaz:
```suggestion
  These files use TOML syntax and take priority over :file:`pyproject.toml` but lower
```

doc/config.rst
webknjaz: While on it, you could also add the file role to the paths.
```suggestion
order: :file:`.coveragerc.toml`, :file:`setup.cfg`, :file:`tox.ini`, or :file:`pyproject.toml`.  The first file found with
```

CHANGES.rst
```diff
@@ -45,6 +45,9 @@ Version 7.10.7 — 2025-09-21
   LCOV reports could take far too long due to some quadratic behavior when
   creating the function and class index pages.  This is now fixed, closing
   `issue 2048`_.  Thanks to Daniel Diniz for help diagnosing the problem.
+- New feature: coverage.py now supports :file:`.coveragerc.toml` configuration
```
webknjaz: This entry ended up under an already released version. We should move it to line 26.

ofek:
I'm very excited to start using this feature, thanks for the effort here @OlenaYefymenko!!!

nedbat:
@webknjaz @OlenaYefymenko What is left to do with this PR?

webknjaz:
@nedbat just the change log move. Olena was going to mark is as ready just about now.

OlenaYefymenko:
> @webknjaz @OlenaYefymenko What is left to do with this PR?

@nedbat the pull request is ready for review. Please let me know if anything needs to be edited.
Thank you for your help and patience

nedbat:
Thanks for doing this, and thanks for persisting.

With these changes, the `.coverage.toml` file would have entries like:
```toml
[tool.coverage.run]
branch = true
```
but wouldn't we want this?:
```toml
[run]
branch = true
```
I'd like to make things seem natural, and I also like to give people options, so it's hard to strike the right balance.  When I read (for example) pytest's explanation of their [configuration rules](https://docs.pytest.org/en/stable/reference/customize.html#configuration-file-formats) my head starts to spin.

webknjaz:
> but wouldn't we want this?:
> 
> ```toml
> [run]
> branch = true
> ```

I've seen both startegies in the wild and I'm leaning towards not having differences with `pyproject.toml`.

For example, [Towncrier has `[tool.towncrier]` in `towncrier.toml`](https://github.com/twisted/towncrier/pull/162/commits/f6f968860169f5032480f972dd7c4477450c566d#diff-a76468ea66643d8ab224ae7aa3aab66c993d28577c52165a85fa2bb3d83b66ceR54) and in cython-lint, [I asked for the same `[tool.cython-lint]` in `.cython-lint.toml`](https://github.com/MarcoGorelli/cython-lint/issues/84). [I also insisted on this when we were getting a PR in pip-tools](https://github.com/jazzband/pip-tools/pull/1863/commits/db2f30b7914f0c935cc700be050aedf9d80ad6a3) — you'll have `[tool.pip-tools]` in `.pip-tools.toml`. [PyLint does the same](https://github.com/pylint-dev/pylint/pull/9230/files#diff-5157ab5b04a2bb909732c57fbf9a6b5e1f7abe50eb884331638cb1fa3a53a25fR99). Pyright seems to have implemented parsing custom `.toml` files as `pyproject.toml` (https://github.com/microsoft/pyright/pull/7997) but it's unclear if it picks up a `.pyright.toml` as the default automatically. Poetry has a `poetry.toml` for the tool configuration that isn't PEP 621 packaging metadata: https://python-poetry.org/docs/configuration/#local-configuration. For Black, it's a bit confusing but it mentions `.black` files in user-global locations + passing a custom name on the command line:  https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#where-black-looks-for-the-file — they are likely TOML with the same structure too, but I haven't checked.
isort doesn't really make it clear how it parses a custom config file: https://pycqa.github.io/isort/docs/configuration/config_files#custom-config-files.

I did mention this opinion in the original issue, FTR: #1643.

My reasoning for this is that it's easier to migrate/move things between `pyproject.toml` and a `.{tool-name}.toml` for the end-users, UX-wise. And it's more straightforward when there's no differences with custom filenames: what if one passes a `--config=an-arbitrary-file-name-that-is-not-pyproject.toml` — which way we'd want to parse it? If they are the same, the same parser is used and there's no confusion between adding or not adding section prefixes. But if there's differences, this would incur more maintenance burden for both project maintenance and end-users.

> I'd like to make things seem natural, and I also like to give people options, so it's hard to strike the right balance.  When I read (for example) pytest's explanation of their [configuration rules](https://docs.pytest.org/en/stable/reference/customize.html#configuration-file-formats) my head starts to spin.

Right. I missed the PR it when was being implemented.. I'd ask for a prefix if I'd seen it, but there's no going back I guess.

Ruff has the section/prefix removed, OTOH: https://docs.astral.sh/ruff/configuration/#__tabbed_1_2. It does save a few keystrokes but I'd rather keep it uniform.
Hatch also has unprefixed sections: https://hatch.pypa.io/latest/config/environment/overview/#environment-configuration.

ofek:
Please, please do not support the `tool.coverage` prefix in this new file.

1. Not matching the structure of the existing `.coveragerc` dedicated config file violates the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).
2. It's unnatural for design decisions of individual projects to be influenced by the language's packaging configuration. For example, `Cargo.toml` files [support](https://doc.rust-lang.org/cargo/reference/manifest.html#the-metadata-table) the same type of table called `package.metadata` that is ignored and is meant for tools. I've never seen config for projects that happen to be written in Rust, or are for Rust development, use that prefix.
3. We should take note of what developers use the most rather than noting precedent for keeping the prefix. Here are examples of popular projects (some already mentioned) that support dedicated config files without a prefix: [pytest](https://docs.pytest.org/en/stable/reference/customize.html#pytest-toml), [Ruff](https://docs.astral.sh/ruff/configuration/), [uv](https://docs.astral.sh/uv/concepts/configuration-files/), [Hatch](https://hatch.pypa.io/latest/intro/#configuration), [tox](https://tox.wiki/en/4.32.0/config.html#tox-toml), etc. It's worth noting as well that projects receiving more maintenance tend to favor this approach to decouple from historical packaging nuances and provide a better UX.
4. Tools that compute a dependency graph like uv cannot distinguish distribution-affecting `pyproject.toml` changes from others because it's all about the file hash (same as other build systems like Bazel). In this case, commands like `uv run` must rebuild the package whenever changes occur and it becomes advantageous to extract configuration that doesn't affect the build artifacts. This is particularly cumbersome for projects with extension modules: one change to coverage measurement means waiting dozens of seconds or more unnecessarily.

@OlenaYefymenko Would you be open to such a change in this PR?

webknjaz:
That's up to @nedbat to decide, then.

ofek:
Oh, for sure! I was just asking if @OlenaYefymenko would personally be fine with that approach.

OlenaYefymenko:
> Would you be open to such a change in this PR?

@ofek thank you for the detailed explanation. Yes, I will make the necessary changes. I'm just waiting for the final decision from @nedbat

nedbat:
I hope this doesn't seem like a cop-out: I think there are good arguments for both side, and I'd like to support both prefixed and un-prefixed settings, aiming for the least surprise. The only complexity here is that in pyproject.toml, we need to insist on the prefix. We already have precedent for this.

In tomlconfig.py we have this:
```python
prefixes = ["tool.coverage."]
for prefix in prefixes:
    real_section = prefix + section
```
This code looks silly (why iterate over a fixed one-element list?), but it's that way because config.py has this:
```python
class HandyConfigParser(configparser.ConfigParser):
    def __init__(self, our_file: bool) -> None:
        """Create the HandyConfigParser.

        `our_file` is True if this config file is specifically for coverage,
        False if we are examining another config file (tox.ini, setup.cfg)
        for possible settings.
        """
        self.section_prefixes = ["coverage:"]
        if our_file:
            self.section_prefixes.append("")
```
That code is there because we insist on the "coverage:" prefix in tox.ini and setup.cfg, but are flexible in .coveragerc or in an explicitly specified config file.

It would take just adding two lines to tomlconfig.py to let it read either style of section, while being strict about the sections in pyproject.toml.

To avoid over-complicating the docs, I'd mention the flexibility of section names in the opening paragraphs about syntax, but in the examples, only show the fully prefixed syntax. Maybe the tab name could be just "toml"?

I hope everyone will be happy with this approach.

webknjaz:
Sounds reasonable.

ofek:
I think it would be good to mention the possibility of either naming in the opening but I disagree about showing the fully prefixed syntax in the examples because then that would encourage folks to copy/paste that forever. I would imagine that there would be a new tab that comes first [here](https://coverage.readthedocs.io/en/7.11.3/config.html#sample-file):

image

This new tab would have the name `.coveragerc.toml` as implemented in this PR and the second tab would be `pyproject.toml`. At this point I think two options would be reasonable for the INI-formatted files:

1. Maintain their tabs and add a note to discourage their use either in the title or in the content after clicking.
2. Extract them into a separate collapsible section for legacy files.

Making the new standalone TOML file the one that is shown by default as well as its syntax would then maintain complete compatibility with the current documentation that starts the sections with the name of the options and no prefix:

image

This would also match strategy of the `pytest` [config docs](https://docs.pytest.org/en/stable/reference/customize.html#configuration-file-formats) placing what users are likely to prefer first on the page. I won't go into it here but there is a strong preference amongst developers for TOML, YAML, etc. over INI-style partly due to how uncommon it is nowadays and partly due to the ease of parsing such configuration with tooling written in other languages (never in the standard library).

In any case even if there is disagreement here about the preferred file format I think the prefix should not be prominent when, like I mentioned, it really is just an artifact of a packaging quirk in our specific language. I'm heavily involved in Python packaging now but I wasn't around when that was happening, so I'm sorry about that.

webknjaz:
Agreed, a dedicated tab sounds good. That's sounds closer to what the original feature request was after.

nedbat:
I appreciate the thought and effort you are putting into this.  Four tabs sounds reasonable. I'd prefer this order, and I've snuck in a comment to help people understand the flexibility in the TOML syntax:
image


This patch does it:
```diff
diff --git a/doc/cog_helpers.py b/doc/cog_helpers.py
index 1ff4e1aa..be681074 100644
--- a/doc/cog_helpers.py
+++ b/doc/cog_helpers.py
@@ -76,7 +76,7 @@ def show_configs(ini, toml):
     The equivalence is checked for accuracy, and the process fails if there's
     a mismatch.

-    A three-tabbed box will be produced.
+    A four-tabbed box will be produced.
     """
     ini, ini_vals = _read_config(ini, "covrc")
     toml, toml_vals = _read_config(toml, "covrc.toml")
@@ -89,11 +89,15 @@ def show_configs(ini, toml):
             )

     ini2 = re.sub(r"(?m)^\[", "[coverage:", ini)
+    toml2 = "# You can also use sections like [tool.coverage.run]\n"
+    toml2 += toml.replace("[tool.coverage.", "[")
+
     print()
     print(".. tabs::\n")
     for name, syntax, text in [
+        ("pyproject.toml", "toml", toml),
         (".coveragerc", "ini", ini),
-        (".coveragerc.toml or pyproject.toml", "toml", toml),
+        (".coverage.toml", "toml", toml2),
         ("setup.cfg or tox.ini", "ini", ini2),
     ]:
         print(f"    .. code-tab:: {syntax}")
```
BTW, I hadn't really noticed until now, but this PR names the file `.coveragerc.toml`, when it should be `.coverage.toml`.

We can work on the exact wording of the opening syntax section when the time comes.

Also, it's impossible to account for all the syntax possibilities. TOML is flexible enough that either of these will work in pyproject.toml:

```toml
[tool.coverage.run]
branch = true
core = "sysmon"

[tool.coverage]
run = { branch = true, core = "sysmon" }
```

webknjaz:
> BTW, I hadn't really noticed until now, but this PR names the file `.coveragerc.toml`, when it should be `.coverage.toml`.

I requested that name originally since it sounded more familiar. The precedent is `.pylintrc`/`.pylintrc.toml`/`pylintrc`/`pylintrc.toml`.

Also, wouldn't `.coverage.toml` be confused with the `.coverage` SQLite db?

nedbat:
>>    BTW, I hadn't really noticed until now, but this PR names the file .coveragerc.toml, when it should be .coverage.toml.
> 
> I requested that name originally since it sounded more familiar. The precedent is .pylintrc/.pylintrc.toml/pylintrc/pylintrc.toml.
> 
> Also, wouldn't .coverage.toml be confused with the .coverage SQLite db?

Good points! `.coveragerc.toml` it is!

serban:
This is slightly off topic, but since it came up:

I prefer not to hide configuration files in my repositories. In addition to `.coveragerc.toml`, it would be great to support the same functionality at `coveragerc.toml`, without the leading dot.

I'd be happy to take my bike shedding to a new issue so as to not derail this pull request. Thanks for your effort on this!

webknjaz:
@serban many people prefer the opposite — not seeing many files in repo roots. That's why a lot of them almost religiously dump hundred of lines into `pyproject.toml`, I imagine. It may be a good idea to support both, it's a common convention. I'm not sure if we should grow the scope of the PR, though.

tests/test_config.py
```diff
@@ -1007,23 +1016,72 @@ def test_no_toml_installed_pyproject_no_coverage(self) -> None:
...
+            branch = false
+            """,
+        )
+        cov = coverage.Coverage()
+
+        assert cov.config.timid is True
+        assert cov.config.data_file == ".toml-data.dat"
+        assert cov.config.branch is True
+
+    def test_coveragerc_toml_unprefixed_syntax(self) -> None:
```
webknjaz, line 1064: It'd be a good idea to also add a test making sure that unprefixed sections are ignored if present in `pyproject.toml`.

OlenaYefymenko: Yep, thank you. I will add it and will also need to fix the linting errors

AA-Turner:
> BTW, I hadn't really noticed until now, but this PR names the file `.coveragerc.toml`, when it should be `.coverage.toml`.

I'm not sure I agree that `.coverage.toml` (config file) would be confused for `.coverage` (results DB). They have different filenames, and one is a plain text file. I'd also suggest that the 'rc' suffix can be unhelpful -- I had to google for what it meant, and it seems that there's [no consensus answer](https://stackoverflow.com/a/11030607)!

Other config (`.pytest.toml`, `.ruff.toml`, `pyrefly.toml`, `ty.toml`, etc) just uses the project name, which I think is more familiar / less surprising when getting started.

A

nedbat:
> I'm not sure I agree that `.coverage.toml` (config file) would be confused for `.coverage` (results DB).

The problem isn't that people will confuse the two. The default database name is `.coverage`, and if you use `parallel=true`, then other databases are named `.coverage.somethingxyz`, and `coverage combine` will look for `.coverage.*` to combine.  We could exclude `*.toml` from combining, but now it's getting fiddly.

The good news is that you can name your configuration file anything you want if you are willing to use a command-line option or environment variable to refer to it.

ofek:
@OlenaYefymenko Heads up that a rebase is now required because there are conflicts.

OlenaYefymenko:
> @OlenaYefymenko Heads up that a rebase is now required because there are conflicts.

@ofek yep. I will do the rebase, yesterday I had some issues with it for some reason. 
In the next commit I will also add the test, and that will be the final update. Thanks for pointing it out.

tests/test_config.py
```diff
@@ -540,15 +545,16 @@ def test_unknown_option(self) -> None:
...
         self.make_file(
-            "pyproject.toml",
+            filename,
             """\
             [tool.coverage.run]
             xyzzy = 17
             """,
         )
-        msg = r"Unrecognized option '\[tool.coverage.run\] xyzzy=' in config file pyproject.toml"
+        msg = f"Unrecognized option '\\[tool.coverage.run\\] xyzzy=' in config file {filename}"
```
webknjaz: This would be a bit nicer.
```suggestion
        msg = fr"Unrecognized option '\[tool.coverage.run\] xyzzy=' in config file {filename}"
```

OlenaYefymenko:
@nedbat please take a look at the PR to see if it's ready to be merged

nedbat (approved):

nedbat:
Thanks for persisting with this! <3

nedbat:
This landed in commit 21e726fcfbb5282bf9f72f44c764dd7bcfb7d9d4

nedbat:

PR #1956, opened 2025-04-21
add free-threaded CI jobs

ngoldbaum:
I happened to notice today and `coveragepy` ships a free-threaded wheel but doesn't have testing against free-threaded Python set up in CI. Seems worth adding.

Let me know if you'd prefer I not add so many more CI jobs and how you'd like to handle that, e.g. by only adding `3.14t` jobs.


nedbat:
I'm not opposed to having more jobs.  I had tried to add free-threading builds to the nightly jobs, and ran into virtualenv problems (https://github.com/nedbat/coveragepy/commits/nedbat/free-threading-tox-nightly/), but this seems to work fine for the test suite?  I don't understand the difference.

ngoldbaum:
Let me try to see if the `nightly` job I've touched here passes on my fork when I manually run it.

ngoldbaum:
Oh I misunderstood, you *already* have nogil testing in the nightly CI.

Maybe the difference is you're using the deadsnakes build? The latest version of `setup-python` has support for 3.13t and 3.14t.

nedbat:
TBH, I'm getting confused myself about the variations.  I tried adding the free-threaded builds to my tox.ini and that didn't work.  But that is not what you are proposing!

PR #1964, opened 2025-05-09
Fix TypeError: Strings must be encoded before hashing

apetti1920:
This PR fixes an issue that occurs when using pytest with both coverage and pytest-testmon plugins. The error happens in the combine_parallel_data function when hasher.update(fobj.read()) tries to hash content that might sometimes be a string instead of bytes, despite opening the file in binary mode.

```
INTERNALERROR> Traceback (most recent call last):
...
INTERNALERROR> File "/opt/venv/lib/python3.10/site-packages/coverage/data.py", line 158, in combine_parallel_data
INTERNALERROR> hasher.update(fobj.read())
INTERNALERROR> TypeError: Strings must be encoded before hashing
```

This just adds a simple check to ensure the content is always bytes before passing it to the hasher's update method. This is a defensive programming approach that makes the code more robust against unexpected input types.

### DIFF
--- a/coverage/data.py
+++ b/coverage/data.py
@@ -162,7 +162,12 @@ def combine_parallel_data(
 
         with open(f, "rb") as fobj:
             hasher = hashlib.new("sha3_256", usedforsecurity=False)
-            hasher.update(fobj.read())
+            content = fobj.read()
+
+            # Ensure content is bytes before hashing
+            if isinstance(content, str):
+                content = content.encode('utf-8')
+            hasher.update(content)
             sha = hasher.digest()
             combine_this_one = sha not in file_hashes
 


nedbat:
Thanks. I don't understand who is calling what in the wrong way.  Can you give me a simple way to reproduce the error?

nedbat:
In particular, I am curious how this can be true:
> content that might sometimes be a string instead of bytes, despite opening the file in binary mode.

nedbat:
In fact, the more I look at this, the less I want to accept this change. On the face of it, a file opened in "rb" mode can only produce bytes.  I know you are getting this actual behavior, but I suspect some component here is inappropriately fiddling with the open() builtin, or some other weirdness.

apetti1920:
> In fact, the more I look at this, the less I want to accept this change. On the face of it, a file opened in "rb" mode can only produce bytes. I know you are getting this actual behavior, but I suspect some component here is inappropriately fiddling with the open() builtin, or some other weirdness.

This is exactly what it was, one of our tests patched out builtin.open and some interaction between coverage, testmon and the patch caused the .coverage file to become corrupted.  Sorry for the PR and thanks for all the work on coverage!

PR #1978, opened 2025-05-28
Prevent module reinitialisation with multi-phase init

AA-Turner:
cc @nedbat 

xref:
* 80e977cafcdfe47b842665e3a971fca7df931694
* #1977

Changes written by a human ;-)

The ctracer module still uses static types (e.g. `CTracerType`), so is not [isolated](https://docs.python.org/3/howto/isolating-extensions.html) and may have undefined behaviour or crashes on use in subintepreters or after `Py_Finalize` cycles. Here, we raise an explicit ImportError for such cases.

Also, change the `Py_GIL_DISABLED` check to a version check for the `Py_mod_gil` slot (it is only the unstable API that is guarded behind `Py_GIL_DISABLED`), use `PyMODINIT_FUNC` for the init function, and use designated initialisers (I find them easier to read).

A

### DIFF
--- a/coverage/ctracer/module.c
+++ b/coverage/ctracer/module.c
@@ -9,10 +9,19 @@
 
 #define MODULE_DOC PyDoc_STR("Fast coverage tracer.")
 
-/* Module execution function for PEP 489 multi-phase initialization */
+static int module_loaded = 0;
+
 static int
 tracer_exec(PyObject *mod)
 {
+    // https://docs.python.org/3/howto/isolating-extensions.html#opt-out-limiting-to-one-module-object-per-process
+    if (module_loaded) {
+        PyErr_SetString(PyExc_ImportError,
+                        "cannot load module more than once per process");
+        return -1;
+    }
+    module_loaded = 1;
+
     if (CTracer_intern_strings() < 0) {
         return -1;
     }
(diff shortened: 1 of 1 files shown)


nedbat:
Thanks, human!

nedbat:
Hmm, turns out this causes a problem in the coverage measurement of coverage itself.  If I comment out the re-import check everything seems to work, but I need to understand more.

AA-Turner:
I added the check & error as multi-phase assumes that modules are safe for subinterpreters by default. The relevant documentation is https://docs.python.org/dev/howto/isolating-extensions.html. On a cursory examination I think `CTracerType` & `CFileDispositionType` would need to be converted to heap types. I don't know how coverage's self-measurement works though, and there might be a simpler fix there.

A

nedbat:
This isn't about subinterpreters explicitly.  The self-measurement involves deleting `sys.modules["coverage"]` and importing it again, which is when the failure happens.   So it's all within one process.  Would I still need to change the types?  Or maybe just avoid initting them twice?  There are also static interned strings which should only be interned once.

AA-Turner:
Sorry for the crossed wires. I was referring to [this note](https://docs.python.org/dev/c-api/module.html#multi-phase-initialization):

> All modules created using multi-phase initialization are expected to support sub-interpreters, or otherwise explicitly signal a lack of support. This is usually achieved by isolation or blocking repeated initialization, as above. A module may also be limited to the main interpreter using the Py_mod_multiple_interpreters slot.

Per that same section, *sharing Python objects between module instances would likely cause crashes or undefined behavior.*. The side effect of the explicit error is that deleting the module from `sys.modules` and then re-importing it is expected to & will fail, as the import system attempts to re-initialise it in the same process. Clearly unwanted behaviour here, though. 

Would `importlib.reload` work instead of deleting modules from `sys.modules`? That function has no effect for extension modules & so wouldn't casue issues. Alternatively, could the `ctracer` extension module be exempted from the deletion?

A

nedbat:
I'm assuming that the warning is actually about sharing mutable Python objects between instances.  In my case, they are types and interned strings.

This commit (https://github.com/nedbat/coveragepy/commit/8381ea787ef56b67ca158d6f43c546e719ef0d32) seems good, but I welcome more experienced views. (Updated to use a mutex: https://github.com/nedbat/coveragepy/commit/2d39c8cb6d0099102d2624baab31671ed786b1e0 (Update: never mind, PyMutex_etc is new in 3.13))

nedbat:
This is now released as part of [coverage 7.9.0](https://pypi.org/project/coverage/7.9.0).

PR #1984, opened 2025-06-14
Enable measuring test coverage for Python archive (.par) files

itamaro:
".par" files are another common(?) (at least at Meta) zip-based binary format for python applications. This change enables coverage to unpack and use the unpacked files under any .par archives as source files.

we've been carrying this patch for years, so I thought I'd try to upstream it, as it could benefit any PAR users out there (and should be harmless for anyone else).


nedbat:
This is now released as part of [coverage 7.9.2](https://pypi.org/project/coverage/7.9.2).

PR #1998, opened 2025-07-12
run: Support system signal as a coverage report dump trigger.

ark-g:
Implement a feature that allows users to generate a coverage report by sending a signal. This is particularly useful for obtaining coverage reports from continuously running Python scripts, such as server implementations or monitoring utilities.
Currently supported only two signals SIGUSR1 and SIGUSR2.


coverage/cmdline.py
```diff
@@ -227,7 +229,15 @@ class Opts:
         "", "--version", action="store_true",
         help="Display version information and exit.",
     )
-
+    dump_signal = optparse.make_option(
+        '', '--dump_signal', action='store', metavar='DUMP_SIGNAL',
```
nedbat:
```suggestion
        '', '--dump-signal', action='store', metavar='DUMP_SIGNAL',
```
Options use hyphens, not underscores.

coverage/cmdline.py
```diff
@@ -227,7 +229,15 @@ class Opts:
         "", "--version", action="store_true",
         help="Display version information and exit.",
     )
-
+    dump_signal = optparse.make_option(
```
nedbat: Overall, I think this should be called "save signal" instead of "dump signal" throughout.

nedbat: These make_option definitions are in rough alphabetical order, so this should be further up.

coverage/cmdline.py
```diff
@@ -525,6 +536,7 @@ def get_prog_name(self) -> str:
             Opts.parallel_mode,
             Opts.source,
             Opts.timid,
+            Opts.dump_signal,
```
nedbat: Keep these items alphabetized.

nedbat (changes requested):
A few small changes.  I'll adjust the English description once this merges.

nedbat:
Thanks for this!  I want to have tests for this feature.  `test_process.py` has some tests you could use as models, or I could give you some suggestions.

ark-g:
> Thanks for this! I want to have tests for this feature. `test_process.py` has some tests you could use as models, or I could give you some suggestions.

Hi,
First, I want to thank you for your great work and support for the Python coverage utility.
I don't see any option to create a test within the current environment. From what I understand, the test for the new feature should send a signal during the execution of the target script, but I haven't found this capability in the existing test examples or infrastructure. I would appreciate your suggestions on how to implement the test for the new feature.

Thanks.

nedbat:
Thanks, what are your thoughts on a test?  Let me know how far you want to take this, I can take it over when you want.

ark-g:
I think that test should run some script with 2 stages.
```
    def test_save_signal(self) -> None:
        test_file = "dummy_hello.py"
        self.make_file(test_file, """\
            import time
            for i in range(10):
              time.sleep(1)
              if i < 5:
                print(f'{i:02d} Hello from the main script!')
              else:
                print(f'0x{i:02x} Hello X!')
            print('Done and goodbye')
            """)
        dummy_prc = self.run_command(f"coverage run --signal-save USR1 {test_file}")
        <Send signal first time>
        <Make sure that report reflects expected coverage>
        <sleep 5 secs>
       <Send signal second time>
       <Make sure that report reflects expected coverage>
       <Sleep 6 secs>
       <Make sure that report reflects 100% coverage>
```
But for the above I need to run script in background and to communicate with the running script dynamically (to monitor it's output).
Current infrastructure doesn't support such an ability, AFAIU.
Let me know, please,  how do you see this test ?

nedbat:
The test would have to be something like that, but I don't want it to take 10 seconds.  I think it would be enough to have two cases in the test: in both, start an infinite loop process with `coverage run --save-signal=USR1`.  In both cases, we use `kill -s KILL` to end the process.  In one case, that's all we do, and we see no .coverage file is created.  In the other case, we use `kill -s USR1` before the KILL, and we see that a .coverage file is created.

nedbat:
BTW, we also need to do the right thing on Windows.  I think it's best to add "not on Windows" to the option help, and print an error if it's used on Windows.  The test should check this also.

ark-g:
> The test would have to be something like that, but I don't want it to take 10 seconds. I think it would be enough to have two cases in the test: in both, start an infinite loop process with `coverage run --save-signal=USR1`. In both cases, we use `kill -s KILL` to end the process. In one case, that's all we do, and we see no .coverage file is created. In the other case, we use `kill -s USR1` before the KILL, and we see that a .coverage file is created.

Do you have an example how to initiate infinite loop process in existing test infrastructure  or I should use Subprocess module in the test ?

ark-g:
> > The test would have to be something like that, but I don't want it to take 10 seconds. I think it would be enough to have two cases in the test: in both, start an infinite loop process with `coverage run --save-signal=USR1`. In both cases, we use `kill -s KILL` to end the process. In one case, that's all we do, and we see no .coverage file is created. In the other case, we use `kill -s USR1` before the KILL, and we see that a .coverage file is created.
> 
> Do you have an example how to initiate infinite loop process in existing test infrastructure or I should use Subprocess module in the test ?

I do not see much more purpose on the second test (with only KILL signal), but as you wish ... ;)

nedbat:
Did you see this pull request?  https://github.com/ark-g/coveragepy/pull/1   It has some other changes for the code in cmdline.py.   I wanted the test without the signal to show that the signal is needed to get the data file.

ark-g:
> Did you see this pull request? [ark-g#1](https://github.com/ark-g/coveragepy/pull/1) It has some other changes for the code in cmdline.py. I wanted the test without the signal to show that the signal is needed to get the data file.

I think that the first test is also demonstrating this fact, since the final coverage report doesn't include the line that is sending KILL signal, but does include the line with USR1 signal.

Do you think that I can merge this PR ?

nedbat:
I want you to take the changes I gave you in the pull request in your repo.

ark-g:
> I want you to take the changes I gave you in the pull request in your repo.

Now half of tests failed

nedbat:
You need to merge the pull request in your repo. I can't do it, and there will be conflicts to resolve.

tests/test_process.py
```diff
@@ -687,6 +716,51 @@ def test_module_name(self) -> None:
         out = self.run_command("python -m coverage")
         assert "Use 'coverage help' for help" in out
 
+    @pytest.mark.skipif(env.WINDOWS, reason="This test is not for Windows")
+    def test_save_signal_usr1(self) -> None:
```
nedbat, line 720: I think maybe we don't need this test since `test_save_signal` covers it?  Same for `test_save_signal_kill`, though maybe there's some subtlety I'm overlooking?

nedbat:
Let me know when you are ready to merge.

tests/test_process.py
ark-g: In general I agree, but the `test_save_signal_usr1`  is more precise and check conditions more tight, so I would leave it. Regarding `test_save_signal_kill` I will remove it since looks like we're both agree that it is redundant.

ark-g:
> Let me know when you are ready to merge.

I am ready

ark-g:
> Let me know when you are ready to merge.

Looks like all done and we are ready to merge

nedbat:
Thanks for all the work! I rebased and combined commits, and landed them at commit 43bf46b1f202164bcf37c209cbde85dca469d5fd.

nedbat:
This is now released as part of [coverage 7.10.0](https://pypi.org/project/coverage/7.10.0).

PR #2004, opened 2025-07-21
Use flaky pytest fixture instead of decorator

### LINKED ISSUE #2003: Please consider improving support for `pytest-rerunfailures` (in place of `flaky`)
We're slowly working towards replacing the old nose-era [flaky](https://pypi.org/project/flaky/) plugin with [pytest-rerunfailures](https://pypi.org/project/pytest-rerunfailures/). The two are largely drop-in replacements, provided pytest fixtures are used rather than `flaky.flaky` decorator. At the same time, there are slight feature differences, and pytest is only able to use one of them at the same time — which means that if you have both `flaky` and `pytest-rerunfailures` installed, you sometimes need to do `-p no:flaky` to prevent `flaky` from taking over.

Given that coverage's test suite already uses pytest, would you accept a patch replacing direct `flaky.flaky` decorator use in favor of the `pytest.mark.flaky` marker? Or possibly `flaky` with `pytest-rerunfailures` entirely?

Local patching, but that's not very efficient.

**Additional context**
n/a

mgorny:
Replace the direct use of `flaky.flaky` decorator with the equivalent pytest fixture.  This is both more canonical, and has the additional advantage of being compatible with the more modern `pytest-rerunfailures` plugin, that can be used in place of `flaky`.

Fixes #2003

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -25,8 +25,6 @@ from typing import (
 )
 from collections.abc import Iterable, Iterator
 
-import flaky
-
 import coverage
 from coverage import env
 from coverage.debug import DebugControl
@@ -388,15 +386,6 @@ class DebugControlString(DebugControl):
     def get_output(self) -> str:
         """Get the output text from the `DebugControl`."""
         return self.io.getvalue()
-
-
-TestMethod = Callable[[Any], None]
-
-def flaky_method(max_runs: int) -> Callable[[TestMethod], TestMethod]:
-    """flaky.flaky, but with type annotations."""
-    def _decorator(fn: TestMethod) -> TestMethod:
-        return cast(TestMethod, flaky.flaky(max_runs)(fn))
-    return _decorator
 
 
 def all_our_source_files() -> Iterator[tuple[Path, str]]:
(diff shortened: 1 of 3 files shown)


nedbat:
Thanks. We can remove it from the requirements.in file (https://github.com/nedbat/coveragepy/blob/master/requirements/pytest.in#L9) too.  `make upgrade` should rebuild the .txt files.

nedbat:
I can handle the requirements later if you want.  But why can't I find docs for `pytest.mark.flaky`?

mgorny:
I think there's a slight misunderstand. `pytest.mark.flaky` is provided both by flaky and pytest-rerunfailures packages:

- https://github.com/box/flaky?tab=readme-ov-file#pytest-marker
- https://pypi.org/project/pytest-rerunfailures/#re-run-individual-failures

For a minimal change, I've left it at flaky. If you want, I can swap it for pytest-rerunfailures — it's a matter of replacing the dep and removign `--no-flaky-report` (pytest-rerunfailures doesn't do such a thing).

nedbat:
You're right, I did misunderstand.  OK, let's make this change and keep pushing forward.

mgorny:
Thanks!

PR #2009, opened 2025-07-25
Loosen dependency version range for requests library

SamuelMarks:
I was installing [`mkdocs-material`](https://pypi.org/project/mkdocs-material/) which requests a newer version than coveragepy requires.

> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
coveragepy 1.6.0 requires requests==2.25.0, but you have requests 2.32.4 which is incompatible.

Or more explicitly on reverse:

> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
mkdocs-material 9.6.15 requires requests~=2.26, but you have requests 2.25.0 which is incompatible.

This PR just loosens up your dependency. Let me know if you want any alterations, like more specific `requests` version range.

### DIFF
--- a/requirements/dev.pip
+++ b/requirements/dev.pip
@@ -6,8 +6,6 @@ attrs==25.3.0
     # via
     #   hypothesis
     #   scriv
-backports-tarfile==1.2.0
-    # via jaraco-context
 build==1.2.2.post1
     # via check-manifest
 cachetools==6.1.0
@@ -20,7 +18,7 @@ charset-normalizer==3.4.2
     # via requests
 check-manifest==0.50
     # via -r requirements/dev.in
-click==8.1.8
+click==8.2.1
     # via
     #   click-log
     #   scriv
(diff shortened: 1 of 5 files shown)


nedbat:
I'm not understanding something: coverage.py has no run-time dependencies, these are development requirements.  How are you using the coveragepy repo?  Also, the message says, "coveragepy 1.6.0 requires requests==2.25.0"  There was never a coverage 1.6.0 release.

SamuelMarks:
@nedbat Ohhh silly me, checking `pip freeze`:
```
coverage==4.5.3
coveragePy==1.6.0
```

Your project:
https://pypi.org/project/coverage/ [Jul 24, 2025]

That other random one:
https://pypi.org/project/coveragePy/ [updated Dec 16, 2020]

PR #2019, opened 2025-08-08
use atomics for `started` and `activity` flags

kumaraditya303:


coverage/ctracer/tracer.c
```diff
@@ -981,9 +981,10 @@ CTracer_call(CTracer *self, PyObject *args, PyObject *kwds)
 static PyObject *
 CTracer_start(CTracer *self, PyObject *args_unused)
 {
+    assert(atomic_load_int(&self->started) == FALSE);
```
kumaraditya303: This assert fails without https://github.com/nedbat/coveragepy/commit/9533be81e553725110fda9c6ee0887de45225ccb while running under pytest-run-parallel

nedbat:
@kumaraditya303 what do you think of #2020?

kumaraditya303:
> what do you think of https://github.com/nedbat/coveragepy/pull/2020?

I think using c11 atomics is definitely better than writing manual atomic platform dependent code if using newer msvc is not an issue, if you prefer it as well then I will update this PR based on that.

nedbat:
#2020 does look better to me.  As I've said elsewhere, I no basically nothing about the versions of compilers and so on.  It seems to work in CI, so let's go with it.

kumaraditya303:
Okay, I have updated the PR to use c11 atomics.

PR #2020, opened 2025-08-08
Demo: use c11 atomics

colesbury:
This shows how to use C11 atomics in `coveragepy`.

### DIFF
--- a/coverage/ctracer/tracer.c
+++ b/coverage/ctracer/tracer.c
@@ -982,8 +982,8 @@ static PyObject *
 CTracer_start(CTracer *self, PyObject *args_unused)
 {
     PyEval_SetTrace((Py_tracefunc)CTracer_trace, (PyObject*)self);
-    self->started = TRUE;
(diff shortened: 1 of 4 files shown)


nedbat:
Thanks.  For us compiler newbs who need to understand what compilers are used for what versions on what platforms, is there a CPython reference I can use?

colesbury:
I think the choice is made by setuptools, configurable by your project. For Windows it will be MSVC. For Linux and macOS it will be GCC or Clang.

colesbury:
For CPython supported platforms and compilers, see [PEP 11](https://peps.python.org/pep-0011/#tier-1). It's basically:

Windows: MSVC
Everything else: GCC or Clang

GCC vs. Clang doesn't really matter much. Clang tries hard to be GCC compatible. Now that Clang is popular, GCC even tries to be Clang compatible.

nedbat:
Thanks, this has been incorporated into #2019.

Issue #2007, opened 2025-07-24
Issues found running the tests with pytest-run-parallel


ngoldbaum:
**Describe the bug**

If I run the coveragepy test suite with some small modifications to run with TSAN support and under [`pytest-run-parallel`](https://github.com/Quansight-Labs/pytest-run-parallel), I see data races inside coveragepy and CPython as well as some test failures.

`pytest-run-parallel` is a pytest plugin that replaces the PyTest test executor with a function that creates a thread pool and spawns many instances of the same test to run simultaneously. It can detect thread safety issues due to use of global state in the implementation of python modules. In practice, it tends to most often find issues in test suites due to the [way it handles fixtures by implicitly sharing module-scope fixtures between tests](https://github.com/Quansight-Labs/pytest-run-parallel?tab=readme-ov-file#caveats). However, it's still worth going through the exercise IMO because it's one of the few automated ways to exercise big chunks of a Python codebase in a multithreaded environment. See https://py-free-threading.github.io/testing/ for more guidance on multithreaded testing from the team working on ecosystem support for free-threaded Python.

The issues in CPython have been reported upstream (https://github.com/python/cpython/pull/136994), I'm reporting the races inside coveragepy here.

I made the following modification to build coveragepy with thread sanitizer when I run a `py314t` environment:

```diff
diff --git a/tox.ini b/tox.ini
index 81bb0481..c162d7e3 100644
--- a/tox.ini
+++ b/tox.ini
@@ -19,6 +19,7 @@ extras =
 deps =
     -r requirements/pip.pip
     -r requirements/pytest.pip
+    pytest-run-parallel
     py3{9,10,11}: -r requirements/light-threads.pip
 
 # Windows can't update the pip version with pip running, so use Python
@@ -38,6 +39,7 @@ setenv =
     PYTHONWARNDEFAULTENCODING=
     py3{10,11,12,13,14}: PYTHONWARNDEFAULTENCODING=1
     py3{10,11,12,13,14}: PYTHONWARNINGS=ignore::EncodingWarning:pip._internal.utils.subprocess
+    py314t: CFLAGS=-fsanitize=thread
     # Disable CPython's color output
     PYTHON_COLORS=0
 
@@ -51,7 +53,7 @@ commands =
     python igor.py remove_extension
 
     # Test with the PyTracer
-    python igor.py test_with_core pytrace {posargs}
+    # python igor.py test_with_core pytrace {posargs}
 
     # Build the C extension and test with the CTracer
[...]

PR #2023, opened 2025-08-09
add tsan CI

kumaraditya303:


nedbat:
Were you planning to describe what this is for and how it improves things?

rgommers:
Related to gh-2007 I'd think. @kumaraditya303?

PR #2028, opened 2025-08-11
test: Find source directory relative to tests rather than `coverage`

mgorny:
Use the path to the `tests/helpers.py` module itself rather than the `coverage` package to determine what the source directory is. The former is always imported from the source tree, while the latter does not necessarily have to -- in particular, in Gentoo we are installing it to a temporary directory to keep the source directory clean of artifacts.


nedbat:
Thanks for keeping me honest :D

PR #2030, opened 2025-08-13
Add run-report to command line

Corentin-pro:
I added a `run-report` command that can be used to show a report without creating a .coverage file.

I may be one of the very few to really want such a feature as I don't like adding unnecessary files and I/O to my SSD storage. This is obviously desirable only for small code base.

Instead of opening an issue I thought it would be better to already give a solution to start a conversation.

Feel free to close this PR if this is definitely not something you think is desirable as a feature.

One caveat of this implementation is the fact that I merged the arguments or `run` and `report` and the `-m` argument is in conflict so I left it for the `run` only and forced it for the report (always showing missing lines).

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -563,7 +563,8 @@ COMMANDS = {
             Opts.timid,
             ] + GLOBAL_ARGS,
         usage="[options] <pyfile> [program options]",
-        description="Run a Python program, measuring code execution and report without saving in file.",
+        description=(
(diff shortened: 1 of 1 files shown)


nedbat:
Thanks, I appreciate the way you approached this.  But I am reluctant to add this to coverage.  It's an unusual case to need to avoid writing a data file.  If you need to, you can use the API to coordinate the coverage steps without writing a data file.

PR #2032, opened 2025-08-17
Small optimizations to combine

alex:
a) Avoid copying lists in executemany_void: while some callers provide an iterator, others were already providing a list and in that case we had to copy it
b) Avoid unpacking and re-packing contexts

In total this shaves about 10% of the combine time in pyca/cryptography's CI

### DIFF
--- a/coverage/sqldata.py
+++ b/coverage/sqldata.py
@@ -685,7 +685,7 @@ class CoverageData:
 
             # Get contexts data.
             with con.execute("select context from context") as cur:
-                contexts = [context for (context,) in cur]
+                contexts = cur.fetchall()
 
             # Get arc data.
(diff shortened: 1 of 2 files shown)


nedbat:
Thanks!

alex:
My pleasure! I have an idea for a bigger perf win, but we'll see if it pans
out.

All that is necessary for evil to succeed is for good people to do nothing.

nedbat:
Combining has had a number of performance problems over the years.  Looking forward to what you have.

alex:
The core idea is "avoid going sqlite3 -> python -> sqlite3" by using `ATTACH DATABASE`.

nedbat:

PR #2034, opened 2025-08-17
use SQLite URIs for in-memory databases

alex:
This lets us ATTACH DATABASE between in-memory databases.


alex:
Should unblock https://github.com/nedbat/coveragepy/pull/2033

nedbat:

PR #2069, opened 2025-10-20
WIP, BUG: futures with dir switching

### LINKED ISSUE #2065: BUG: tmpdir directory context and ProcessPoolExecutor prevent tracing report
It seems that the combination of the `pytest` fixture usage `tmpdir.as_cwd()` and a `ProcessPoolExecutor` prevents the tracing of a dispatched parallel function. A very simple reproducer is possible per the code below, using Python `3.13.2`, `coverage` `7.10.7`, `pytest` `8.4.2`. Is the coverage report likely getting lost because of the combination of an extra process *and* a directory change? Should this be an easy fix? Do you want me to try to patch it? Suggested workaround in the meanwhile?

To reproduce, you can copy the 3 source files below to a single directory, and run the commands:

```
coverage run -m pytest
coverage combine
coverage html
```

```python
# probe.py
from concurrent.futures import ProcessPoolExecutor

def add(a, b):
    return a + b

def dispatcher():
    futures = []
    with ProcessPoolExecutor() as executor:
        futures.append(executor.submit(add, 2, 2))
    for future in futures:
        future.result()

```

```python
# test_probe.py
from probe import dispatcher

def test_add(tmpdir):
    with tmpdir.as_cwd():
        dispatcher()
```

`.coveragerc`:

```
[run]
branch = true
concurrency = multiprocessing
parallel = true
```

And you end up with the failure to report the flushing of the dispatched function in another process:

Image

You can just deactivate the directory switching to restore a proper coverage report:

```diff
--- a/test_probe.py
+++ b/test_probe.py
@@ -1,5 +1,6 @@
 from probe import dispatcher
+import contextlib
 
 def test_add(tmpdir):
-    with tmpdir.as_cwd():
+    with contextlib.nullcontext():
         dispatcher()
```

Image

`coverage combine` only finds one data file in the problem scenario, so does indeed seem likely that one of the coverage files gets "misplaced" by the directory context. I'm happy to try to fix this if you want to provide some basic guidance--clearly it is easy to reproduce and regression test. Hopefully it is obvious why we wouldn't just remove the usage of `tmpdir`--we have real-world scenarios where files are being written during testing and we don't want those polluting the working directory. It is worth noting that `tmpdir` is no longer the preferred fixture for accomplishing this kind of thing, but in some local testing I don't see any better behavior with other approaches to switching the directory context.

tylerjereddy:
* Fixes gh-2065.

* Add a regression test for the failure to collect `.coverage` files when parent and child (`concurrent.futures`) processes have different directory contexts. This test fails for the expected reason it seems.

* Add a patch that fixes the problem described in the matching issue, but for some reason does not manage to restore the desired behavior in the regression test (something I'm not understanding about the "meta" testing?)

### DIFF
--- a/coverage/multiproc.py
+++ b/coverage/multiproc.py
@@ -113,6 +113,8 @@ def patch_multiprocessing(rcfile: str) -> None:
             """Get the original preparation data, and also insert our stowaway."""
             d = original_get_preparation_data(name)
             d["stowaway"] = Stowaway(rcfile)
+            with open(".coverpath", "w", encoding="utf-8") as cpath:
+                cpath.write(d["orig_dir"])
             return d
 
         spawn.get_preparation_data = get_preparation_data_with_stowaway
--- a/coverage/sqldata.py
+++ b/coverage/sqldata.py
@@ -11,6 +11,7 @@ import functools
 import glob
 import itertools
 import os
+import shutil
 import random
 import socket
 import sqlite3
@@ -232,6 +233,8 @@ class CoverageData:
         no_disk: bool = False,
         warn: TWarnFn | None = None,
         debug: TDebugCtl | None = None,
+        *,
+        orig_dir = None,
     ) -> None:
         """Create a :class:`CoverageData` object to hold coverage-measured data.
 
@@ -271,6 +274,7 @@ class CoverageData:
         self._current_context: str | None = None
         self._current_context_id: int | None = None
         self._query_context_ids: list[int] | None = None
+        self.orig_dir = orig_dir
 
     __repr__ = auto_repr
 
@@ -898,6 +902,11 @@ class CoverageData:
     def write(self) -> None:
         """Ensure the data is written to the data file."""
         self._debug_dataio("Writing (no-op) data file", self._filename)
+        if self.orig_dir is not None:
+            for fname in os.listdir():
+                if fname.startswith(".coverage."):
+                    if not os.path.exists(os.path.join(self.orig_dir, fname)):
+                        shutil.copy(fname, self.orig_dir)
 
     def _start_using(self) -> None:
         """Call this before using the database at all."""
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -13,6 +13,7 @@ import random
 import re
 import sys
 import threading
+import concurrent.futures
 import time
 
 from types import ModuleType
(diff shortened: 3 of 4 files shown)


coverage/sqldata.py
```diff
@@ -898,6 +902,11 @@ def read(self) -> None:
     def write(self) -> None:
         """Ensure the data is written to the data file."""
         self._debug_dataio("Writing (no-op) data file", self._filename)
+        if self.orig_dir is not None:
+            for fname in os.listdir():
+                if fname.startswith(".coverage."):
+                    if not os.path.exists(os.path.join(self.orig_dir, fname)):
+                        shutil.copy(fname, self.orig_dir)
```
tylerjereddy, line 909: This line here does the trick to restore coverage of the missing `add()` function in the example below the fold when I install this feature branch locally:

<details>

```python
import os
import tempfile
from concurrent.futures import ProcessPoolExecutor

def add(a, b):
    return a + b

def probe_dispatcher():
    orig_dir = os.getcwd()
    with tempfile.TemporaryDirectory() as temp_dir:
        os.chdir(temp_dir)
        dispatcher()
    os.chdir(orig_dir)

def dispatcher():
    futures = []
    with ProcessPoolExecutor() as executor:
        futures.append(executor.submit(add, 2, 2))
    for future in futures:
        future.result()

if __name__ == "__main__":
    probe_dispatcher()
```

</details>

image
 
whereas suppressing that `shutil.copy()` line restores the original bug:

image

While I *can* reproduce this lack of coverage with the new regression test I added here, I can't make that test pass on this feature branch at the moment... am I missing something about the meta testing/coverage testing harness?

coverage/multiproc.py
```diff
@@ -113,6 +113,8 @@ def get_preparation_data_with_stowaway(name: str) -> dict[str, Any]:
             """Get the original preparation data, and also insert our stowaway."""
             d = original_get_preparation_data(name)
             d["stowaway"] = Stowaway(rcfile)
+            with open(".coverpath", "w", encoding="utf-8") as cpath:
+                cpath.write(d["orig_dir"])
```
tylerjereddy, line 117: This feels kind of bad... even more configuration/hacking via the `multiprocessing` monkeypatching, but it does seem to work as part of storing the original working directory for the originally described issue, so that `.coverage.` files may be copied back there.

tests/test_concurrency.py
```diff
@@ -522,6 +549,20 @@ def test_multiprocessing_simple(self, start_method: str) -> None:
             start_method=start_method,
         )
 
+    def test_gh_2065(self, start_method: str) -> None:
+        nprocs = 1
+        upto = 30
+        code = (MULTI_CODE_DIR_CHANGE).format(NPROCS=nprocs, UPTO=upto)
+        total = 0
```
tylerjereddy, line 556: `total` is unused now since I'm not really wanting to write output from my sample test case... but perhaps more importantly I can't make this new regression test pass at the moment, despite the *installed* version of this feature branch doing the right thing for the reproducing external example as noted at https://github.com/nedbat/coveragepy/pull/2069/files#r2446215263. Am I missing something about the meta-testing/test harness here?

Issue #2081, opened 2025-11-11
Discussion: HTML coverage report layout (Statement/Branch columns and excluded column)


Ryuta1005:
## Summary
This issue is to discuss potential updates to the HTML coverage report layout.

From recent conversation, two related points came up:

1. Displaying Statement Coverage and Branch Coverage as separate columns
- This could make it easier to visually notice when one metric is falling behind the other.

2. Reconsidering the `excluded` column
- The excluded column may not provide meaningful value for many users.
- It is shown only for statements, with no corresponding column for branches.
- This creates an asymmetry in the UI.

## Proposal
Discuss the HTML report layout as a whole:

- Add separate Statement and Branch coverage columns.
- Remove or relocate the excluded column to improve clarity and consistency.

## Goal
The goal is not to decide immediately, but to gather feedback from maintainers and users to determine which layout is most clear and useful.

Looking forward to hearing your thoughts.

nedbat:
In an earlier [discussion in Discord](https://discord.gg/RBw567RyWc), @Ryuta1005 shared this image:

<img width="1756" height="526" alt="Image" src="https://github.com/user-attachments/assets/87cbb776-b17a-4471-865f-c5c58d5467a2" />

fniessink:
Maybe not the feedback you're looking for, but as a developer who routinely sets their coverage target to 100% for both line and branch coverage, I'm really only interested in which lines or branches need coverage, not the counts and percentages.

Ignas:
As someone who deals with less than 100% coverage - first thing I do is sort by uncovered statements not %. I don't care about all the empty files with 100% coverage, nor do I care about small files with 20%.

How much this file contributes to the coverage % total is more important than it's actual score

nedbat:
> I don't care about all the empty files with 100% coverage

Don't overlook that you can skip fully covered files. There's a checkbox in the upper-right of the summary page.

Ryuta1005:
Thanks everyone for the comments.
I found it very interesting that the indicators developers care about can differ depending on the project’s coverage goals.

Regarding the two points in this issue:

> 1. Displaying Statement Coverage and Branch Coverage as separate columns
> 2. Reconsidering the excluded column

Personally, I think (1) is valuable, especially for teams whose coverage target is not 100%.
And for (2), I feel that the excluded column may not be necessary.

I’d love to hear how others feel about these two points.

Ryuta1005:
I’ve prepared a draft PR with the two new columns added.
[...]

PR #2085, opened 2025-11-13
Add Statement and Branch coverage columns to index.html report

Ryuta1005:
## Summary
This PR adds two new columns — Statement Coverage and Branch Coverage — to the HTML coverage report layout.
These columns appear only when branch coverage is enabled (using the --branch option).
The goal is to provide a concrete basis for discussion in [Issue #2081](https://github.com/coveragepy/coveragepy/issues/2081).
If these two columns are considered useful, this change could be merged as-is.

## Changes
- Added Statement Coverage and Branch Coverage columns to the main HTML index report (visible only when --branch is enabled).
- No functional or behavioral changes to coverage calculation logic.
- The excluded column remains unchanged for now (to be discussed in Issue #2081).

### Before
image

### After
image

## Related Issue
Discussion base: [#2081](https://github.com/coveragepy/coveragepy/issues/2081)


Ryuta1005:
Note about the `py314t` / `py315t` CI jobs:

On macOS CI, both `py314t` and `py315t` are currently red due to failures in:

- `tests/test_concurrency.py::test_coverage_stop_in_threads`

For example:

```text
assert has_started_coverage == [t.ident]
E       assert [] == [6126907392]
```
and:
```text
assert has_stopped_coverage == [t.ident]
E       assert [] == [6192033792]
```
This looks like a timing / threading issue specific to the free-threaded (nogil) builds and the tracer, rather than something caused by the HTML report changes in this PR.

Locally, I ran:
```bash
make test
```
with py313t, py314t, and py315t available via pyenv, and all test environments (including the t builds) passed:
```text
py314t: OK (...)
py315t: OK (...)
congratulations :)
```
If it would be helpful, I’m happy to open a separate issue to track the test_coverage_stop_in_threads behavior on free-threaded builds, but I kept this PR focused on the HTML layout only.

nedbat:
> This looks like a timing / threading issue specific to the free-threaded (nogil) builds

Yes, those tests are usually ok, but occasionally fail. An improvement would be welcome if you know what to do about them.

coverage/results.py
```diff
@@ -323,6 +323,16 @@ def n_executed_branches(self) -> int:
         """Returns the number of executed branches."""
         return self.n_branches - self.n_missing_branches
 
+    @property
+    def statements_ratio(self) -> tuple[int, int]:
```
nedbat: This should be `ratio_statements`, similar for branches.

nedbat:
Can you add some testing in test_results.py? We don't need separate tests for each property, but they should be covered by some test in there.

nedbat:
I like the way this is going. I'd like to work on the CSS styling once it lands.

coverage/results.py
Ryuta1005: Thanks! Renamed in this commit:[https://github.com/coveragepy/coveragepy/pull/2085/commits/1321a8c13713f0e59790200df1ae9d70e20a7557](https://github.com/coveragepy/coveragepy/pull/2085/commits/1321a8c13713f0e59790200df1ae9d70e20a7557)

Ryuta1005:
> Can you add some testing in test_results.py? We don't need separate tests for each property, but they should be covered by some test in there.

Added tests in this commit:[https://github.com/coveragepy/coveragepy/pull/2085/commits/46105c653df0c1f9424f2cb311f55a746b5f0622](https://github.com/coveragepy/coveragepy/pull/2085/commits/46105c653df0c1f9424f2cb311f55a746b5f0622)

Ryuta1005:
I've addressed the feedback so far (naming + tests),
and the PR is now ready for further review.

I’ve also removed the draft status.

nedbat:
Also, I wonder if we should add these new Results properties to the JSON report? I guess that could be a second PR.

Ryuta1005:
Thanks! Yes, I agree — adding these new Results properties to the JSON report would be useful.

Once this PR is merged, I can open a follow-up PR for the JSON report.

coverage/results.py
```diff
@@ -333,29 +333,29 @@ def ratio_branches(self) -> tuple[int, int]:
         """Return numerator/denominator for branch coverage."""
         return self.n_executed_branches, self.n_branches
 
+    def _percent(self, numerator: int, denominator: int) -> float:
```
nedbat, line 336: Great, this was what I was thinking.

coverage/results.py
```diff
@@ -333,29 +333,29 @@ def ratio_branches(self) -> tuple[int, int]:
...
-        if self.n_statements > 0:
-            numerator, denominator = self.ratio_covered
-            pc_cov = (100.0 * numerator) / denominator
-        else:
-            pc_cov = 100.0
-        return pc_cov
+        numerator, denominator = self.ratio_covered
+        if self.n_statements == 0:
+            return 100.0
+        return self._percent(numerator, denominator)
```
nedbat: I think we can make this simpler because if n_statements is zero, so is the denominator:
`return self._percent(*self.ratio_covered)`.  The star syntax will also simplify the other two `pc_*` properties.

coverage/htmlfiles/coverage_html.js
```diff
@@ -140,12 +140,19 @@ coverage.wire_up_filter = function () {
     const table_body_rows = table.querySelectorAll("tbody tr");
     const no_rows = document.getElementById("no_rows");
 
+    const header_row = table.tHead ? table.tHead.rows[table.tHead.rows.length - 1] : table.rows[0];
```
nedbat: Why are we conditionalizing on `tHead` and `tFoot`? When won't the table have a head or a foot?

Ryuta1005: Thanks! I've removed the unnecessary conditionalizing as suggested.

coverage/results.py
Ryuta1005: Thanks! Addressed in this commit: [https://github.com/coveragepy/coveragepy/pull/2085/commits/8e883137e6e557b16e11c121558caa665caeb22d](https://github.com/coveragepy/coveragepy/pull/2085/commits/8e883137e6e557b16e11c121558caa665caeb22d)

nedbat:
Thanks for all this, I'll merge this and make some adjustments.  Let me know what name I should use to credit you in the changelog and contributors files.

nedbat:
After my styling tweaks it looks like this:
image

Ryuta1005:
Thank you so much!  
It’s an honor to be included.

Please credit me as: Ryuta Otsuka.

nedbat:
This is now released as part of [coverage 7.12.0](https://pypi.org/project/coverage/7.12.0).

PR #2092, opened 2025-11-20
Improve partial branch documentation: add missing branch examples

### LINKED ISSUE #1597: Add explanation and examples for symbols used when displaying branch coverage
Yesterday, CI was broken as one branch was not covered, see

```
plugins.py                       210      0     50      1    99%   303->302
```

Screenshot from 2023-03-30 19-41-17

While the arrow symbols for if/else branches are more or less self explanatory, here I did not understand what it means.

I then had a look at
- https://coverage.readthedocs.io/en/7.2.2/branch.html
- https://coverage.readthedocs.io/en/7.2.2/cmd.html#coverage-summary-coverage-report
- https://coverage.readthedocs.io/en/7.2.2/faq.html
but did not find any explanation. I then reached out to a online chat, where I was helped: The for-loop is not exhausted

I'd like to have examples for both if/else and for-loops (and maybe other special things), and explanations for those examples, ideally at the branch coverage page, and maybe a link from the `cmd.html` page, where the branch coverage example is shown, a la "more examples and explanations at ...".

Thank you!

Ayisha-Mohammed:
This PR improves the partial branch coverage documentation by adding the missing examples.
Fixes #1597


nedbat:
Thanks for getting this started!

I think all the examples should be put in branch.rst, since they will be helpful for people looking at the HTML report also. You can put something in cmd_report.rst that explains what `->` means, and reference the examples in branch.rst for more details.

For the formatting, try using `make dochtml` on your machine to see the finished HTML files. This can help you get the RST markup correct.

Ayisha-Mohammed:
@nedbat  made changes as requested.

doc/branch.rst
```diff
@@ -41,33 +40,30 @@ To measure branch coverage, run coverage.py with the ``--branch`` flag::
...
 the percentage of branch possibilities taken will be included in the percentage
-covered total for each file.  The coverage percentage for a file is the actual
-executions divided by the execution opportunities.  Each line in the file is an
-execution opportunity, as is each branch destination.
+covered total for each file. Each line in the file is an execution opportunity,
+as is each branch destination.
 
 The HTML report gives information about which lines had missing branches. Lines
-that were missing some branches are shown in yellow, with an annotation at the
-far right showing branch destination line numbers that were not exercised.
```
nedbat, line 50: I'm not sure why you removed "at the far right".  I thought people might not notice it at the edge that usually has no information, so I wanted to direct their attention there.

doc/branch.rst
```diff
@@ -41,33 +40,30 @@ To measure branch coverage, run coverage.py with the ``--branch`` flag::
...
+that were missing some branches are shown in yellow, with an annotation showing
+branch destination line numbers that were not exercised.
 
 The XML and JSON reports produced by ``coverage xml`` and ``coverage json``
 also include branch information, including separate statement and branch
 coverage percentages.
 
-
```
nedbat, line 56: I like having extra space before headings. Can we keep the double-blank lines?

doc/branch.rst
```diff
@@ -85,47 +81,139 @@ as a branch if one of its choices is excluded::
...
         if eventually():
             break
 
-Here the while loop will never complete because the break will always be taken
-at some point.  Coverage.py can't work that out on its own, but the "no branch"
-pragma indicates that the branch is known to be partial, and the line is not
-flagged.
-
 Generator expressions
-=====================
```
nedbat, line 121: This was meant to be a section within "Structurally partial branches", which is why it has a different header rule style.

doc/branch.rst
```diff
@@ -41,33 +40,30 @@ To measure branch coverage, run coverage.py with the ``--branch`` flag::
 
 When you report on the results with ``coverage report`` or ``coverage html``,
 the percentage of branch possibilities taken will be included in the percentage
-covered total for each file.  The coverage percentage for a file is the actual
-executions divided by the execution opportunities.  Each line in the file is an
-execution opportunity, as is each branch destination.
+covered total for each file. Each line in the file is an execution opportunity,
+as is each branch destination.
```
nedbat: The last sentence mentions "execution opportunity", which was part of the sentence you dropped.  I'm not sure the last sentence makes sense without it.

doc/branch.rst
```diff
@@ -85,47 +81,139 @@ as a branch if one of its choices is excluded::
...
             break
         do_something()
 
-Here the while loop will never exit normally, so it doesn't take both of its
-"possible" branches.  For some of these constructs, such as "while True:" and
-"if 0:", coverage.py understands what is going on.  In these cases, the line
-will not be marked as a partial branch.
+Here the ``while`` loop will never exit normally, so it doesn't take both of
+its "possible" branches. For some of these constructs, such as ``while True:``
+and ``if 0:``, coverage.py understands what is going on.
```
nedbat: Again, I feel like something important is lost by dropping the last sentence of the paragraph.

doc/branch.rst
```diff
@@ -85,47 +81,139 @@ as a branch if one of its choices is excluded::
...
 
     i = 0
     while i < 999999999:  # pragma: no branch
         if eventually():
             break
 
-Here the while loop will never complete because the break will always be taken
-at some point.  Coverage.py can't work that out on its own, but the "no branch"
-pragma indicates that the branch is known to be partial, and the line is not
-flagged.
```
nedbat, line 118: You didn't like this paragraph?

doc/branch.rst
```diff
@@ -85,47 +81,139 @@ as a branch if one of its choices is excluded::
...
+---------------------
+
+Generator expressions may also report partial branch coverage. Example::
+
+    value = next(i for i in range(1))
+
+The generator does not iterate until ``StopIteration`` is raised. This is
+another case where adding ``# pragma: no branch`` may be desirable.
+
+Missing branches in branch coverage
```
nedbat: This section is what #1597 was asking for: explanations and examples.  So let's call it that.

doc/branch.rst
```diff
@@ -85,47 +81,139 @@ as a branch if one of its choices is excluded::
...
+-----------------------------------
+
```
nedbat, line 136: We need a paragraph here about the section.  "Here are some examples of missing branches with explanations of how they are reported," or something like that.

doc/branch.rst
```diff
@@ -85,47 +81,139 @@ as a branch if one of its choices is excluded::
...
+    Missing:
+        12 -> 10
+
+Coverage.py will report::
+
+    12->10
+
+Below are examples showing how common control-flow structures appear in the
+``Missing`` column when branch coverage is enabled.
```
nedbat, line 174: This is the paragraph that could be at the top of the section. Nice.

doc/branch.rst
```diff
@@ -85,47 +81,139 @@ as a branch if one of its choices is excluded::
...
+another case where adding ``# pragma: no branch`` may be desirable.
+
+Missing branches in branch coverage
+-----------------------------------
+
+For example::
+
+    10: for x in items:
+    11:     print(x)
+    12:     # end of loop body
```
nedbat: 12 isn't an executable line.  Change the comment to a real statement, it could just be another print().

doc/branch.rst
```diff
@@ -85,47 +81,139 @@ as a branch if one of its choices is excluded::
...
+    Missing:
+        10 -> 11
+        12 -> 10
+
+Case 2 — ``items = [1]``::
+
+    Missing:
+        12 -> 10
+
+Coverage.py will report::
```
nedbat: We can probably skip this, since you've already said "12 -> 10" just above.

doc/branch.rst
```diff
@@ -85,47 +81,139 @@ as a branch if one of its choices is excluded::
...
+Example::
+
+    10: if flag:
+    11:     do_true()
+    12: else:
+    13:     do_false()
+
+If ``flag`` is always true, the false branch is never taken.
+
+The report will show::
```
nedbat: we should include the word "missing" here somehow to underscore that the branches reported are the missing branches.

doc/branch.rst
```diff
@@ -85,47 +81,139 @@ as a branch if one of its choices is excluded::
...
+
+Example::
+
+    20: for x in items:
+    21:     print(x)
+
+A ``for`` loop has:
+
+* ``20->21`` (enter body)
+* ``21->20`` (repeat)
```
nedbat: You used bullets here, but not above. We should choose a consistent style.

doc/commands/cmd_report.rst
```diff
@@ -104,6 +104,19 @@ inclusive are missing coverage.  Missed branches are shown with an arrow:
 backwards in a file, so you might see a branch from a later line to an earlier
 line, like "55->50".
 
+
+Missing branches
+----------------
+
+Missing branches are shown such as ``12->10``.why is it backward?
```
nedbat: This make it sound like branches are always reported in reverse. We should be clear that it's only sometimes.

doc/commands/cmd_report.rst
```diff
@@ -104,6 +104,19 @@ inclusive are missing coverage.  Missed branches are shown with an arrow:
...
+This means that execution never moved from line 12 back to line 10. Backward
+branches often come from loops: if a loop never repeats (or never runs),
+the backward branch is missing.
+
+For detailed examples of how branch coverage works in loops and other control
+flow constructs (``if/else``, ``for`` loops, ``while`` loops, ``try/except``),
+see :doc:`branch </branch>`.
```
nedbat: This reads as "see branch." which isn't the form we want.  Use:
```
see :ref:`branch`.
```

nedbat (changes requested):
This is great. You've removed more text than I wanted you to, so we need to find the right balance.

doc/branch.rst
Ayisha-Mohammed: you mean to remove the case2 block?

Ayisha-Mohammed:
@nedbat  i've made the changes you requested. Thank you for the guidance , learnt a lot though this process.

doc/branch.rst
```diff
@@ -129,3 +127,112 @@ While we might expect this line of code to be reported as covered, the
...
+Case 1 — ``items`` is empty::
+
+    Missing:
+        10 -> 11
+        12 -> 10
+
+Case 2 — ``items = [1]``::
+
+    Missing:
+        12 -> 10
```
nedbat: I'm sorry i didn't realize this earlier: this code doesn't have a branch from 12->10, because branches only occur where a statement can go next to two different places.  Line 12 always goes to line 10, even on the last iteration of the loop.  So we need to re-think this example.

doc/branch.rst
nedbat: I meant that we show "12->10" here on line 164 but we already showed it on line 160.  But this whole example needs to be re-thought.

doc/branch.rst
nedbat: It's important to test the example to see that coverage.py actually reports the missing branches we're claiming.

Ayisha-Mohammed: @nedbat thank you , i  fixed it

doc/branch.rst
```diff
@@ -129,3 +127,102 @@ While we might expect this line of code to be reported as covered, the
...
+Case 1 — ``items`` is empty::
+
+    Missing:
+    2 -> 3
+    4 -> 2
+
+Case 2 — ``items = [1]``::
+
+    Missing:
+    4 -> 2
```
nedbat: This case is correct.

doc/branch.rst
```diff
@@ -129,3 +127,102 @@ While we might expect this line of code to be reported as covered, the
...
+* ``2 -> 6`` (exit)
+* ``4 -> 5`` (if True)
+* ``4 -> 2`` (if False, backward branch)
+
+
+Case 1 — ``items`` is empty::
+
+    Missing:
+    2 -> 3
+    4 -> 2
```
nedbat: You have to actually run these examples to see what coverage produces:
```
% cat empty.py
items = []
for x in items:
    print(x)
    if x:
        print("x is true")
print("done")

% coverage run --branch empty.py
done

% coverage report -m
Name       Stmts   Miss Branch BrPart  Cover   Missing
------------------------------------------------------
empty.py       6      3      4      1    40%   3-5
------------------------------------------------------
TOTAL          6      3      4      1    40%
```
Here the report only says `3-5` because none of those lines are run at all. Branches that are missing because the target line isn't run aren't reported as missing because that would create too much noise.

Ayisha-Mohammed: i understand , coverage doesn't report any missing branches . because when item is empty ,(enter loop) never executes hence branch lines never executes.

doc/branch.rst
```diff
@@ -129,3 +127,108 @@ While we might expect this line of code to be reported as covered, the
...
+* ``30->34`` (exit)
+* ``32->33`` (if True)
+* ``32->30`` (if False)
+
+The false branch of the ``if`` statement is
+never taken, so coverage reports the missing branch::
+
+    32->30
+
+``try / except`` example
```
nedbat, line 218: This example has no branches.  Running the code would show what coverage would actually report.

nedbat:
I'm going to merge this and make some small edits. Thanks for persisting.

Ayisha-Mohammed:
I'm glad and i learned a lot , thank you very much @nedbat

nedbat:
This is now released as part of [coverage 7.13.1](https://pypi.org/project/coverage/7.13.1).

PR #2100, opened 2025-12-03
fix: put the pth file in the correct location

henryiii:
Running:

```
uv build
unzip -l dist/*.whl
```

Shows this now going into the correct location in the wheel, alongside coverage.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,6 @@
 import re
 import os
 import os.path
-import site
 import sys
 
 from setuptools import Extension, errors, setup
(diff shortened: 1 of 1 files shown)


nedbat:
Thanks! It didn't occur to me that I could use ".." in a file path there, but it seems to work great.  I've added this to the much larger re-working of setup.py that I'm doing on a branch now.

nedbat:
This is fixed in commit 3fd5727f21b32a11927debc11d8d3b4177f164a3

nedbat:

PR #2101, opened 2025-12-03
chore: incorrect classifier usage

henryiii:
The Development Status classifier is _not_ a proxy for the version status. It's an indicator for how stable the _project as a whole_ is. That's why the don't match dev/alpha/beta/rc/final/post, but instead have things like 6 - Mature and 7 - Inactive. Coverage is a 5 - Production / Stable. It is, as a project, Production / Stable even in development and beta releases. The dev release _version_ tells you it's a dev release (of a stable project), not the classifier.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,7 @@ from setuptools.command.build_ext import build_ext  # pylint: disable=wrong-impo
 
 # PYVERSIONS
 classifiers = """\
+Development Status :: 5 - Production/Stable
 Environment :: Console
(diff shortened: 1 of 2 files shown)


nedbat:
Thanks for pointing this out. I've added this to the much larger re-working of setup.py that I'm doing on a branch now.

nedbat:
This is fixed in commit da833a18b7670c2df84ab437b72560b627610063

nedbat:

PR #2104, opened 2025-12-08
fix: skip empty coverage data in CoverageData.update

aeeeeeep:
### Problem Description
In distributed training scenarios using `torchrun` with custom coverage plugins, we encounter empty `other_data` during `coverage combine`. Specifically, when the `tracer`, `arc`, and `line_bits` tables in the coverage database contain zero records, this leads to a critical error:

```python
Conflicting file tracer name for 'xxx': '' vs 'coverage_plugin'
```

### Configuration Context
```ini
[run]
parallel = True
plugins =
    coverage_plugin
...
```

### Root Cause
- **Distributed Process Behavior**: Multiple processes in `torchrun` may generate coverage data with different plugin configurations
- **Empty Coverage Data**: Some processes produce coverage databases with no actual measurements (all three key tables empty)

### Solution
Added a pre-check in `CoverageData.update()` to skip empty coverage data by querying record counts from `tracer`, `arc`, and `line_bits` tables, and skipping update operations when all three tables contain zero records.

### DIFF
--- a/coverage/sqldata.py
+++ b/coverage/sqldata.py
@@ -720,9 +720,23 @@ class CoverageData:
         self._start_using()
         other_data.read()
 
-        # Ensure other_data has a properly initialized database
-        with other_data._connect():
-            pass
+        # Check if the coverage data contains any actual measurements
+        should_skip = False
+        with other_data._connect() as other_con:
+            try:
+                with other_con.execute("SELECT COUNT(*) FROM tracer") as cur:
+                    tracer_count = cur.fetchone()[0]
+                with other_con.execute("SELECT COUNT(*) FROM arc") as cur:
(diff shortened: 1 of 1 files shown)


nedbat:
My reaction is, "I guess it's fine to skip empty data files" but also, "it seems wrong that the data file is empty."  Can you give me a way to reproduce this? You mentioned custom plugins, are those coverage plugins? Can you share them?

aeeeeeep:
Please wait a while, this scene is very complicated.

aeeeeeep:
The root cause is that the non-standard practice of monkey patching interferes with correct path resolution. I resolved this by filtering out the monkey-patched code within the custom plugin, and the issue no longer persists.

PR #2120, opened 2026-01-22
Fix PathAliases auto-generated rules matching nested directories incorrectly

### LINKED ISSUE #2072: `coverage combine` with `relative_files=true` collapses nested directories with duplicate names, causing data loss
# Title

`coverage combine` with `relative_files=true` collapses nested directories with duplicate names, causing data loss

---

# Description

## Summary

When using `relative_files=true`, `coverage combine` generates overly broad path mapping rules that collapse files from nested directories sharing the same name. For example, both `features/templates.py` and `pkg/features/templates.py` map to `features/templates.py`, causing coverage data loss.

## Expected vs Actual

**Expected:** Both files tracked separately
- `features/templates.py` → `features/templates.py`
- `pkg/features/templates.py` → `pkg/features/templates.py`

**Actual:** Nested file collapses into root-level file
- `features/templates.py` → `features/templates.py` ✓
- `pkg/features/templates.py` → `features/templates.py` ❌ (data lost)

## Reproducer

Minimal reproduction: https://github.com/dephiros/pytest-coverage-path-bug

```bash
# Clone and setup
git clone https://github.com/dephiros/pytest-coverage-path-bug.git
cd pytest-coverage-path-bug
uv sync

# Run workflow
uv run coverage erase
COVERAGE_DEBUG=pathmap uv run coverage run --parallel-mode -m pytest
COVERAGE_DEBUG=pathmap uv run coverage combine
uv run coverage report
```

**Result:** Report shows 44 statements instead of 48. `pkg/features/templates.py` is missing.

**Without combine (for comparison):**
```bash
uv run coverage erase
uv run coverage run -m pytest  # No --parallel-mode
uv run coverage report
```
**Result:** Report correctly shows 48 statements with all files.

## Debug Output

`coverage combine` shows the collision:

```
Generating rule: '*/features' -> 'features/' using regex '^(.*[\\\\/])?features[\\\\/]'
Matched path 'features/templates.py' to rule '*/features' -> 'features/', producing 'features/templates.py'
Matched path 'pkg/features/templates.py' to rule '*/features' -> 'features/', producing 'features/templates.py'
```

The `*/features` rule matches any path ending with `/features/`, causing the collision.

## Versions

- Python: 3.11.10
- coverage.py: 7.11.0
- pytest: 8.4.2
- pytest-cov: 7.0.0

## Config

```toml
[tool.coverage.run]
relative_files = true
branch = true
source = ["."]
```

## Repository Structure

```
.
├── features/
│   └── templates.py          # 4 statements
├── pkg/
│   └── features/
│       └── templates.py      # 4 statements - GETS COLLAPSED
└── pyproject.toml
```

## Related Issues

- pytest-dev/pytest-cov#724 - Initially filed against pytest-cov, but investigation revealed the bug is in `coverage combine` itself

dephiros:
## Summary

When using `relative_files=true`, the auto-generated path mapping rules used a regex pattern `^(.*[\\/])?{dir1}[\\/]` that could match directory names anywhere in a path, not just from the start. This caused paths like `pkg/features/templates.py` to incorrectly match a rule generated for `features/templates.py`, resulting in lost coverage data during combine operations.

## The Problem

For example, with two files:
- `features/templates.py`
- `pkg/features/templates.py`

The auto-generated rule for `features/` would use regex `^(.*[\\/])?features[\\/]`, which incorrectly matches `pkg/features/templates.py` because `(.*[\\/])?` allows any prefix.

## The Fix

Changed the auto-generated regex from:
```
^(.*[\\/])?{dir1}[\\/]  (matches anywhere)
```
to:
```
^{dir1}[\\/]            (anchored to start)
```

This ensures that nested directories with the same name as a top-level directory are not incorrectly collapsed during coverage combine.

## Testing

- Added a new test `test_nested_directories_same_name` that verifies the fix
- All existing PathAliases tests pass (45 tests)
- All test_files.py tests pass (205 tests)

Fixes #2072

### DIFF
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -499,12 +499,16 @@ class PathAliases:
             path = relative_filename(path)
 
         if self.relative and not isabs_anywhere(path):
-            # Auto-generate a pattern to implicitly match relative files
+            # Auto-generate a pattern to implicitly match relative files.
+            # Use an anchored pattern that matches from the start of the path,
+            # rather than a wildcard pattern that could match anywhere.
+            # This prevents "pkg/features/x.py" from incorrectly matching
+            # a pattern generated for "features/x.py". See issue #2072.
             parts = re.split(r"[/\\]", path)
             if len(parts) > 1:
                 dir1 = parts[0]
-                pattern = f"*/{dir1}"
-                regex_pat = rf"^(.*[\\/])?{re.escape(dir1)}[\\/]"
+                pattern = f"{dir1}/"
+                regex_pat = rf"^{re.escape(dir1)}[\\/]"
                 result = f"{dir1}{os.sep}"
                 # Only add a new pattern if we don't already have this pattern.
                 if not any(p == pattern for p, _, _ in self.aliases):
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -531,12 +531,16 @@ class PathAliases:
             path = relative_filename(path)
 
         if self.relative and not isabs_anywhere(path):
-            # Auto-generate a pattern to implicitly match relative files
+            # Auto-generate a pattern to implicitly match relative files.
+            # Use an anchored pattern that matches from the start of the path,
+            # rather than a wildcard pattern that could match anywhere.
+            # This prevents "pkg/features/x.py" from incorrectly matching
+            # a pattern generated for "features/x.py". See issue #2072.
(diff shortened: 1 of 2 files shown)


dephiros:
@nedbat , I know you are busy but want to check in with you for a review 🙏

PR #2123, opened 2026-01-27
fix: sysmon coverage for nested context managers

### LINKED ISSUE #1987: 3.14: branch coverage incorrect with triple-nested context manager
**Describe the bug**

for some reason a triple-nested context manager seems to confuse branch coverage in python 3.14

**To Reproduce**

minimal reproduction:

```python
import contextlib


@contextlib.contextmanager
def ctx():
    yield



def f():
    with ctx():
        with ctx():
            with ctx():
                print(1)

f()
```

with 3.14:

```console
$ bash -xc 'python --version --version && coverage --version && coverage erase && coverage run --branch -m t && coverage report --show-missing'
+ python --version --version
Python 3.14.0b3 (main, Jun 18 2025, 08:59:11) [GCC 13.3.0]
+ coverage --version
Coverage.py, version 7.9.2a0.dev1 with C extension
Full documentation is at https://coverage.readthedocs.io/en/7.9.2a0.dev1
+ coverage erase
+ coverage run --branch -m t
1
+ coverage report --show-missing
Name    Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------
t.py       10      0      2      2    83%   13->14, 13->exit
---------------------------------------------------
TOTAL      10      0      2      2    83%
```

**Expected behavior**

3.13 behaviour:

```console
$ bash -xc 'python --version --version && coverage --version && coverage erase && coverage run --branch -m t && coverage report --show-missing'
+ python --version --version
Python 3.13.5 (main, Jun 12 2025, 08:56:14) [GCC 13.3.0]
+ coverage --version
Coverage.py, version 7.9.2a0.dev1 with C extension
Full documentation is at https://coverage.readthedocs.io/en/7.9.2a0.dev1
+ coverage erase
+ coverage run --branch -m t
1
+ coverage report --show-missing
Name    Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------
t.py       10      0      2      0   100%
---------------------------------------------------
TOTAL      10      0      2      0   100%
```

**Additional context**

this was tickled by flake8's testsuite -- it works fine with two contexts for some reason though 🤷

miketheman:
Resolves #1987 

- Extract helper for frame references
- Track last lines during sysmon branch coverage to catch them all, even during exceptions

### DIFF
--- a/coverage/bytecode.py
+++ b/coverage/bytecode.py
@@ -253,3 +253,30 @@ def always_jumps(code: CodeType) -> dict[TOffset, TOffset]:
         elif inst.opcode in NOPS:
             jumps[inst.offset] = inst.offset + 2
     return jumps
+
+
+def multi_visit_lines(code: CodeType) -> set[TLineNo]:
+    """Find lines that appear at non-contiguous bytecode ranges.
+
+    These are lines that can be executed multiple times in a single function
+    call, such as `with` statement lines (executed on entry and cleanup) or
+    lines duplicated for exception handling. We need to keep receiving LINE
+    events for these lines to track arc transitions correctly.
+    """
+    line_offsets: dict[TLineNo, list[TOffset]] = collections.defaultdict(list)
+    iwalker = InstructionWalker(code)
+    for inst in iwalker.walk(follow_jumps=False):
+        if inst.line_number:
+            line_offsets[inst.line_number].append(inst.offset)
+
+    result: set[TLineNo] = set()
+    for line, offsets in line_offsets.items():
+        # offsets are already in order from walk()
+        for i in range(1, len(offsets)):
+            # Bytecode instructions are 2 bytes apart; a larger gap means
+            # the line appears in multiple non-contiguous code sections.
+            if offsets[i] - offsets[i - 1] > 2:
+                result.add(line)
+                break
+
+    return result
--- a/coverage/sysmon.py
+++ b/coverage/sysmon.py
@@ -15,7 +15,7 @@ import threading
 import traceback
 from collections.abc import Callable
 from dataclasses import dataclass
-from types import CodeType
+from types import CodeType, FrameType
 from typing import Any, NewType, Optional, cast
 
 from coverage import env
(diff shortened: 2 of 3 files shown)


nedbat:
Can you tell me more about the theory of this fix? I was surprised to see UNWIND events handled, since this problem didn't involve exceptions. Indeed, deleting that addition didn't change the outcome. We don't need the UNWIND handling.

What does last_lines do for us here and how does it fix the measurement of with statements?

miketheman:
> Can you tell me more about the theory of this fix? I was surprised to see UNWIND events handled, since this problem didn't involve exceptions. Indeed, deleting that addition didn't change the outcome. We don't need the UNWIND handling.

I can back that out, however my understanding is if there's an observed exception during the test suite run, memory will bloat since `last_lines` won't be "emptied" as a result.

> What does last_lines do for us here and how does it fix the measurement of with statements?

It maps each active function call (by frame id) to the last line we saw in it.
The same function can be called multiple times, and each call needs its own tracking.

Why do we need to track stuff? When we see a LINE event, remember it. When we see the next LINE event, record the transition.

An example:

  LINE event: line 5  →  remember "last line was 5"
  LINE event: line 6  →  record arc (5→6), remember "last line was 6"
  LINE event: line 7  →  record arc (6→7), remember "last line was 7"

Previously, sysmon would say "I've seen line 5, don't tell me about it again." Now we need to hear about every line execution to track transitions.

nedbat:
Oh, I hadn't noticed this yet:
```py
        # Don't return DISABLE: keep getting LINE events for arc tracking.
        return None
```
This undoes the entire benefit of sys.monitoring.  If we're going to change that we might as well use the C tracer.

miketheman:
> Oh, I hadn't noticed this yet:
> 
> ```python
>         # Don't return DISABLE: keep getting LINE events for arc tracking.
>         return None
> ```
> 
> This undoes the entire benefit of sys.monitoring. If we're going to change that we might as well use the C tracer.

Very good point - shows how much I know about sysmon! 🤦 

I took another pass at this, and added a helper method that checks the bytecode and makes a decision based on that, so the sysmon performance would be preserved for most code except those using these deeply-nested context managers.

Let me know what you think!

coverage/bytecode.py
```diff
@@ -253,3 +253,30 @@ def always_jumps(code: CodeType) -> dict[TOffset, TOffset]:
         elif inst.opcode in NOPS:
             jumps[inst.offset] = inst.offset + 2
     return jumps
+
+
+def multi_visit_lines(code: CodeType) -> set[TLineNo]:
```
nedbat, line 258: This is interesting. Can you show the dis output of the with statements to help me understand what this function is identifying?

miketheman: I tried to get some information this way, by replacing the behavior from `InstructionWalker` with inline calls to `dis` - I'm not 100% certain it's right, but maybe it helps? https://gist.github.com/miketheman/0ff15be2efea2573ae46e52bef069c77#file-debug1_output-txt

PR #2130, opened 2026-02-07
fix: set fixed paths_list in TreeMatcher init

colorlace:
This lil fix simply ensures we don't get hit with `RuntimeError: Set changed size during iteration`  if there's any concurrent update to `sys.paths` while `TreeMatcher.__init__` runs.

My CI pipeline has encountered this error:
```
.venv/lib/python3.11/site-packages/coverage/inorout.py:354: in should_trace
    self.set_matchers_depending_on_syspath()
.venv/lib/python3.11/site-packages/coverage/inorout.py:308: in set_matchers_depending_on_syspath 
    self.third_match = TreeMatcher(self.third_paths, "third", "Third-party lib", self._debug)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/coverage/files.py:266: in __init_
    for p in paths: 
E     RuntimeError: Set changed size during iteration
```

This seems like a very similar situation to https://github.com/coveragepy/coveragepy/issues/1733
which was fixed by something similar:  https://github.com/coveragepy/coveragepy/commit/08fc997b6b3d881e8e5082b71ad6d6bcc4e2f752.

### DIFF
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -260,10 +260,11 @@ class TreeMatcher(Matcher):
         caption: str = "",
         debug: DebugFn = None,
     ) -> None:
-        self.original_paths = human_sorted(paths)
+        paths_list = list(paths)
+        self.original_paths = human_sorted(paths_list)
         super().__init__(self.original_paths, name=name, caption=caption, debug=debug)
         self.paths = []
-        for p in paths:
+        for p in paths_list:
             ap = abs_file(p)
             if ap != p and debug:
                 debug(f"        Normalized {p!r} to {ap!r}")


nedbat:
Thanks for this.  Can you share the project that caused the error? I'd like to understand better where the concurrent access is coming from.

colorlace:
> Thanks for this. Can you share the project that caused the error? I'd like to understand better where the concurrent access is coming from.

Unfortunately it's a closed-source company project. But I believe it occurs because, during testing, we run  
```
from pyspark.sql import SparkSession
```
 which updates the `sys.path`.  I suppose this happens concurrently while `coveragepy` is doin its thang.

nedbat:
It looks like a good fix, thanks.

colorlace:
Thanks Ned you're a legend! (Unicode sandwich saved me ten years ago when I started coding)

nedbat:
Wow, Unicode sandwich, a historical deep cut! :D

colorlace:
Haha I did a double take when I saw your name on the contributors list. Definitely uncovered some old memories of string processing pain (and retribution) XD

nedbat:
This is now released as part of [coverage 7.13.4](https://pypi.org/project/coverage/7.13.4).

PR #2135, opened 2026-02-18
Allow excluding region reporting in all report types that support it

astei:
This PR adds a new flag, `--omit-regions`, to the JSON and HTML reports.

I use coverage.py on a particularly large monolithic application (several millions of lines of code), and we have custom tooling that parses JSON reports created by coverage.py. When region-level reporting was added to JSON reports, this tooling had a notable regression in memory usage and performance. We don't use this information, so it's helpful if coverage.py doesn't emit it in the first place.

### DIFF
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -190,18 +190,31 @@ class JsonReportTest(UsingModulesMixin, CoverageTest):
         expected_result: dict[str, Any],
         mod_name: str,
         source_code: str,
+        omit_regions: bool = False,
     ) -> None:
         """
         Helper that creates a file and compares its JSON report to expected results.
         """
         self.make_file(f"{mod_name}.py", source_code)
-        self._compare_json_reports(cov, expected_result, mod_name)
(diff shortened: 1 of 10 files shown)


tests/test_json.py
```diff
@@ -190,18 +190,31 @@ def _assert_json_report(
...
     ) -> None:
         """
         Helper that creates a file and compares its JSON report to expected results.
         """
         self.make_file(f"{mod_name}.py", source_code)
-        self._compare_json_reports(cov, expected_result, mod_name)
+        self._compare_json_reports(cov, expected_result, mod_name, omit_regions)
+
+    @staticmethod
+    def _approx_floats(obj: Any) -> Any:
```
devdanzin, line 202: This seems unrelated to your goals, does it become necessary due to the other changes?

astei: Yes, the tests did not pass because of float representation being inexact, despite not changing the code that was being covered.

Issue #2141, opened 2026-03-07
Filenames containing pipe (|) character break report markdown table


ellieayla:
**Describe the bug**

The markdown report doesn't fully escape filenames.

In markdown files, some characters are meaningful. In the markdown report, some characters in filenames are escaped. For example, underscores `_` are super common in filenames and escaped by the markdown report generator, since they would start/stop an underline.

But a markdown report does _not_ escape pipe `|` characters names in filenames. Pipe characters are used to make tables.

A markdown report with pipe characters in filenames is broken.

Yes, pipe characters in filenames is kinda silly.

**To Reproduce**

How can we reproduce the problem?  Please *be specific*. Don't link to a failing CI job. Think about the time it will take us to recreate your situation: the easier you make it, the more likely your issue will be addressed.

1. What version of Python are you using?: `cpython 3.14.2 / macOS-15.6.1-arm64-arm-64bit-Mach-O`
1. What version of coverage.py shows the problem? `Coverage.py, version 7.13.4 with C extension`
1. What versions of what packages do you have installed? The output of `pip freeze` is helpful. Nothing?
1. What code shows the problem?  Give us a *specific commit* of a *specific repo* that we can check out. If you've already worked around the problem, please provide a commit before that fix. See below.
1. What commands should we run to reproduce the problem? *Be specific*. Include everything, even `git clone`, `pip install`, and so on. Explain like we're five!

```sh
touch 'pipe|.py'
uv tool run coverage erase
uv tool run coverage run 'pipe|.py'
uv tool run coverage report --format=markdown
| Name      |    Stmts |     Miss |    Cover |   Missing |
|---------- | -------: | -------: | -------: | --------: |
| pipe|.py  |        0 |        0 |     100% |           |
| **TOTAL** |    **0** |    **0** | **100%** |           |
```

Which renders in GFM as:

| Name      |    Stmts |     Miss |    Cover |   Missing |
|---------- | -------: | -------: | -------: | --------: |
| pipe|.py  |        0 |        0 |     100% |           |
| **TOTAL** |    **0** |    **0** | **100%** |           |

Note the filename was split into a new column.

**Alternatively, a testcase for `tests/test_report.py`:**

```py
    @pytest.mark.skipif(env.WINDOWS, reason="No pipe characters in filenames on Windows")
    def test_markdown_escape_filename(self) -> None:
        self.make_file("subdir/so_me|file.py", "print(1)")
[...]

PR #2142, opened 2026-03-07
fix: escape filename in markdown report #2141

ellieayla:
Both underscore and pipe characters are valid in filenames and meaningful in markdown tables. Underscores were already escaped by the markdown report, using a simple string replacement, but pipe characters were not.

```sh
touch 'pipe|.py'
uv tool run dist/coverage-7.13.5a0.dev1-cp314-cp314-macosx_11_0_arm64.whl erase
uv tool run dist/coverage-7.13.5a0.dev1-cp314-cp314-macosx_11_0_arm64.whl run 'pipe|.py'
uv tool run dist/coverage-7.13.5a0.dev1-cp314-cp314-macosx_11_0_arm64.whl report --format=markdown
| Name      |    Stmts |     Miss |    Cover |
|---------- | -------: | -------: | -------: |
| pipe\|.py |        0 |        0 |     100% |
| **TOTAL** |    **0** |    **0** | **100%** |
```

aka

| Name      |    Stmts |     Miss |    Cover |
|---------- | -------: | -------: | -------: |
| pipe\|.py |        0 |        0 |     100% |
| **TOTAL** |    **0** |    **0** | **100%** |

### DIFF
--- a/coverage/report.py
+++ b/coverage/report.py
@@ -126,8 +126,13 @@ class SummaryReporter:
         `end_lines` is a list of ending lines with information about skipped files.
 
         """
+
+        def escape_markdown(input_value: str) -> str:
+            """Crudely replace characters meaningful in markdown tables."""
+            return input_value.replace("_", "\\_").replace("|", "\\|")
+
         # Prepare the formatting strings, header, and column sorting.
-        max_name = max((len(line[0].replace("_", "\\_")) for line in lines_values), default=0)
+        max_name = max((len(escape_markdown(line[0])) for line in lines_values), default=0)
         max_name = max(max_name, len("**TOTAL**")) + 1
         formats = dict(
             Name="| {:{name_len}}|",
@@ -160,7 +165,7 @@ class SummaryReporter:
             self.write_items(
                 (
                     formats[item].format(
-                        str(value).replace("_", "\\_"), name_len=max_name, n=max_n - 1
+                        escape_markdown(str(value)), name_len=max_name, n=max_n - 1
                     )
                     for item, value in zip(header, values)
                 )
--- a/coverage/report.py
+++ b/coverage/report.py
@@ -25,9 +25,9 @@ ESCAPE_PUNCTUATION_TABLE = "".maketrans(
 )
 
 
-def escape_markdown(input_value: str) -> str:
+def escape_markdown(text: str) -> str:
     """Prefix all characters meaningful in markdown tables with backslashes."""
-    return input_value.translate(ESCAPE_PUNCTUATION_TABLE)
+    return text.translate(ESCAPE_PUNCTUATION_TABLE)
 
 
 class SummaryReporter:
--- a/coverage/report.py
+++ b/coverage/report.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import sys
 from collections.abc import Iterable
+from string import punctuation
 from typing import IO, TYPE_CHECKING, Any
 
 from coverage.exceptions import ConfigError, NoDataError
@@ -18,6 +19,15 @@ from coverage.types import TMorfs
 
 if TYPE_CHECKING:
     from coverage import Coverage
+
+ESCAPE_PUNCTUATION_TABLE = "".maketrans(
+    {char: f"\\{char}" for char in punctuation if char not in ".,/-"}
+)
+
+
+def escape_markdown(input_value: str) -> str:
+    """Prefix all characters meaningful in markdown tables with backslashes."""
+    return input_value.translate(ESCAPE_PUNCTUATION_TABLE)
 
 
 class SummaryReporter:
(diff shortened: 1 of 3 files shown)


ellieayla:
`ruff format --check` was mad about 'single quotes' and a wrapped function definition. Force-pushed with its recommendations included.

nedbat:
Thanks! I think we need to escape more than just underscore and pipe.  I'm surprised this hasn't come up already: don't we need to escape backslashes? If a file name starts with pipe, on Windows, it would be `foo\|pipe.py` which would be escaped as `foo\\|pipe.py`, the two backslashes become one, and the pipe isn't actually escaped. File names can also contain stars, or be valid markdown urls: `hello[x](y).py`. Those characters need escaping.

Since Commonmark says all ASCII punctuation can be escaped, we can do all of `string.punctuation`.  That will require adjusting the tests.  The `get_report()` helper function already knows the output_format, so it can do backslash-aware converting of backslashes to forward slashes for the OS-insensitivity we need in the tests.

coverage/report.py
```diff
@@ -126,8 +126,13 @@ def report_markdown(
         `end_lines` is a list of ending lines with information about skipped files.
 
         """
+
+        def escape_markdown(input_value: str) -> str:
```
nedbat: I'd prefer a function like this to be at the module top level rather than an inner function.

nedbat (changes requested):

ellieayla:
If the whole of string.punctuation is needed, then a chain of s.replace() becomes super unwieldy.

Instead, how about str.maketrans + str.translate?

edit: 4 tests are affected. But since `.,/-` aren't meaningful, leave those characters out and avoid changing the common cases - then 0 of the existing tests fail (on mac)

```diff
diff --git a/coverage/report.py b/coverage/report.py
index 544106f3..bf6a7982 100644
--- a/coverage/report.py
+++ b/coverage/report.py
@@ -7,6 +7,7 @@
 
 import sys
 from collections.abc import Iterable
+from string import punctuation
 from typing import IO, TYPE_CHECKING, Any
 
 from coverage.exceptions import ConfigError, NoDataError
@@ -19,6 +20,13 @@
 if TYPE_CHECKING:
     from coverage import Coverage
 
+ESCAPE_PUNCTUATION_TABLE = "".maketrans(
+    {char: f"\\{char}" for char in punctuation if char not in ".,/-"}
+)
+
+
+def escape_markdown(input_value: str) -> str:
+    """Prefix all characters meaningful in markdown tables with backslashes."""
+    return input_value.translate(ESCAPE_PUNCTUATION_TABLE)
+
 
 class SummaryReporter:
     """A reporter for writing the summary report."""
@@ -127,10 +135,6 @@ def report_markdown(
 
         """
 
-        def escape_markdown(input_value: str) -> str:
-            """Crudely replace characters meaningful in markdown tables."""
-            return input_value.replace("_", "\\_").replace("|", "\\|")
-
         # Prepare the formatting strings, header, and column sorting.
         max_name = max((len(escape_markdown(line[0])) for line in lines_values), default=0)
         max_name = max(max_name, len("**TOTAL**")) + 1
```

nedbat:
TBH, this comment (not yours!) kind of horrified me:
```
# get_report() escapes backslash so we expect forward slash escaped underscore
```
That slipped past me in a previous pull request. It seems like a hacky way to deal with the windows backslashes.

I would amend the `get_report()` helper function to have an explicit `output_format` argument, and then:
```diff
-        cov.report(file=repout, **kwargs)
-        report = repout.getvalue().replace("\\", "/")
+        if output_format == "markdown":
+            report = report.replace("\\\\", r"\/")
+        else:
+            report = report.replace("\\", "/")
```

ellieayla:
Is `SummaryTest.test_empty_files` really the only test that failed on Windows at commit 5967aab? Or just the first & fast-failed?

coverage/report.py
```diff
@@ -19,6 +20,15 @@
 if TYPE_CHECKING:
     from coverage import Coverage
 
+ESCAPE_PUNCTUATION_TABLE = "".maketrans(
+    {char: f"\\{char}" for char in punctuation if char not in ".,/-"}
+)
+
+
+def escape_markdown(input_value: str) -> str:
```
nedbat: A minor point: `text` would be a better (and shorter!) name than `input_value`, which could be any argument to any function.

ellieayla:
I added a parametrized test `test_markdown_escape_filename` to cover all the punctuation, with carveouts for Windows.

I think this is ready to review.

There's several separate commits for various pieces of feedback. Do you want to squash (some of) them?

nedbat:
Thanks, this looks good. I'll squash the commits and update the changelog.

nedbat:
This is finished in commit e06eb348a071c0ebaaaf89883ecf62bba5d65f42, and tweaked just a tad in fe9d227addd9ba435e90357fd983d558b1c74f2a.

nedbat:
This is now released as part of [coverage 7.13.5](https://pypi.org/project/coverage/7.13.5).

PR #2162, opened 2026-04-25
Automatically combine coverage in report

### LINKED ISSUE #1781: Add a hint that `coverage combine` might be needed

thatch:
This does not modify the state on disk, but one common newbie error is
doing `coverage run ...` followed by `coverage report` which shows some,
but not the true coverage.

Default to doing the combine, but allow people to pass `--no-combine` if
this breaks their workflows.

Fixes #1781

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -886,10 +886,8 @@ class CoverageScript:
         # plugins may try to, for example, to read Django settings.
         sys.path.insert(0, "")
 
-        if options.no_combine:
-            self.coverage.load()
-        else:
-            self.coverage.load()
+        self.coverage.load()
+        if not options.no_combine:
             self.coverage.combine(strict=False, keep=True)
 
         total = None
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -889,7 +889,8 @@ class CoverageScript:
         if options.no_combine:
             self.coverage.load()
         else:
-            self.coverage.combine(strict=False)
+            self.coverage.load()
+            self.coverage.combine(strict=False, keep=True)
 
         total = None
         if options.action == "report":
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -320,6 +320,12 @@ class Opts:
         action="store_true",
         help="Skip files with 100% coverage.",
     )
+    no_combine = optparse.make_option(
+        "--no-combine",
+        action="store_true",
+        dest="no_combine",
+        help="Don't combine data files before reporting; use the existing data file as-is.",
+    )
     no_skip_covered = optparse.make_option(
         "--no-skip-covered",
         action="store_false",
@@ -402,6 +408,7 @@ class CoverageOptionParser(optparse.OptionParser):
             include=None,
             keep=None,
             module=None,
+            no_combine=None,
             omit=None,
             parallel_mode=None,
             precision=None,
@@ -535,6 +542,7 @@ COMMANDS = {
             Opts.datafile_input,
             Opts.ignore_errors,
             Opts.include,
+            Opts.no_combine,
             Opts.omit,
         ]
         + GLOBAL_ARGS,
@@ -607,6 +615,7 @@ COMMANDS = {
             Opts.fail_under,
             Opts.ignore_errors,
             Opts.include,
+            Opts.no_combine,
             Opts.omit,
             Opts.precision,
             Opts.quiet,
@@ -634,6 +643,7 @@ COMMANDS = {
             Opts.fail_under,
             Opts.ignore_errors,
             Opts.include,
+            Opts.no_combine,
             Opts.omit,
             Opts.output_json,
             Opts.json_pretty_print,
@@ -651,6 +661,7 @@ COMMANDS = {
             Opts.fail_under,
             Opts.ignore_errors,
             Opts.include,
+            Opts.no_combine,
             Opts.output_lcov,
             Opts.omit,
             Opts.quiet,
@@ -668,6 +679,7 @@ COMMANDS = {
             Opts.format,
             Opts.ignore_errors,
             Opts.include,
+            Opts.no_combine,
             Opts.omit,
             Opts.precision,
             Opts.sort,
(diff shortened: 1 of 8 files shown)


coverage/cmdline.py
```diff
@@ -873,7 +886,11 @@ def command_line(self, argv: list[str]) -> int:
         # plugins may try to, for example, to read Django settings.
         sys.path.insert(0, "")
 
-        self.coverage.load()
+        if options.no_combine:
+            self.coverage.load()
+        else:
+            self.coverage.load()
```
nedbat: Kind of disappointed Claude didn't factor this out!

thatch:
I went this route because if coverage is already combined, `coverage combine` is an error?  In case people are doing

```sh
coverage report
coverage combine
coverage report
```

but this is the least invasive way (it only changes the behavior of that first `report`, and does not introduce any new errors)

nedbat:
Sorry, which route? I'm missing something: with this code it's an error to combine if it's already been combined?

thatch:
The choices for fixing this were:
* just show a message
* combine for the user (and save to disk)
* combine for the user in memory only for reports
* fail to report if coverage needs to be combined

I chose the third option (the "route I went"), because it's the most helpful one that won't break someone's workflow if they already correctly use `coverage combine`.

For example:

```
ick$ coverage run -p -m pytest
ick$ ls .coverage*
.coverage.rygel.pid516142.X36NpWex.HJ0Mof874Lyh
.coverage.rygel.pid516286.XJ8maiIx.HMo3vtrY00vh
.coverage.rygel.pid516287.X2ZksaEx.Hpqw0TtWiBLh
.coverage.rygel.pid516311.XM9ZYrrx.HNQIp837rWOh
.coverage.rygel.pid516314.XzudWwcx.HNQIp837rWOh
ick$ coverage combine
Combined data file .coverage.rygel.pid516142.X36NpWex.HJ0Mof874Lyh
Combined data file .coverage.rygel.pid516286.XJ8maiIx.HMo3vtrY00vh
Combined data file .coverage.rygel.pid516287.X2ZksaEx.Hpqw0TtWiBLh
Combined data file .coverage.rygel.pid516311.XM9ZYrrx.HNQIp837rWOh
Skipping duplicate data .coverage.rygel.pid516314.XzudWwcx.HNQIp837rWOh
ick$ coverage combine
No data to combine
ick$ echo $?
1
```

If I made `coverage report` do a combine (just like the existing `combine` does, saving back to disk), it would make a later `coverage combine` now error out because it's already been done.  I thought that was too backwards-incompoatible to do the second choice; the first choice wouldn't have saved me much frustration.  The fourth choice also doesn't feel great.

nedbat:
I see, thanks. Quickly reading the code, I had missed that the combined data was not written out. My only concern with combining but not saving is that `coverage report; coverage json; coverage html` will combine three times, but explicitly combining first will fix that, and we can explain it in the docs.

nedbat:
I think we can do without `--no-combine`. I can't think of a reason why someone would want to skip combining, and if they really need to, they can copy data files around, or ask us later to add the option.

thatch:
Done, PTAL

nedbat:
We have a problem.  If the test cycle is `run; report`, the parallel data files are kept. The next `run; report` will combine twice as many files. The third cycle will have three times as many. Nothing is cleaning up the data files.  With the old `run; combine; report`, the combine step turned .coverage.* into .coverage, overwriting the previous .coverage.

I think it will be error-prone for these files to hang around, so we should re-consider having the implicit combine keep the data files. What would be bad about the implicit combine deleting the parallel files? Doing that would also mean that `run; report; html` wouldn't combine twice.

nedbat:
I'm making the change to remove the data files during the implicit combine.

nedbat:
The removal is commit 1cd47aa6ac1da4e150da44055295d4e4f3a014e8, and the changelog entry (etc) is in c2a3a284078556c911e0d9b6c6af1b7082a363ea.

nedbat:
This is now released as part of [coverage 7.14.0](https://pypi.org/project/coverage/7.14.0).

Druckles:
Hi, this is a very sensible feature, but guess what: [it breaks our workflow](https://xkcd.com/1172/)! (Please let me know if you'd rather I opened this as a separate ticket.)

We currently run tests over several modules and report them separately. Then we report the combined coverage over all modules, i.e.

```shell
cd modules/one
coverage run --parallel-mode
coverage combine --keep
coverage report --fail-under=80
# Files:
#  modules/one/.coverage
#  modules/one/.coverage.abc
#  modules/one/.coverage.def

cd ../two
coverage run --parallel-mode
coverage combine --keep
coverage report --fail-under=80
# Files:
#  modules/two/.coverage
#  modules/two/.coverage.ghi
#  modules/two/.coverage.jkl
```

and then:

```shell
coverage combine modules/*
coverage report
```

will combine the `modules/one/.coverage.*` and `modules/two/.coverage.*` files into a single `.coverage` file.

I can see the logic on combining for a report, as it saves a step in the majority of cases, but it's also less flexible. I've had to rollback to 7.13.5 for now, as copying the files first is more file management than I'd like (given that it worked before).

I can think of a few solutions:

1. `--no-combine` or similar, as suggested above
2. `--keep` like `coverage combine` so that it doesn't remove the old `.coverage.*` files
3. Use a .coverage if it's already there (and don't combine)
4. `coverage combine` has some sort of option to combine `.coverage` files

Option 2 would allow us to skip our first `coverage combine`-steps completely.

nedbat:
I've added `--keep-combined` to the reporting commands in commit f24a91f6d00b888c1d3d7c438cfe70294ad0c6ea.

PR #2171, opened 2026-05-15
Add contributor prerequisite notes

### LINKED ISSUE #2051: [docs] List platform-level prerequisites in the contributing doc
https://coverage.readthedocs.io/en/7.10.6/contributing.html mentions `git`, `make` and `tox` as implied. People will probably figure out that `tox` needs to be installed somehow. But while looking into #2050, I noticed that `tox` does run but ends up with a bunch of test failures that don't make it obvious that `gcc` is missing, for example.

I propose at least mentioning mandatory stuff explicitly. In case of a clean Fedora 43 container, a command to get all the deps is `dnf install -y gcc git make python3.13-freethreading python3.14-freethreading tox` (I think `make` might be pulled in by `gcc` so that one may not be strictly necessary; plus the downstream `tox` RPM depends on a list of normal Python builds but the FT ones need to be retrieved explicitly).

If you don't want to maintain OS-specific commands, just listing `git`, `make`, `tox` as prerequisites plus optional Python runtimes in addition, could be a good middle ground.

Lovlace777:
Summary
- Adds a short prerequisite note to the contributing guide before the setup steps.
- Calls out the platform tools used by the documented workflow: git, make, and a C compiler for the C tracer extension.
- Clarifies that the development requirements install tox, and that extra Python runtimes are only needed for the tox environments a contributor wants to run.

Validation
- `git diff --check` passed.
- `python -m tox --version` passed after installing the pinned tox requirements from `requirements/tox.pip`.
- `python -m tox -e doc --override testenv:doc.basepython=python` was attempted, but could not complete on this machine because only Python 3.11 is installed and the doc dependency `linklint==0.4.1` requires Python >=3.12.

Notes
- Fixes #2051.
- Documentation-only change; no runtime behavior impact.

### DIFF
--- a/doc/contributing.rst
+++ b/doc/contributing.rst
@@ -49,6 +49,13 @@ The coverage.py code is hosted on a GitHub repository at
 https://github.com/coveragepy/coveragepy.  To get a working environment, follow
 these steps:
 
+These steps assume you already have ``git`` and ``make`` installed.  Building
+and testing coverage.py also needs a C compiler, such as ``gcc`` or ``clang``,
+because the test suite builds the C tracer extension.  The development
+requirements install ``tox``; if you skip that step, install ``tox`` yourself
+before running tests.  Install any extra Python versions you want ``tox`` to
+test, including free-threading builds only if you are testing that support.
+
 #.  `Fork the repo`_ into your own GitHub account.  The coverage.py code will
     then be copied into a GitHub repository at
     ``https://github.com/GITHUB_USER/coveragepy`` where GITHUB_USER is your


doc/contributing.rst
```diff
@@ -49,6 +49,13 @@ The coverage.py code is hosted on a GitHub repository at
 https://github.com/coveragepy/coveragepy.  To get a working environment, follow
 these steps:
 
+These steps assume you already have ``git`` and ``make`` installed.  Building
+and testing coverage.py also needs a C compiler, such as ``gcc`` or ``clang``,
+because the test suite builds the C tracer extension.  The development
+requirements install ``tox``; if you skip that step, install ``tox`` yourself
+before running tests.  Install any extra Python versions you want ``tox`` to
```
nedbat, line 56:
```suggestion
requirements install ``tox``. Install any extra Python versions you want ``tox`` to
```
I would rather not add extra instruction as alternatives to the real instructions. It's smoother and easier if the steps flow linearly without having to make choices.

nedbat:
Thanks for doing this, it's hard to know how much of the foundations to describe, but this is a simple paragraph that can only help.

PR #2176, opened 2026-05-21
chore: improve coveragepy maintenance path

lphuc2250gma:
Summary:
- Add edge-case unit tests in tests/test_numbits.py for numbits.py round-tripping (empty input, single line, large line numbers, idempotency of nums_to_numbits then numbits_to_nums) and assert numbits_any_intersection on disjoint vs overlapping sets — pure mechanical test additions with no behavior change.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.


nedbat:
This is interesting. I'm curious what drew you to it?

PR #2177, opened 2026-05-22
Ignore exclusions inside multiline strings

### LINKED ISSUE #2112: Default `...` exclusion can exclude things by accident
By default, this regex is excluded from reporting:
```
^\s*(((async )?def .*?)?\)(\s*->.*?)?:\s*)?\.\.\.\s*(#|$)
```
This is meant to exclude `...` on a line by itself, as well as any function that has only that as its body.

Scriv has a [YAML literal](https://github.com/nedbat/scriv/blob/main/tests/test_literals.py#L139-L152) in its tests that looks like this:
```python
YAML_LITERAL = """\
---
version: 1.2.3

myVersion:
  MAJOR: 2
  MINOR: 3
  PATCH: 5

myproduct:
  version: [mayor=5, minor=6, patch=7]
  versionString: "8.9.22"
...
"""
```
The whole statement is mistakenly excluded from reporting because of the three dots at the end (YAML syntax for end of document).

puneetdixit200:
Summary:
- prevent exclusion regex matches inside multi-line string tokens from excluding the whole containing statement
- add a regression test for a YAML literal ending with `...`, which should still count the assignment as executable

Tests:
- `PYTHONWARNDEFAULTENCODING= /tmp/coveragepy-venv/bin/python -m pytest -q tests/test_coverage.py::ExcludeTest::test_default_ellipsis_exclusion_ignores_multiline_strings tests/test_coverage.py::ExcludeTest::test_default tests/test_parser.py::ExclusionParserTest::test_over_exclusion_bug1779`
- `PYTHONWARNDEFAULTENCODING= /tmp/coveragepy-venv/bin/python -m pytest -q tests/test_coverage.py::ExcludeTest`
- `PYTHONWARNDEFAULTENCODING= /tmp/coveragepy-venv/bin/python -m pytest -q tests/test_parser.py::ExclusionParserTest`
- `/tmp/coveragepy-venv/bin/python -m compileall -q coverage/parser.py tests/test_coverage.py`
- `git diff --check`

Fixes #2112


nedbat:
I'm wondering if multiline is important here, or if there are cases where a single line string would also cause a false match.

I'm also curious why you use such long tortuous test lines when `tox` would be enough? Should I update the contribution instructions?

nedbat:
Why did you close this PR?

puneetdixit200:
I closed this one because I replaced it with #2180 after your feedback. The newer PR adds the single-line string case, keeps the multiline coverage, and has the updated validation/results in the PR body.

nedbat:
OK, for the future, you don't need to close the first PR, you can keep pushing changes to your branch no matter how large the changes.

PR #2180, opened 2026-05-23
Ignore exclusions inside multiline strings

### LINKED ISSUE #2112: Default `...` exclusion can exclude things by accident

puneetdixit200:
Summary:
- prevent exclusion regex matches inside string tokens from excluding the containing statement
- add regression tests for YAML-style multiline literals and single-line string literals containing `...`
- keep real pragma comments on string assignment lines covered

Tests:
- `PYTHONWARNDEFAULTENCODING= /tmp/coveragepy-venv/bin/python -m pytest -q tests/test_coverage.py::ExcludeTest::test_default_ellipsis_exclusion_ignores_multiline_strings tests/test_coverage.py::ExcludeTest::test_default tests/test_parser.py::ExclusionParserTest::test_over_exclusion_bug1779`
- `PYTHONWARNDEFAULTENCODING= /tmp/coveragepy-venv/bin/python -m pytest -q tests/test_coverage.py::ExcludeTest`
- `PYTHONWARNDEFAULTENCODING= /tmp/coveragepy-venv/bin/python -m pytest -q tests/test_parser.py::ExclusionParserTest`
- `/tmp/coveragepy-venv/bin/python -m compileall -q coverage/parser.py tests/test_coverage.py`
- `git diff --check`

Supersedes #2177
Fixes #2112


coverage/parser.py
```diff
@@ -158,6 +158,18 @@ def _raw_parse(self) -> None:
                         ltext,
                     )
                 )
+            if toktype == token.STRING and slineno != elineno:
```
nedbat, line 161: This doesn't seem right to me: This is saying that any lines that are part of a multi-line string won't be excluded.  That's not taking into account what else is on those lines, or whether the multi-line string matched one of the exclusion rules.

For example, this test fails:
```python
        self.check_coverage(
            """\
            a = 1
            ignore_var = "2"
            ignore_var = '''
                3,4,5
                '''
            ignore_var = "6"
            a = 7
            """,
            excludes=["ignore_var"],
            lines=[1, 7],
        )
```

nedbat: I've added this as a test.

PR #2183, opened 2026-05-30
Feature/multiline json

realInt2k:
I added the multiline map thing from html report for json report

I also added --multiline option with comment so it's optional.

Purpose: I'm doing a thing where I get coverage for base execution, then coverage for another execution, and I want to diff them. With the original coverage.py, I wouldn't get multiline statement and have to do extra static analysis to get the full statement

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -368,6 +368,12 @@ class Opts:
         action="store_true",
         help="Display version information and exit.",
     )
+    multiline = optparse.make_option(
+        "",
+        "--multiline",
+        action="store_true",
+        help="Output actual covered lines and not just statements"
+    )
 
 
 class CoverageOptionParser(optparse.OptionParser):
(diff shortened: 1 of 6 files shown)


realInt2k:
image
for example, original command would just give 1218

nedbat:
I need to review this more closely, but some things stand out:
- There is no test for the new code path in jsonreport.py
- I don't understand why jsonreport.py needs to examine the source text itself. Isn't that information already analyzed someplace?

realInt2k:
Yes, I did not write any test for the multiline purpose yet (I tried to look for similar test case such as --pretty-print to get a bootstrap but couldn't, so far, just a simple cmdline call test)

>  don't understand why jsonreport.py needs to examine the source text itself. Isn't that information already analyzed someplace?

I don't think I fully understand this. By "examining the source text" you mean the mapping between the line and the parent statement that I did in my pull request? 

----
I wanted this function because I saw a discrepancy between json's report in the "executed lines" and the green lines in html report. Turned out json report only report covered statements, not actual lines (so a covered multiline statement would only have the first of its line reported)

The mappping (multiline map) was directly taken from FileReporter class, which was used in html report to show all the covered lines.

PR #2194, opened 2026-06-15
Fix default exclusion for multiline return type annotations with ellipsis body

### LINKED ISSUE #2185: Default `...` exclusion rule misses when return type annotation is split over multiple lines
**Describe the bug**

**To Reproduce**

`example.py`:

```python
from typing import Protocol

class C:
    def f() -> dict[
        str, str
    ]: ...
```

```console
$ coverage erase
```

```console
$ python --version --version

$ coverage --version

$ uv pip freeze
ast-serialize==0.5.0
coverage==7.14.1
librt==0.11.0
markdown-it-py==4.2.0
mdurl==0.1.2
mypy==2.1.0
mypy-extensions==1.1.0
pathspec==1.1.1
pip==26.1.1
pygments==2.20.0
rich==15.0.0
typing-extensions==4.15.0
```

**Expected behavior**

I-am-Uchenna:
Fixes #2185

When a return type annotation is split across multiple lines by a formatter
like ruff (e.g. `-> dict[\n    str, str\n]: ...`), the default exclusion
regex for ellipsis-body stubs fails to match the continuation line where
the closing bracket, colon, and `...` appear (e.g. `]: ...`).

The existing regex expects the `def` signature and `: ...` on the same
logical line, but formatters break long annotations so the closing
`]: ...` or `): ...` lands on a separate physical line.

## Changes

**coverage/config.py** - Added a new default exclusion regex:
`r"^\s*[\])]+\s*:\s*\.\.\.\s*(#|$)"`

This matches continuation lines that consist of closing brackets
followed by `: ...`, covering patterns like:
```python
def method() -> dict[
    str, str
]: ...          # now excluded
```

**tests/test_coverage.py** - Added `test_default_multiline_annotation`
to verify that multiline return type annotations with ellipsis bodies
are properly excluded from coverage measurement.

### DIFF
--- a/coverage/config.py
+++ b/coverage/config.py
@@ -150,6 +150,7 @@ TConfigParser = HandyConfigParser | TomlConfigParser
 DEFAULT_EXCLUDE = [
     r"#\s*(pragma|PRAGMA)[:\s]?\s*(no|NO)\s*(cover|COVER)",
     r"^\s*(((async )?def .*?)?\)(\s*->.*?)?:\s*)?\.\.\.\s*(#|$)",
+        r"^\s*[\])]+\s*:\s*\.\.\.\s*(#|$)",
     r"if (typing\.)?TYPE_CHECKING:",
 ]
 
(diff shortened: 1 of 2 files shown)


nedbat:
PR #2196 is handling this issue.

Issue #2187, opened 2026-06-05
sys.monitoring use_tool_id exception


gnufede:
**Describe the bug**
If coverage.py with the `sys.monitoring` API is used at the same time as another tool is using it to collect coverage, an unhandled exception will be raised:
```
ValueError: tool 1 is already in use
```

**To Reproduce**
Using Python 3.12+
```
# repro.py
import sys
import coverage


tool_id = sys.monitoring.COVERAGE_ID
sys.monitoring.use_tool_id(tool_id, "another coverage tool")

cov = coverage.Coverage(config_file=False)
cov.start()
```

`$ COVERAGE_CORE=sysmon python repro.py`

**Expected behavior**
The Exception is handled instead of exiting the process.

**Additional context**

nedbat:
> If coverage.py with the sys.monitoring API is used at the same time as another tool is using it to collect coverage

What is the other tool?

devdanzin:
It seems there's one called "scalpel": 

https://github.com/fun-altF4/scalpel/blob/072e5d4d4ffafd8c6029ffdef15ce0b0117a1825/probes/python/probe.py#L222-L230

And SlipCover too, but it checks for conflicts:

https://github.com/plasma-umass/slipcover/blob/dd9fdbb6c85896b6fd2c29f8d7d87559b94de793/src/slipcover/slipcover.py#L382-L386

gnufede:
> What is the other tool?

dd-trace-py used like `pytest --ddtrace` in my case

nedbat:
What would you like coverage to do? Raise a different exception? [dd-trace-py does this](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/internal/coverage/instrumentation_py3_12.py#L80-L83):
```python
    coverage_tool = sys.monitoring.get_tool(sys.monitoring.COVERAGE_ID)
    if coverage_tool is not None and coverage_tool != "datadog":
        log.debug("Coverage tool '%s' already registered, not gathering coverage", coverage_tool)
        return code, CoverageLines()
```

nedbat:
In retrospect, the sys.monitoring tool-id scheme seems too simplistic. Why pre-allocate an ID to coverage if more than one tool might want to use it?

gnufede:
[...]

PR #2196, opened 2026-06-18
Exclude `...` bodies after multi-line return-type annotations

### LINKED ISSUE #2185: Default `...` exclusion rule misses when return type annotation is split over multiple lines

mokashang:
The default exclusion regex for `...` function bodies expects the def
signature to end with `)`. When a formatter (e.g. Ruff) splits a long
return-type annotation across multiple lines, the last line of the
signature ends with `]` instead — for example:

```python
class C:
    def f() -> dict[
        str, str
    ]: ...
```

With the current regex this function's body is reported as a missing
line, so coverage drops below 100% even though the body is just `...`.

This change widens the optional def-header group so the closing
bracket can be any (non-empty) run of `]` and `)`. That lets the
regex still match:

- `def f(): ...`
- `def f() -> int: ...`
- multi-line def signatures that close with `): ...` (existing case)
- multi-line def signatures that close with `]: ...` (the bug)
- multi-line def signatures that close with `]) -> int: ...`
- `Tuple[int, ...]` parameter types whose inner `...` should not
  itself trigger the rule

It does not start matching unintended things: an annotated assignment
like `x: list[int] = ...` still doesn't match (no `:` after the `]`),
and a regular function with a real body doesn't match either.

End-to-end reproducer from #2185:

```console
$ cat example.py
class C:
    def f() -> dict[
        str, str
    ]: ...
$ coverage run --branch -m example && coverage report --fail-under=100
```

Before the change: `example.py 3 0 2 1 80%` (fail-under trips).
After:  `example.py 2 0 0 0 100%`.

### Testing

- Added `ExcludeTest::test_default_ellipsis_multiline_return_type` covering
  single-line, split-return, split-param-and-return, and inner-`Tuple[..., ...]`
  shapes.
- Existing `ExcludeTest::test_default` and the rest of `ExcludeTest` still
  pass (7/7).
- `tests/test_coverage.py tests/test_config.py tests/test_parser.py` —
  233 passed, 7 skipped.
- `ruff format --check` and `mypy --strict` on the changed files — clean.

Fixes #2185.


nedbat:
Thanks, I appreciate that you updated the changelog as well.  The PR description includes things that should not be excluded. Please add those to the test to demonstrate that they are not matched.

You'll also need to resolve conflicts.

mokashang:
Thanks for the review. Pushed an update:

- Resolved the `CHANGES.rst` conflict by folding the entry into the existing Unreleased block beside the #2197 and #2187 notes.
- Extended `test_default_ellipsis_multiline_return_type` with two negative cases right under the stub block:
  - `x: list[int] = ...` — an annotated assignment whose RHS happens to be `Ellipsis`. It ends in `...` and contains `]`, so it's the most obvious shape that could be over-matched by the widened pattern.
  - `def with_body() -> dict[str, str]: return {}` — a real one-line def whose return type contains `]`.

`lines=[1, 15, 16, 17]` now asserts that both of those, plus the trailing `z = 17`, remain executable statements (i.e. they are *not* excluded by the rule). I also re-ran the suite locally:

- `ExcludeTest` — 7/7 pass
- `tests/test_coverage.py tests/test_config.py tests/test_parser.py` — 233 passed, 7 skipped

(Re-checked the regex out of paranoia against a handful of other shapes that contain `...` or `]` — bare `y = ...`, `y: int = ...`, `class C: ...`, `x = (1, 2, ...)`, `foo[...]` — none of them match.)

nedbat:
Thanks, what name should I put in the contributors list?

mokashang:
Thank you! You can put Mengjia Shang as my name. :)

nedbat:
This is now released as part of [coverage 7.14.3](https://pypi.org/project/coverage/7.14.3).

PR #2202, opened 2026-06-21
Reject switch_context with sysmon core

itscloud0:
## Summary
- raise `CoverageException` if `switch_context()` is called while running on `core=sysmon`
- keep the existing startup-time fallback for configured dynamic contexts unchanged
- document that explicit context switching requires `ctrace` or `pytrace`, and add a regression test

## Problem
`switch_context()` worked under `ctrace`, but under `sysmon` a shared line hit by two contexts only kept the first one. The API path had no guard, so Python 3.14's default sysmon core could silently produce wrong context data.

## Solution
Instead of silently downgrading to a slower tracing core after measurement has already started, reject the API call and tell users to choose `core=ctrace` or `core=pytrace` explicitly when they need dynamic contexts.

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -678,38 +678,21 @@ class Coverage(TConfigurable):
             concurrency=concurrency,
         )
 
-    def _switch_to_tracing_core_for_contexts(self) -> None:
-        """Move off sysmon before explicit context switching."""
+    def _warn_no_sysmon_context_switching(self) -> bool:
+        """Warn when sysmon can't honor explicit context switching."""
         assert self._collector is not None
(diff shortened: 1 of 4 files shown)


nedbat:
I'm not sure I like this change. Switch from sysmon (the fastest core) to pytrace (the slowest) doesn't seem like a good way to handle sysmon's inability to trace contexts. If someone is willing to pay the performance penalty for tracing contexts, I would rather they explicitly choose the right core. 

I'm also not sure why it falls back to pytrace instead of ctrace?

Maybe there's some subtlety here I'm missing.

PR #2204, opened 2026-06-22
Handle loop-completion arc before function exit

### LINKED ISSUE #2168: 3.14: uncovered branch for loop completion with break
**Describe the bug**
In python 3.14 there is a uncovered branch detected in for loops that have a conditional break and a statement thereafter, when using `COVERAGE_CORE=ctrace`.

**To Reproduce**
```python
def foo(*x):
    for i in x:
        if i:
            break
    pass


foo()
foo(0)
foo(1)
```

The coverage shows for line 2: `line 2 didn't jump to line 5 because the loop on line 2 didn't complete`

Versions:
- CPython: 3.14.4
- coverage: 7.13.5
- pytest: 9.0.3
- pytest-cov: 7.1.0

**Expected behavior**
The line should be considered covered, since e.g. `foo(0)` will exhaust the loop.
It works correctly if:
- using CPython 3.13.3
- using `COVERAGE_CORE=sysmon` (or no override)
- if there is a statement after the `if` within the `for` loop
- if there is no statement after the `for` loop
- if the `for` loop has an `else` branch

**Additional context**
This may be related to: #1987

itscloud0:
## Summary
- fix the missing loop-completion arc on Python 3.14 when a `break` exits the final loop iteration immediately before function return
- apply the same correction in both the C tracer and Python tracer
- add a focused regression covering the `break` case

## Problem
Fixes #2168.

`coverage report -m` could show a missing branch after a `break`-driven loop exit on Python 3.14 even though the code path was executed.

## Solution
Record the current line as the final arc before emitting the function-exit arc when returning from a loop-completion path.

### DIFF
--- a/coverage/ctracer/tracer.c
+++ b/coverage/ctracer/tracer.c
@@ -771,6 +771,13 @@ CTracer_handle_return(CTracer *self, PyFrameObject *frame)
             real_return = !(is_yield || is_yield_from);
 #endif
             if (real_return) {
+                int lineno = PyFrame_GetLineNumber(frame);
+                if (lineno != self->pcur_entry->last_line) {
+                    if (CTracer_record_pair(self, self->pcur_entry->last_line, lineno) < 0) {
+                        goto error;
+                    }
+                    self->pcur_entry->last_line = lineno;
+                }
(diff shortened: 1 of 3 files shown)


nedbat:
Closing in favor of #2281

PR #2206, opened 2026-06-23
Skip excluded functions in LCOV function totals

### LINKED ISSUE #2205: LCOV function records include functions excluded by report config
## Describe the bug

`coverage lcov` includes function records for functions that are excluded by `[report] exclude_also`. The line records respect the exclusions, but `FN`/`FNDA`/`FNF`/`FNH` still count the excluded function, causing tools like `genhtml` to report lower function coverage even when line coverage is 100%.

This was observed with excluded abstract methods such as `@(abc\.)?abstractmethod`.

This issue was drafted with AI assistance, but it describes a real observed issue from an actual codebase. The behavior was reproduced with a minimal example and human-validated before filing.

## To Reproduce

`module_under_test.py`:

```python
from abc import ABC, abstractmethod


class Base(ABC):
    @abstractmethod
    def excluded_abstract(self) -> int:
        raise NotImplementedError

    def covered(self) -> int:
        return 1


class Concrete(Base):
    def excluded_abstract(self) -> int:
        return 2


assert Concrete().covered() == 1
assert Concrete().excluded_abstract() == 2
```

`.coveragerc`:

```ini
[run]
source = .

[report]
exclude_also =
    @(abc\.)?abstractmethod
    raise NotImplementedError
```

Commands:

```bash
python -m coverage run module_under_test.py
python -m coverage report
python -m coverage lcov -o coverage.lcov
```

Text report:

```text
Name                   Stmts   Miss  Cover
------------------------------------------
module_under_test.py       9      0   100%
```

LCOV output includes the excluded abstract method:

```text
FN:6,7,Base.excluded_abstract
FNDA:0,Base.excluded_abstract
FN:9,10,Base.covered
FNDA:1,Base.covered
FN:14,15,Concrete.excluded_abstract
FNDA:1,Concrete.excluded_abstract
FNF:3
FNH:2
```

## Expected behavior

Because `Base.excluded_abstract` is excluded from reportable statements, it should not contribute to LCOV function totals. Expected function records:

```text
FN:9,10,Base.covered
FNDA:1,Base.covered
FN:14,15,Concrete.excluded_abstract
FNDA:1,Concrete.excluded_abstract
FNF:2
FNH:2
```

## Versions tested

- coverage.py 7.13.4
- coverage.py 7.14.3

I also checked the current `main` implementation and the LCOV function record path still appears to count all function code regions before filtering out fully excluded regions.

## Context

This is similar in spirit to #1684, but for LCOV function records rather than line records. The line totals are correct; only function records ignore/report around excluded regions.

TrapsterDK:
Fixes #2205.

## Summary

- Skip LCOV function records when a function region has no reportable statements after exclusions are applied.
- Count `FNF` from emitted functions rather than all discovered function regions.
- Add a regression test for an excluded `@abstractmethod` that raises `NotImplementedError`.

## Details

Line records already respect report exclusions. Function records were still generated from all discovered function code regions, so a fully excluded abstract method could appear as `FNDA:0` and reduce downstream LCOV function coverage even when the text report shows 100% line coverage.

This narrows each function region before emission and skips regions with `n_statements == 0`, matching the line-reporting behavior for fully excluded code.

## Tests

- `.venv/bin/python -m pytest tests/test_lcov.py::LcovTest::test_excluded_functions -q -n 0`
- `.venv/bin/python -m pytest tests/test_lcov.py -q -n 0`

### DIFF
--- a/coverage/lcovreport.py
+++ b/coverage/lcovreport.py
@@ -100,9 +100,8 @@ def lcov_functions(
         outfile.write(f"FN:{first_line},{last_line},{region.name}\n")
         outfile.write(f"FNDA:{hit},{region.name}\n")
 
-    if functions_found:
-        outfile.write(f"FNF:{functions_found}\n")
-        outfile.write(f"FNH:{functions_hit}\n")
+    outfile.write(f"FNF:{functions_found}\n")
+    outfile.write(f"FNH:{functions_hit}\n")
 
 
 def lcov_arcs(
--- a/coverage/lcovreport.py
+++ b/coverage/lcovreport.py
@@ -84,19 +84,25 @@ def lcov_functions(
     narrower.add_regions(r.lines for _, _, r in functions)
 
     functions.sort()
+    functions_found = 0
     functions_hit = 0
     for first_line, last_line, region in functions:
         # A function counts as having been executed if any of it has been
         # executed.
         analysis = narrower.narrow(region.lines)
+        if analysis.numbers.n_statements == 0:
+            continue
+
+        functions_found += 1
         hit = int(analysis.numbers.n_executed > 0)
         functions_hit += hit
 
         outfile.write(f"FN:{first_line},{last_line},{region.name}\n")
         outfile.write(f"FNDA:{hit},{region.name}\n")
 
-    outfile.write(f"FNF:{len(functions)}\n")
-    outfile.write(f"FNH:{functions_hit}\n")
+    if functions_found:
+        outfile.write(f"FNF:{functions_found}\n")
+        outfile.write(f"FNH:{functions_hit}\n")
 
 
 def lcov_arcs(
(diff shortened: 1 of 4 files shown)


coverage/lcovreport.py
```diff
@@ -84,19 +84,25 @@ def lcov_functions(
...
+        functions_found += 1
         hit = int(analysis.numbers.n_executed > 0)
         functions_hit += hit
 
         outfile.write(f"FN:{first_line},{last_line},{region.name}\n")
         outfile.write(f"FNDA:{hit},{region.name}\n")
 
-    outfile.write(f"FNF:{len(functions)}\n")
-    outfile.write(f"FNH:{functions_hit}\n")
+    if functions_found:
```
nedbat: Are you sure we should omit these lines if no functions are found/hit? Wouldn't it be better to output `FNF:0` ?

TrapsterDK: I believe you are correct, I also added a regression test to assert this behaviour.

nedbat:
This is now released as part of [coverage 7.15.0](https://pypi.org/project/coverage/7.15.0).

PR #2209, opened 2026-06-28
Start attaching the loader on `__spec__`

### LINKED ISSUE #2208: `DeprecationWarning: Module globals; __loader__ != __spec__.loader` on 3.15
**Describe the bug**
When I use `coverage`, `pytest`, and `traceback.print_stack()` together, I get a deprecation warning on 3.15. I'm not convinced it's `coverage`'s fault, but opening here first.

**To Reproduce**

`x.py`:
```py
import traceback

traceback.print_stack()
```

Now run `coverage run -m pytest x.py` (after `pip install coverage pytest`). Note that the `pytest` part *and* the `coverage` part is necessary, if it's missing either things work correctly. You should see something like this:
```
a5rocks@whimsy /v/f/p/j/T/tmp.yE0qQYS20D> uv run coverage run -m pytest x.py
==================================== test session starts ====================================
platform darwin -- Python 3.15.0a6, pytest-9.1.1, pluggy-1.6.0
rootdir: /private/var/folders/py/j2lyj17j2sxg5xgjg2sxkx1h0000gn/T/tmp.yE0qQYS20D
collected 0 items

===================================== warnings summary ======================================
../../../../../../../Users/a5rocks/.local/share/uv/python/cpython-3.15.0a6-macos-aarch64-none/lib/python3.15/linecache.py:273
  /Users/a5rocks/.local/share/uv/python/cpython-3.15.0a6-macos-aarch64-none/lib/python3.15/linecache.py:273: DeprecationWarning: Module globals; __loader__ != __spec__.loader
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
==================================== 1 warning in 0.00s =====================================
```

Answer the questions below:
1. What version of Python are you using? 3.15.0a6
1. What version of coverage.py shows the problem? The output of `coverage debug sys` is helpful. 7.14.1
1. What versions of what packages do you have installed? The output of `pip freeze` is helpful. `pytest==9.0.3` also
1. What code shows the problem?  Give us a *specific commit* of a *specific repo* that we can check out. If you've already worked around the problem, please provide a commit before that fix. `x.py` above
1. What commands should we run to reproduce the problem? *Be specific*. Include everything, even `git clone`, `pip install`, and so on. Explain like we're five! command given above.

**Expected behavior**
No deprecation warning.

A5rocks:
Fixes https://github.com/coveragepy/coveragepy/issues/2208

[The documentation for `__loader__`](https://docs.python.org/3/reference/datamodel.html#module.__loader__) says:
> Deprecated since version 3.12, will be removed in version 3.16: Setting `__loader__` on a module while failing to set `__spec__.loader` is deprecated. In Python 3.16, `__loader__` will cease to be set or taken into consideration by the import system or the standard library.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,9 +32,17 @@ Unreleased
   as described in `issue 2205`_. This is now fixed thanks to `Martin Kuntz
   Jacobsen <pull 2206_>`_.
 
+- When running your program, coverage now correctly sets
+  ``yourmodule.__spec__.loader`` as `strongly recommended <--loader--_>`_,
+  avoiding the deprecation warning described in `issue 2208`_. Thanks, `A5rocks
+  <pull 2209_>`_.
+
+.. _--loader--: https://docs.python.org/3/reference/datamodel.html#module.__loader__
 .. _issue 2198:  https://github.com/coveragepy/coveragepy/issues/2198
 .. _issue 2205:  https://github.com/coveragepy/coveragepy/issues/2205
 .. _pull 2206: https://github.com/coveragepy/coveragepy/pull/2206
+.. _issue 2208:  https://github.com/coveragepy/coveragepy/issues/2208
+.. _pull 2209: https://github.com/coveragepy/coveragepy/pull/2209
 
 
 .. start-releases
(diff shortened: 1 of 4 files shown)


nedbat:
This is now released as part of [coverage 7.15.0](https://pypi.org/project/coverage/7.15.0).

PR #2213, opened 2026-07-04
Remove a cast that occurs in the hot path in AstArcAnalyzer

reaperhulk:
I'm not sure why this was a cast, but it doesn't appear to be necessary any more. Since cast() evaluates its first argument, this constructs a new `collections.abc.Callable` on every call. This change speeds up arc analysis somewhere around 10% for pyca/cryptography (which is, admittedly, only about 100-200ms).

### DIFF
--- a/coverage/parser.py
+++ b/coverage/parser.py
@@ -752,9 +752,8 @@ class AstArcAnalyzer:
 
         """
         node_name = node.__class__.__name__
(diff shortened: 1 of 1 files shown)


nedbat:
I'm not sure why it was there either!

nedbat:

PR #2214, opened 2026-07-04
Use per-instance caches in PythonParser

reaperhulk:
first_line and exit_counts were cached with functools.lru_cache on the methods.  A cache on a method is keyed by (self, args), so it pins the parser instances themselves.

Caching on the instance instead makes reporting on a 99-file corpus 29% faster (1.78s to 1.27s, nearly all of it collector time) and drops live objects after the run from ~387k to ~17k.

### DIFF
--- a/coverage/parser.py
+++ b/coverage/parser.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 
 import ast
 import collections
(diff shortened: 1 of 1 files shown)


nedbat:

PR #2215, opened 2026-07-04
cache file reporters and analyses for the whole reporting phase

reaperhulk:
_analyze and _get_file_reporter used lru_cache(maxsize=1), which unfortunately doesn't work when reporting sweeps the file list, e.g. pytest-cov's total pass followed by its term report.

We now cache them in per-instance dicts instead, clearing whenever measured data changes (flush, load, combine, erase).

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -305,6 +305,8 @@ class Coverage(TConfigurable):
         self._file_mapper: Callable[[str], str] = abs_file
         self._data_suffix = self._run_suffix = None
         self._exclude_re: dict[str, str] = {}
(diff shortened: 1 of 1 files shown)


reaperhulk:
For cryptography, running this on a 4 core Linux VM (5 averaged runs)

||total pass|term report|total|
|-|-|-|-|
|7.15.0|3.74s|3.81s|7.55s|
|PR|3.78s|0.04s|3.81s|

(The total pass is done first to do fail under checks, then it produces whatever potential output you might want, e.g. the terminal report, HTML, etc)

nedbat:

PR #2216, opened 2026-07-04
only walk statement nodes when scanning for defs and classes

reaperhulk:
`_raw_parse` and `AstArcAnalyzer.analyze` use `ast.walk` to find module, function, and class nodes, visiting every node in the tree. Those are all statements, which can never appear inside expression subtrees.

We can optimize by adding `walk_statement_nodes` which filters out statements.

This also replace analyze()'s per-node getattr dispatch with isinstance checks against the three handler types.

Some performance tests as run by claude (4 core x86_64 VM):

| command | 7.15.0 | ast-walk | delta  |
|---------|--------|----------|--------|
| report  | 4.25s  | 3.83s    | -10.0% |
| html    | 12.88s | 12.27s   | -4.7%  |
| xml     | 4.77s  | 4.38s    | -8.2%  |
| json    | 5.38s  | 5.10s    | -5.2%  |
| lcov    | 7.85s  | 7.21s    | -8.2%  |

~0.4s saved (with some variance due to standard benchmarking challenges).

### DIFF
--- a/coverage/parser.py
+++ b/coverage/parser.py
@@ -744,8 +744,6 @@ class AstArcAnalyzer:
         self.process_return_exits(exits)
         self.block_stack.pop()
 
-    _code_object__AsyncFunctionDef = _code_object__FunctionDef
-
     def _code_object__ClassDef(self, node: ast.ClassDef) -> None:
         start = self.line_for_node(node)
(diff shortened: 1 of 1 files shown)


nedbat:

PR #2218, opened 2026-07-05
enable only sys.monitoring events that have registered callbacks

reaperhulk:
Local events included `PY_RETURN|PY_RESUME`  even in modes that register no callback for them. Enabled events with no callback still dispatch on every hit, and never get disabled since only a callback can return DISABLE -- so every return/resume of traced code paid dispatch cost Line mode also skips the byte-to-line map now, which only arc callbacks read.

Some benchmarks (all values relative to an uninstrumented Python 3.14.6).

### Line mode

| workload | overhead before | overhead after |
|---|---:|---:|
| generators, sysmon_bench micro (generator resumes) | 1.29x | 1.02x |
| nqueens, pyperformance | 1.26x | 1.14x |
| richards, pyperformance | 1.20x | 1.02x |

### Branch mode

| workload  | overhead before | overhead after |
|---|---:|---:|
| generators, sysmon_bench micro (generator resumes) | 1.29x | 1.00x |
| nqueens, pyperformance | 1.26x | 1.14x |
| richards, pyperformance | 1.03x | 1.02x | 

I'm sorry about all these separate PRs @nedbat, let me know if there's anything I can do to make this easier (or just stop until you've had time to digest them!).

### DIFF
--- a/coverage/sysmon.py
+++ b/coverage/sysmon.py
@@ -359,7 +359,8 @@ class SysMonitor(Tracer):
                 finally:
                     self.unlock_data()
                 file_data = self.data[tracename]
-                b2l = bytes_to_lines(code)
+                # byte_to_line is only read by the arc callbacks
+                b2l = bytes_to_lines(code) if self.trace_arcs else None
             else:
                 file_data = None
                 b2l = None
@@ -381,10 +382,12 @@ class SysMonitor(Tracer):
                 with self.lock:
                     if self.sysmon_on:
                         assert sys_monitoring is not None
-                        local_events = events.PY_RETURN | events.PY_RESUME | events.LINE
+                        local_events = events.LINE
                         if self.trace_arcs:
                             assert env.PYBEHAVIOR.branch_right_left
-                            local_events |= events.BRANCH_RIGHT | events.BRANCH_LEFT
+                            local_events |= (
+                                events.PY_RETURN | events.BRANCH_RIGHT | events.BRANCH_LEFT
+                            )
                         sys_monitoring.set_local_events(self.myid, code, local_events)
 
                         if LOG:  # pragma: debugging


nedbat:
> I'm sorry about all these separate PRs @nedbat, let me know if there's anything I can do to make this easier (or just stop until you've had time to digest them!).

These are all fine, and I would prefer separate PRs. My attention is divided, but I expect to be able to get to these soon.  I really appreciate you asking about this. I have another prolific PR writer at the moment who won't engage with me, which I find frustrating.

I'm interested to learn more about how you are finding these issues, and whether we can capture some of that work as a repeatable methodology. I don't know if you've seen https://github.com/coveragepy/benchmark for example?

nedbat:
Do you use Discord? There's a [#coverage-py channel](https://discord.com/channels/267624335836053506/1253355750684753950) if more interactivity would help.

reaperhulk:
I don't typically use Discord but I joined the server so we can chat there as needed!

The way these were found was by aggressively profiling the pyca/cryptography test suite looking for ways to cut wall-clock time. I pointed an LLM at a variety of potential paths (mostly non-coverage) and kept fanning out, then triaging the results and distilling them into useful/minimal changes that could be understood and submitted upstream. One of the surprises was that reports were expensive in our environment, which is what many of the PRs actually target. (We do a coverage report on every job + a final combine report across all jobs as a dependent job with a 100% requirement to pass)

So it vaguely follows the coveragepy/benchmark strategy of real-world test suites, but entirely focused on one repository 😆. 

This PR was found via similar methodology after I did some CTracer optimizations I had thought about a few years ago but never gotten around to. Those optimizations actually yielded a moderate performance boost, but are a relatively large chunk of C code I didn't want to submit, so I looked more closely at the sysmon behavior since 3.14+ is both the future and already the majority of our jobs.

Since most of these PRs are either focused on a very small slice of behavior (reporting) coveragepy/benchmark doesn't really show improvement. That said, for this PR the coverage overhead drops substantially on line coverage while being within the noise floor with branch enabled (c'est la vie, that's the path we care about for cryptography).

nedbat:

PR #2220, opened 2026-07-07
compute multiline maps cheaply in the sysmon core

reaperhulk:
(This PR was generated via claude but has undergone significant review and changes based on that review before submission)

When measuring branch coverage under `sys.monitoring`, the tracer needs each file's multiline map to resolve branch events. It currently gets it by running a full `PythonParser.parse_source()`. This change extracts the map computation into a shared `multiline_map_from_tokens()`. This is refactored to be the single place the map is built, used by both `PythonParser._raw_parse` and the sysmon core.

On the cryptography test suite (Python 3.14, branch mode), sysmon overhead drops from +35.2% to +30.0%; the remaining overhead is mostly `branch_trails()` analysis, which is the next PR. 😄


nedbat:

PR #2221, opened 2026-07-07
resolve sysmon branch events lazily, one pair at a time

reaperhulk:
Like #2220, this PR was generated via claude but has undergone significant review and changes based on that review before submission.

sysmon branch mode currently pre-analyzes every branch site in a code object on that object's first branch event. This is two `dis.get_instructions()` decodes plus trail walks per conditional jump. Since branch events are one-shot (each site/direction is DISABLEd after first fire), at most two (source, destination) pairs per site ever need resolving, so this PR drops the precompute and resolves each pair lazily as its event arrives by walking the raw co_code bytes from the destination; `branch_trails()`, `always_jumps()`, and `InstructionWalker` are removed.

This results in a quite significant performance improvement in cryptography's test suite (these numbers do not include the other performance optimizations that have recently landed as I ran them before rebasing to main)

| Configuration | Wall time | Overhead|
-- | -- | --
No coverage | 35.44 s | —
sysmon branch, main | 47.90 s | +35.2%
sysmon branch, this PR | 38.33 s | +8.2%

In conjunction with #2220 it looks like we'll have branch coverage down to a similar overhead as line coverage with the sysmon core.

I'm very open to ideas on how we can improve the confidence in correctness/reviewability for this! 😄  I've run it against both this branch and latest coverage.py on several suites without seeing regression, but LLM-driven significant refactors like this (even with ongoing human feedback during the process, as this had) make me nervous since I lack a deep understanding. I've reviewed this and at the very least it isn't clearly wrong, but that's not a sufficient bar.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,7 +30,8 @@ Unreleased
 
 - A number of performance improvements thanks to Paul Kehrer, in pull requests
   `2213 <pull 2213_>`_, `2214 <pull 2214_>`_, `2215 <pull 2215_>`_, `2216
-  <pull 2216_>`_, `2218 <pull 2218_>`_, and `2220 <pull 2220_>`_.
+  <pull 2216_>`_, `2218 <pull 2218_>`_, `2220 <pull 2220_>`_, and `2221 <pull
+  2221_>`_.
 
 .. _pull 2213: https://github.com/coveragepy/coveragepy/pull/2213
 .. _pull 2214: https://github.com/coveragepy/coveragepy/pull/2214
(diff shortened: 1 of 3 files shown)


nedbat:
I'm disheartened to realize again how most of bytecode.py is run inside the tracers, and so shows as not covered in the coverage reports. I'd like to have direct tests of that code. But that's not the fault of this PR.

nedbat:
Now this has a complicated merge conflict... :(

reaperhulk:
Resolving the branch conflict right now. I think we can improve coverage on bytecode.py as well, but would you prefer that in this PR or as a follow-up?

nedbat:

PR #2224, opened 2026-07-11
fix: escape context labels in html report inline script block

rajath201:
With `coverage html --show-contexts`, each file page carries its context labels as JSON inside an inline `<script>` element. `switch_context` takes any string (pytest-cov feeds it test node ids, which can hold arbitrary characters through parametrization), so a label like `</script>...` closes the script element early and the browser parses whatever follows as live markup in the published report.

Escape the HTML-significant characters in that JSON as `\uXXXX` right where it's built. The escapes are still valid JSON, so `coverage_html.js` reads the same label values back, and keeping the escaping at the serialization point covers every label without each context producer having to sanitize its own strings.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,11 @@ upgrading your version of coverage.py.
 Unreleased
 ----------
 
+- Fix: in the HTML report with ``show_contexts`` enabled, a context label
+  containing ``</script>`` (for example a parametrized pytest node id) could
+  close the inline ``<script>`` element in a file page early, injecting markup.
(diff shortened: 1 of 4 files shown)


nedbat:
Thanks!

rajath201:
Thanks for merging! This is my first open-source contribution. I'd love any feedback to help me improve.

nedbat:
You did great. I really appreciate that you included a test and a changelog entry. Most people include tests, but not the changelog.  I hope I gave you the right attribution.

nedbat:

PR #2226, opened 2026-07-12
fix: neutralize control characters in lcov report fields

rajath201:
The lcov writer drops file names and the function and branch fields straight into its newline-delimited records with no escaping, unlike the html, markdown, xml, and json reports which each neutralize untrusted values. A source file whose name contains a newline (legal on POSIX, and reachable whenever coverage measures a tree you don't fully control) spills past the `SF:` line and forges its own `SF`/`DA`/`LH` records, so a downstream reader like genhtml or a CI coverage gate sees fabricated files and inflated hit counts.

`lcov_field` replaces control characters in each field right before it's written, keeping the value on one line while leaving normal names (printable, including non-ASCII) untouched. Keeping the guard at the write sites covers every field the format treats as free text in one spot, with no change to valid output. The added test measures a file whose name embeds a fake record and checks the report stays a single `SF` record.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,12 @@ upgrading your version of coverage.py.
 Unreleased
 ----------
 
+- Fix: the LCOV report wrote file names and other fields into its
+  line-oriented records without neutralizing control characters. A measured
+  file whose name contained a newline (legal on POSIX) could forge extra
+  records, inflating the coverage seen by tools that read the report. Control
+  characters in a field are now replaced.
+
 - Fix: in the HTML report with ``show_contexts`` enabled, a context label
   containing ``</script>`` (for example a parametrized pytest node id) could
   close the inline ``<script>`` element in a file page early, injecting markup.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,8 +28,16 @@ Unreleased
   and prev/next links, so it could close the attribute early and inject markup.
   Page URLs are now escaped. Thanks, `Rajath Mohare <pull 2227_>`_.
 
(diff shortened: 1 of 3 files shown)


rajath201:
gentle ping

nedbat:
TBH, this seems like a very unlikely scenario, and isn't a security concern. A file name with a newline will likely cause havoc in many places. A broken .lcov file seems like the least of their worries, and will only cause troubles for themselves.

Unless you have a more compelling argument, I am inclined to close this.

devdanzin:
I independently found this same issue.

The only (pretty weak) argument I can muster is consistency: dd806350 and e06eb348 did something similar.

I think every line-oriented LCOV consumer like genhtml, Coveralls, Codecov, Sonar would trip on the invalid line created, but not sure that counts for much.

rajath201:
No argument on the security framing, and newline filenames are certainly rare. The case I'd make is the one devdanzin raised: after dd806350 and e06eb348, lcov is the only writer left that can produce output its own format can't represent, and the failure can be silent since the spilled name reads back as extra SF/LH records with bogus totals rather than a parse error. The guard is a few lines at the write sites and leaves all valid output byte-identical. If that still doesn't clear the bar, no objection to closing it.

nedbat:
OK, rebase onto main, and we can get it merged.

rajath201:
Rebased onto main. Only fixup needed was re-placing the changelog entry in the current Unreleased section, since the old one has since shipped. Code and test are unchanged, and the lcov tests pass locally.

nedbat:
This is now released as part of [coverage 7.15.4](https://pypi.org/project/coverage/7.15.4).

PR #2227, opened 2026-07-12
fix: escape filename urls in html report href attributes

rajath201:
The HTML report turns each source file name into a page URL with flat_rootname and drops it into the href of the index rows, region links, and prev/next nav without escaping. A measured file whose name contains a double quote (legal on POSIX, and what you get running coverage over a tree you don't control) closes the href early, so the rest of the name becomes live attributes on the link, like an onmouseover handler that runs in the viewer's browser.

escape() is only meant for text and doesn't touch quotes, so this adds a small escape_url filter that entity-escapes the URL and applies it at each href in index.html and pyfile.html. Escaping in the template keeps the on-disk page names intact while the emitted links, which browsers decode back to the real names, stay valid.


rajath201:
any update?

coverage/html.py
```diff
@@ -859,6 +860,18 @@ def escape(t: str) -> str:
     return t.replace("&", "&amp;").replace("<", "&lt;")
 
 
+def escape_url(url: str) -> str:
```
nedbat: Somehow I missed `html.escape` [in the stdlib](https://docs.python.org/3/library/html.html#html.escape) which can be used for both URLs and text. We don't need either `escape` or `escape_url` it seems.

rajath201: Done, dropped both helpers and registered stdlib html.escape as the escape filter. It also escapes quotes and > in text positions, so a handful of gold files and expected strings picked up `&quot;`/`&#x27;` entities; updated those to match. test_html.py is green.

nedbat:
I fixed the two failures

nedbat:
This is now released as part of [coverage 7.15.4](https://pypi.org/project/coverage/7.15.4).

Issue #2229, opened 2026-07-13
7.15.1 regression: `coverage json` uses ~5-6x more memory than 7.15.0


harrismirza-decagon:
[coveragepy-json-memory-repro.zip](https://github.com/user-attachments/files/29964225/coveragepy-json-memory-repro.zip)

**Describe the bug**

Coverage.py 7.15.1 introduces a large memory-usage regression when generating a JSON report from a sizeable branch-coverage database.

Using the attached synthetic reproducer with the same source tree, configuration, and `.coverage` database:

| Version | Runtime | Maximum RSS |
| --- | ---: | ---: |
| 7.15.0 | 30.00s | 197 MiB |
| 7.15.1 | 25.79s | 1,182 MiB |

This is approximately a 6x increase in peak RSS between 7.15.0 and 7.15.1.

This was originally causing a CI runner on our project to hang due to it being OOM (we observed a 5x increase there.

The excessive memory usage occurs during the `coverage json` command.

**To Reproduce**

1. What version of Python are you using?

   CPython 3.10.17.

2. What version of coverage.py shows the problem?

   Coverage.py 7.15.1 shows the regression. Coverage.py 7.15.0 does not.

   Relevant output from `coverage debug sys`:

   ```text
   coverage_version: 7.15.1
   core: -none-
   CTracer: available
   configs_read: pyproject.toml
   python: 3.10.17
   platform: macOS-26.5.1-arm64-arm-64bit
   implementation: CPython
   ```

3. What versions of what packages do you have installed?

   The reproducer only requires Coverage.py and its TOML extra:

   ```text
   coverage==7.15.1
   tomli==2.0.1
   ```

4. What code shows the problem?

   The attached `coveragepy-json-memory-repro.zip` is a self-contained, synthetic reproducer. It generates:

   ```text
   4,114 synthetic Python modules
   744,634 recorded arcs
   a 26 MiB coverage database
   a 27 MiB JSON report
   ```

   It contains no application or third-party source code.

5. What commands should we run to reproduce the problem?

   ```bash
   mkdir coveragepy-json-memory-repro
   unzip coveragepy-json-memory-repro.zip -d coveragepy-json-memory-repro
   cd coveragepy-json-memory-repro

   python3.10 -m venv .venv
   source .venv/bin/activate

   python -m pip install 'coverage[toml]==7.15.0'
   python make_reproducer.py
   COVERAGE_FILE=.coverage \
[...]

PR #2230, opened 2026-07-13
fix: bound reporting caches and release parser ASTs (issue 2229)

reaperhulk:
(The usual disclaimer: this PR is agent generated with human feedback and review)

Version 7.15.1 regressed memory during reporting: the whole-phase caches added in #2215 retained every file's FileReporter and Analysis, and each cached PythonFileReporter pinned its parser's AST, so a 1,500-module branch corpus peaked at 555 MB vs 98 MB on 7.15.0. 

PythonParser now releases its AST once _analyze_ast() has memoized everything derived from it, and the analysis and file reporter caches are bounded with oldest-first eviction, so modest projects keep the full repeated-sweep speedup while large projects get bounded memory; with both changes the corpus peaks at 127 MB.

The bound is configurable as the new [report] analysis_cache_size setting or the --analysis-cache-size flag on reporting commands, defaulting to 256 files.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,19 @@ upgrading your version of coverage.py.
 
 .. start-releases
 
+Unreleased
+----------
+
+- Fix: reporting on projects with many files used far more memory in 7.15.1
+  than in 7.15.0, as described in `issue 2229`_.  The reporting caches
+  introduced in 7.15.1 are now bounded, and parsers release their ASTs once
+  arc analysis is complete.  The size of the caches can be adjusted with the
+  new :ref:`analysis_cache_size <config_report_analysis_cache_size>` setting,
+  or the ``COVERAGE_ANALYSIS_CACHE_SIZE`` environment variable, which defaults
+  to 256 files.
+
+.. _issue 2229: https://github.com/coveragepy/coveragepy/issues/2229
+
 .. _changes_7-15-1:
 
 Version 7.15.1 — 2026-07-12
--- a/coverage/config.py
+++ b/coverage/config.py
@@ -218,6 +218,7 @@ class CoverageConfig(TConfigurable, TPluginConfig):
         self._crash: str | None = None
 
         # Defaults for [report]
+        self.analysis_cache_size = 256
         self.exclude_list = DEFAULT_EXCLUDE[:]
         self.exclude_also: list[str] = []
         self.fail_under = 0.0
@@ -428,6 +429,7 @@ class CoverageConfig(TConfigurable, TPluginConfig):
         ("_crash", "run:_crash"),
         #
         # [report]
+        ("analysis_cache_size", "report:analysis_cache_size", "int"),
         ("exclude_list", "report:exclude_lines", "regexlist"),
         ("exclude_also", "report:exclude_also", "regexlist"),
         ("fail_under", "report:fail_under", "float"),
@@ -705,6 +707,11 @@ def read_coverage_config(
     env_data_file = os.getenv("COVERAGE_FILE")
     if env_data_file:
         config.data_file = env_data_file
+
+    # $set_env.py: COVERAGE_ANALYSIS_CACHE_SIZE - Max files held in the reporting caches.
+    env_cache_size = os.getenv("COVERAGE_ANALYSIS_CACHE_SIZE")
+    if env_cache_size:
+        config.analysis_cache_size = int(env_cache_size)
 
     # $set_env.py: COVERAGE_DEBUG - Debug options: https://coverage.rtfd.io/cmd.html#debug
     debugs = os.getenv("COVERAGE_DEBUG")
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -1017,8 +1017,14 @@ class Coverage(TConfigurable):
         file_reporter = self._get_file_reporter(morf)
         filename = self._file_mapper(file_reporter.filename)
         analysis = analysis_from_file_reporter(data, self.config.precision, file_reporter, filename)
-        self._analysis_cache[morf] = analysis
+        self._cache_put(self._analysis_cache, morf, analysis)
         return analysis
+
+    def _cache_put(self, cache: dict[TMorf, Any], morf: TMorf, value: Any) -> None:
+        """Add to a bounded cache, evicting the oldest entry if it's full."""
+        cache[morf] = value
+        if len(cache) > self.config.analysis_cache_size:
+            del cache[next(iter(cache))]
 
     def _clear_analysis_caches(self) -> None:
         """Forget cached analyses and file reporters, because data changed."""
@@ -1068,7 +1074,7 @@ class Coverage(TConfigurable):
             file_reporter = PythonFileReporter(morf, self)
 
         assert isinstance(file_reporter, FileReporter)
-        self._file_reporter_cache[morf] = file_reporter
+        self._cache_put(self._file_reporter_cache, morf, file_reporter)
         return file_reporter
 
     def _get_file_reporters(
--- a/coverage/parser.py
+++ b/coverage/parser.py
@@ -336,8 +336,10 @@ class PythonParser:
         `_all_arcs` is the set of arcs in the code.
 
         """
-        assert self._ast_root is not None
-        aaa = AstArcAnalyzer(self.filename, self._ast_root, self.raw_statements, self.multiline_map)
+        ast_root = self._ast_root
+        if ast_root is None:
+            ast_root = ast.parse(self.text)
+        aaa = AstArcAnalyzer(self.filename, ast_root, self.raw_statements, self.multiline_map)
         aaa.analyze()
         arcs = aaa.arcs
         self._with_jump_fixers = aaa.with_jump_fixers()
@@ -352,6 +354,10 @@ class PythonParser:
                 self._all_arcs.add((fl1, fl2))
 
         self._missing_arc_fragments = aaa.missing_arc_fragments
+
+        # The AST is large and no longer needed: everything derived from it is
+        # memoized above.  Release it so long-lived parsers don't hold it.
+        self._ast_root = None
 
     def fix_with_jumps(self, arcs: Iterable[TArc]) -> set[TArc]:
         """Adjust arcs to fix jumps leaving `with` statements.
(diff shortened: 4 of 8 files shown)


nedbat:
A command-line option seems like a lot for something this rarely used. Maybe an environment variable would be enough.

For my own experiments, can you give a quick synopsis of how to run the cryptography suite to see these factors in action?

reaperhulk:
To be honest, I'm not sure any configurability at all is necessary beyond "cache on/cache off"? Cache off would effectively replicate the previous behavior (where the lru cache was getting overwritten per-file) and with the parser AST change we should actually have lower peak memory.

You can run the cryptography test suite with coverage using `nox -e tests` (`nox -e local` is what we use locally, but that runs additional rust and linting tests). The memory numbers above weren't derived from that though, the agent created a temporary 1500 module corpus to force a worst-case for memory consumption.

Edit: You will need a bunch of deps installed for cryptography unfortunately, including both a C and rust compiler along with openssl.

harrismirza (approved):

devdanzin:
I tested this PR against `.coverage` databases produced by real PyPI package test suites (gathered while hunting reproducers for #2229 — details in that issue). Method: `coverage run --branch` on each suite with 7.15.0, then `coverage json -i` on the same database under released 7.15.0, released 7.15.1, and this PR (`6c68acb`), on CPython 3.14.4/Linux, peak RSS via `/usr/bin/time -v`.

| dataset                         | files | RSS 7.15.0 | RSS 7.15.1 | RSS PR      | t 7.15.0 | t 7.15.1 | t PR     |
| ------------------------------- | ----: | ---------: | ---------: | ----------: | -------: | -------: | -------: |
| cloudflare-python (import-only) | 5,346 | 1,092 MiB  | 2,790 MiB  | **977 MiB** | 3:21     | 2:52     | **2:48** |
| twisted (import-only)           | 760   | 1,200      | 1,905      | **1,034**   | 1:46     | 1:29     | **1:19** |
| xlsxwriter                      | 1,323 | 782        | 963        | **811**     | 0:30     | 0:26     | **0:24** |
| django-allauth                  | 1,239 | 732        | 886        | **707**     | 0:23     | 0:21     | **0:20** |
| auth0-python                    | 1,243 | 766        | 878        | **738**     | 0:28     | 0:25     | **0:21** |
| databricks-sdk                  | 132   | 1,392      | 1,354      | **1,000**   | 0:57     | 0:47     | **0:44** |
| dynaconf                        | 204   | 651        | 622        | **599**     | 0:07     | 0:06     | **0:06** |

- **Fixes the regression**: the worst leaker (cloudflare, 5,346 files) drops from 2,790 MiB back to 977 MiB — below the 7.15.0 baseline.
- **No regression below the cache bound**: databricks-sdk (132 files) and dynaconf (204 files) stay under the default 256-file limit and both *improve* — databricks-sdk by 392 MiB even vs 7.15.0, presumably from the AST release, since its 132 files are huge generated modules.
- **7.15.1's speed gains are retained**: the PR matches or beats 7.15.1 wall time on every dataset.

The PR beating 7.15.0 on memory in six of seven datasets (the exception is within RSS noise) suggests the AST pinning predates 7.15.1 and the new caches only amplified it.

*Comparisons and comment draft helped by Claude*

devdanzin:
## Python version effect (cloudflare collect-only DB, same .coverage file)

| Python | RSS 7.15.0 | RSS 7.15.1 | RSS PR 2230 | ratio | t 7.15.0 | t 7.15.1 | t PR |
|---|---:|---:|---:|---:|---:|---:|---:|
| 3.10.18 | 231 MiB | 922 MiB | **229 MiB** | **4.0x** | 0:41 | 0:39 | **0:37** |
| 3.14.4 | 1,092 MiB | 2,790 MiB | **977 MiB** | 2.6x | 3:21 | 2:53 | **2:48** |

1. The regression *ratio* is larger on 3.10 (4.0x vs 2.6x) because the
   baseline is far smaller; the absolute leak is larger on 3.14
   (+1,698 MiB vs +691 MiB). On 3.10 the real cloudflare data approaches
   the synthetic reproducer's 6.2x (also measured on 3.10).
2. PR 2230 restores the 7.15.0 baseline on both versions (229 and
   977 MiB).
3. Side observation, independent of the regression: the same JSON report
   costs ~5x more memory and ~5x more wall time on CPython 3.14.4 than on
   3.10.18 in *both* released versions (231 -> 1,092 MiB, 0:41 -> 3:21 for
   7.15.0). Possibly worth its own investigation.

devdanzin:
I re-tested this PR against `.coverage` databases produced by real PyPI package test suites (gathered while hunting reproducers for #2229). Method: `coverage run --branch` on each suite with 7.15.0, then `coverage json -i` on the same database under released 7.15.0, released 7.15.1, and this PR (`6c68acb`), on CPython 3.14.3 (python-build-standalone release, Linux), peak RSS via `/usr/bin/time -v`.

*(Note: an earlier version of these measurements was taken on a pydebug+ASan interpreter, which inflated baselines and understated the regression ratios — the numbers below supersede them.)*

| dataset                         | files | RSS 7.15.0 | RSS 7.15.1 | RSS PR      | t 7.15.0 | t 7.15.1 | t PR     |
| ------------------------------- | ----: | ---------: | ---------: | ----------: | -------: | -------: | -------: |
| cloudflare-python (import-only) | 5,346 | 236 MiB    | 1,146 MiB  | **219 MiB** | 0:27     | 0:29     | **0:23** |
| twisted (import-only)           | 760   | 218        | 627        | **179**     | 0:12     | 0:13     | **0:10** |
| django-allauth                  | 1,239 | 65         | 144        | **54**      | 0:03     | 0:03     | **0:03** |
| xlsxwriter                      | 1,323 | 101        | 204        | **98**      | 0:04     | 0:04     | **0:04** |
| auth0-python                    | 1,243 | 122        | 175        | **99**      | 0:04     | 0:04     | **0:03** |
| databricks-sdk                  | 132   | 378        | 392        | **208**     | 0:09     | 0:07     | **0:06** |
| dynaconf                        | 204   | 65         | 68         | **47**      | 0:01     | 0:01     | **0:01** |

- **Fixes the regression**: the worst real-world case (cloudflare, 5,346 files, 4.9× blowup on 7.15.1) drops from 1,146 MiB to 219 MiB — below the 7.15.0 baseline. Same DB on CPython 3.10.18: 231 → 921 → **229 MiB**, so the fix holds across versions.
- **No regression below the cache bound**: databricks-sdk (132 files) and dynaconf (204 files) stay under the default 256-file limit and both improve — databricks-sdk by nearly half even vs 7.15.0, presumably from the AST release, since its files are huge generated modules.
- **7.15.1's speed gains are retained**: the PR is the fastest of the three on every dataset.
- The reporter's synthetic case (4,114 files) also returns to baseline: 189 → 1,167 → **195 MiB** (py3.10).

One observation from comparing versions: the leaked per-file objects are ~30% larger on 3.14 than 3.10 (same-source `ast.parse` trees measure +49–63% bigger under tracemalloc), so the unpatched leak grows with newer Pythons — one more reason this fix is worth landing.

devdanzin:
Analyzing how the caching works with Claude, the conclusion is that FIFO is pessimal for the current access pattern, and that MRU would be a strict improvement. Can you test this diff to see if it improves anything?

```diff
diff --git a/coverage/control.py b/coverage/control.py
index 1a76d812..1285e1c7 100644
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -1024,7 +1024,7 @@ class Coverage(TConfigurable):
         """Add to a bounded cache, evicting the oldest entry if it's full."""
         cache[morf] = value
         if len(cache) > self.config.analysis_cache_size:
-            del cache[next(iter(cache))]
+            cache.popitem()  # evict newest: optimal for cyclic sequential sweeps
 
     def _clear_analysis_caches(self) -> None:
         """Forget cached analyses and file reporters, because data changed."""

```

nedbat:
To sum up(?): each file will be processed twice, but all files in the same order. So a cache only helps if it is larger than the number of files. Running against the 4114-file repro from #2229, completely disabling the cache doesn't change the running time.

I also tried using the repo from #2229 to make a 250-file version, and the cache didn't seem to help there either? Maybe I'm not doing a good job tracking the experiments.

devdanzin:
That is expected because the cache is invisible in any single-report run.

1. The expensive call, `_analyze`, runs once per file per report run. So within one `coverage json` invocation there are zero repeated analysis requests, and no cache of any size can change its running time.
2. The caches only pay off on the second and later report sweeps in the same process, the pattern #2215 was written for (pytest-cov's total pass + term report, or term+xml+html). Two CLI invocations are two processes, so nothing carries over there either: the experiment has to go through the API.

Concrete experiment, run in the #2229 reproducer directory:

```python
import time, coverage
cov = coverage.Coverage()
cov.load()
for i in (1, 2, 3):
    t0 = time.time()
    cov.json_report(outfile=f"sweep{i}.json")
    print(f"sweep {i}: {time.time()-t0:.1f}s")
```

On my machine against the 4,114-file repro, with 7.15.1 with its unbounded cache:

|        | sweep 1 | sweep 2  | sweep 3  |
| ------ | ------: | -------: | -------: |
| 7.15.0 | 44.3s   | 45.0s    | 46.2s    |
| 7.15.1 | 38.8s   | **7.8s** | **9.4s** |

That ~5× on warm sweeps is the entire benefit the caches ever deliver: 7.15.0's flat line is why we concluded its `maxsize=1` caches never worked for reporting (the two reporter requests are N calls apart so a size-1 cache can't bridge them).

Predictions for this PR's build, same script: on the 4,114-file repro, sweeps 2–3 stay slow (bound 256 < 4,114 files; FIFO evicts each entry before its next use) unless you raise `COVERAGE_ANALYSIS_CACHE_SIZE=5000`, which restores the fast warm sweeps at far less memory than 7.15.1 since the ASTs are released. On your 250-file version, now under the 256 default, warm sweeps should be fast out of the box.

*Analysis and draft by Claude*

nedbat:
See #2231 for a better approach.

reaperhulk:
That is definitely a better approach. On cryptography it’s both faster and uses less memory. For scenarios where a second pass may occur for additional reporting it doesn’t cache, but it appears that it’s faster even in this case. Closing in favor of yours 😄

PR #2231, opened 2026-07-15
fix: reduce memory use during reporting. #2229

nedbat:
Reduce FileReporter memory usage. No need for caches. This is both faster and smaller than 7.15.0. Using the reproducer from #2229:
- 7.15.0: 35.19s,   206Mb mem
- 7.15.1: 33.65s, 1,243Mb mem
- this PR: 26.38s, 160Mb mem

This approach drops objects that aren't needed anymore to reduce the memory footprint, and uses existing FileReporters where we already have them rather than re-creating them.

I thought about not making a complete list of FileReporters, but they are small at first. If we think that could be a benefit, that can come later.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,7 +23,12 @@ upgrading your version of coverage.py.
 Unreleased
 ----------
 
-Nothing yet.
+- Fix: one of the performance improvements in 7.15.1 (pull 2215) dramatically
+  increased memory use during reporting for large projects. Now we use a
+  different approach that is both faster and slimmer than 7.15.0. Fixes `issue
+  2229`_.
+
+.. _issue 2229:  https://github.com/coveragepy/coveragepy/issues/2229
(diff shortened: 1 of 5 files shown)


devdanzin:
Nice, we converged into very similar solutions :)

It's important to notice that with no caches, the multi-sweep  (pytest-cov term+xml+html) speedup from #2215 is gone: warm sweeps cost the same as cold (we measured 6s -> 21s per extra sweep on  cloudflare-co, 2s -> 6s on databricks vs 7.15.1). Given sweep 1  dominates a triple-report, totals remain within ~2x of 7.15.1's best  case and equal-or-better than 7.15.0 everywhere. But our would solution also get rid of caches, so same result.

nedbat:
This is now released as part of [coverage 7.15.2](https://pypi.org/project/coverage/7.15.2).

PR #2238, opened 2026-07-21
Add benchmarks for coverage.py performance paths

KRRT7:
## Summary

Add an opt-in pytest-benchmark suite for coverage.py's tracing, data handling, and reporting costs, alongside a pinned Jinja2 test-suite workload. Each case validates the work performed outside the timed region.

The suite covers:

- Line and branch tracing with Python, C, and sys.monitoring cores across loops, calls, imports, generators, and async execution.
- Dynamic contexts, dense and sparse database writes, parallel-data combination, path remapping, unused-source discovery, and multiprocessing.
- Analysis and text, HTML, XML, JSON, and LCOV reporting over mixed coverage data, including context filtering and incremental HTML updates.
- Tokenization of frozen source snapshots, with hashes and workload versions for comparable results.
- Jinja2 3.1.6's 909 tests, uninstrumented and under each coverage core, plus separate reporting and fresh-process peak-memory measurements.

Assertions check core selection, context membership, combined data, worker coverage, regenerated HTML pages, and matching real-project test identities.

## Usage

```bash
make bench
make bench ARGS='-k html --benchmark-autosave'

# Validate component workloads without collecting timings:
python3 -m pytest -n0 --benchmarks --bench-smoke \
    -m 'benchmark and not real_project' tests/benchmarks

# Prepare and run the real-project workload with Python 3.14:
make bench-prepare
make bench-real
```

Benchmarks require explicit opt-in and serial execution. Real-project preparation downloads verified inputs and installs separately pinned dependencies before measurement. Setup-dependent cases use ten measured rounds, or five for slow cases, adjustable with `--bench-rounds`.

See the [benchmark README](https://github.com/KRRT7/coveragepy/blob/asv-bench/tests/benchmarks/README.md) for workload sizes, timing boundaries, and comparison guidance.

## CI and validation

Relevant PRs run component timings on Linux Python 3.14 and smoke checks on Linux Python 3.10/3.15 and macOS Python 3.14. A weekly/manual job runs the real-project and slow workloads. JSON artifacts and summaries report timing variability, sample counts, workload metadata, and peak RSS where available. Results are informational, without a regression threshold.

Validation completed:

- Local timing runs: 64 component cases and 16 extended cases passed.
- Component and real-project smoke runs passed, including verification of all 909 Jinja2 test identities.
- Targeted regression and testing/core checks: 59 passed, 1 skipped. Workload regressions also passed under metacoverage with both C and Python tracers.
[...]

PR #2239, opened 2026-07-21
perf: speed up SQLite numbits aggregation

KRRT7:
## Summary

Optimize SQLite numbits aggregation by accumulating packed bitsets as integers instead of repeatedly creating intermediate byte strings.

The aggregate now converts the final integer back to bytes only once during finalization.

## Impact

This reduces Python-level work when combining many numbits values, especially during coverage data combination and reporting.

While reviewing the SQLite data-combination path, we found that NumbitsUnionAgg repeatedly called numbits_union() for every row.

Each call iterated over the byte strings and allocated a new byte string for the intermediate result. For aggregations over many rows, this caused repeated Python-level work and memory allocations.

The optimization keeps the accumulated bitset as a Python integer and combines each value with a bitwise OR. The result is converted back to the packed byte representation only once, during finalization. The original byte length is tracked so the serialized result preserves the expected representation, including empty aggregates.

this was found using a proprietary AI agent.


devdanzin:
Nice! Do you have (micro?)benchmarks showing the effect of this change?

KRRT7:
> https://github.com/coveragepy/coveragepy/pull/2239#issuecomment-5033553737

Hi Daniel, big fan of your work :) 

I'm not sure if I still have the benchmarks still, but I'll check.

KRRT7:
Hi Daniel — I no longer have the original standalone microbenchmarks, but I reran the comparison using the ASV benchmark suite from PR #2238.

The relevant benchmark is `TimeCombine`, which exercises SQLite coverage-data combination:

```text
main:
  time_combine_many_files_line_only       136 ± 0.7 ms
  time_combine_many_files_with_contexts   209 ± 10 ms

perf/sqlite-numbits:
  time_combine_many_files_line_only       120 ± 2 ms
  time_combine_many_files_with_contexts   193 ± 10 ms
```

That’s approximately 12% faster for line-only combination and 8% faster with contexts. The contexts result is less conclusive because the measurement variance overlaps.

The benchmark suite is in PR #2238. To reproduce the comparison:

```bash
git fetch origin pull/2238/head:asv-bench
git fetch origin pull/2239/head:perf/sqlite-numbits

git worktree add /tmp/coverage-main origin/main
git worktree add /tmp/coverage-sqlite perf/sqlite-numbits

git -C /tmp/coverage-main restore --source=asv-bench --worktree -- asv.conf.json benchmarks
git -C /tmp/coverage-sqlite restore --source=asv-bench --worktree -- asv.conf.json benchmarks

cd /tmp/coverage-main
uvx --from asv asv run --bench TimeCombine 'origin/main^!'

cd /tmp/coverage-sqlite
uvx --from asv asv run --bench TimeCombine 'perf/sqlite-numbits^!'
```

If `asv` is already installed, `asv run` can be used directly instead of `uvx --from asv asv run`.

devdanzin:
Great result, thank you!

nedbat:
Thanks!

nedbat:

PR #2240, opened 2026-07-21
perf: skip AST parsing without soft keywords

KRRT7:
## Summary

Avoid parsing the full AST when looking for soft-keyword lines if the source contains no possible soft keywords.

The fast path checks for `match`, `case`, `type`, and `lazy` before calling `ast.parse`, while preserving the existing version-specific handling for each keyword.

## Impact

Most Python files do not use soft keywords. Skipping the AST parse for those files reduces tokenization overhead during reporting.

While reviewing the physical-tokenization path, we found that `find_soft_key_lines()` parsed every source file even when the file could not contain a soft keyword. A cheap source-text check can identify that common case before doing any AST work.


nedbat:
I changed the match/case check so that it would only parse the file if both "match" and "case" were found in the file. A check across some random source trees:
```
162711 .py files
81574 use "or": 50% will be parsed
95486 use "and", 59% will be parsed
```

nedbat:
BTW, I also tried a regex check for complete words, but it only reduced to 48% of files being parsed.
```
re.search(r"\b(match|case|type|lazy)\b", source)
```

nedbat:

PR #2241, opened 2026-07-21
perf: optimize physical tokenization hot path

KRRT7:
## Summary

Optimize the hot path used by `source_token_lines()`.

The implementation:

- Emits lightweight token tuples instead of `tokenize.TokenInfo` objects.
- Avoids repeated string splitting for single-line tokens.
- Uses faster newline handling for multiline tokens.
- Caches frequently used helper functions and configuration values.
- Updates parser annotations for the lightweight token representation.

## Impact

Benchmarked against real coverage.py source files on Python 3.14:

- `coverage/phystokens.py`: 17% faster
- Parser and reporting files: 21% faster
- Entire `coverage/` package: 21% faster

Coverage behavior is unchanged.

While reviewing the physical-tokenization path, we found repeated allocations and string-processing work in `_phys_tokens()` and `source_token_lines()`.

The hot path created `tokenize.TokenInfo` objects, split token text repeatedly, and looked up the same helper functions and configuration values for each token. These operations were replaced with lightweight tuples, direct string handling, and cached local references.


nedbat:
It would be helpful to provide measurement of the performance impact for a complete coverage command like `coverage html`.

nedbat:
The code seems to duplicate a large swath of code, which I don't like. Is this necessary?

PR #2242, opened 2026-07-23
perf: reuse SQLite connection during HTML reports

KRRT7:
## Summary

Reuse the SQLite connection while generating an HTML coverage report.

Previously, each database context used during reporting closed and reopened the
connection. HTML reporting performs many read queries, causing significant
connection setup overhead.

This change keeps the connection open for the duration of `coverage html` and
force-closes it when reporting completes. Normal connection lifecycle behavior
outside report generation is unchanged.


KRRT7:
## Reproducing the performance difference

Requires Python 3.14, [uv](https://docs.astral.sh/uv/), and [hyperfine](https://github.com/sharkdp/hyperfine).

```bash
git clone https://github.com/coveragepy/coveragepy.git
cd coveragepy

# Fetch this PR as a local branch.
git fetch origin pull/2242/head:sqlite-bench

uv venv .venv --python 3.14
uv pip install \
  --python .venv/bin/python \
  -r requirements/pip.pip \
  -r requirements/pytest.pip \
  -r requirements/light-threads.pip \
  -e .

export PATH="$PWD/.venv/bin:$PATH"

# Generate one shared coverage dataset on main.
git switch main
python setup.py --quiet build_ext --inplace
COVERAGE_CORE=ctrace coverage erase
COVERAGE_CORE=ctrace coverage run -m pytest -n0 \
  -k 'not test_current and not test_debug_callers and not test_short_stack and not test_short_stack_skip'

# Benchmark main.
hyperfine --warmup 5 --runs 20 \
  'coverage html --ignore-errors -d /tmp/coverage-html-main'

# Switch source only. Keep .coverage unchanged.
git switch sqlite-bench

# Benchmark the PR.
hyperfine --warmup 5 --runs 20 \
  'coverage html --ignore-errors -d /tmp/coverage-html-pr'
```

The same `.coverage` file is used for both runs; only the coverage.py source code changes between measurements.

nedbat:
What performance impact did you see?

KRRT7:
> What performance impact did you see?

`main`: 1.154s ± 0.015s
this branch: 0.846s ± 0.010s
 so 27% improvement.

KRRT7:
looks like GitHub didn't render my original comment properly, editing it now.

edit: actually it's a known codex issue (at least on my end) when posting comments.

reaperhulk:
Speaking strictly on the perf side: The potential performance improvement of this will be heavily dependent on details of the codebase. For example, if you set paths in the config then all the data is copied into an in-memory SQLite DB, where close/open are no-ops:
https://github.com/coveragepy/coveragepy/blob/e964efdc1807d9dc19c3df458158382c371d2075/coverage/control.py#L1081-L1082

However, this does not appear to hurt the perf in that case (it just doesn't matter).

yangfan-yf-yf (changes requested):
The contributing guide expects tests for new functionality and fixes. This PR changes the `SqliteDb` connection lifecycle used by `Coverage.html_report()`, but the diff adds no regression coverage for that behavior.

Please add a focused test using a real temporary data file. It should verify that HTML reporting keeps one physical SQLite connection across its reporting queries and force-closes that disk-backed connection when the report finishes. An error-path assertion would also protect the `finally` cleanup.

I ran the existing HTML suite and exercised the new lifecycle locally: it reused one disk connection for a report and left the connection closed afterward. That is useful validation, but the behavior needs a committed test so a later context-manager refactor cannot silently reintroduce connection churn or a handle leak.

nedbat:
@yangfan-yf-yf: I really appreciate you taking a look at this pull request and trying it out. I can see that you've also been validating and commenting on other pull requests.  This is very helpful, thanks.

> The contributing guide expects tests ... Please add a focused test ...

The way you've worded these message though, it sounds like you are a decision-maker for this repo. It's better to take a collaborative tone than a prescriptive one. Do you have ideas or examples about how such a test could be written?

nedbat:
@KRRT7 The latest commits added here seem to be about something other than re-using SQLite connections. Am I misunderstanding?

KRRT7:
> @KRRT7 The latest commits added here seem to be about something other than re-using SQLite connections. Am I misunderstanding?

You're not, but I noticed a small perf regression for smaller projects so I'm trying to recover that loss via other means, will create a stacked for this in a moment

KRRT7:
Unless you're fine with me continuing the work here and publishing the full report when I'm done.

nedbat:
> Unless you're fine with me continuing the work here and publishing the full report when I'm done.

It's better for a pull request to be one idea. If this pull request isn't ready to be reviewed, mark it as draft.

KRRT7:
> It's better for a pull request to be one idea.

it's part of the same idea, the changes came after I ran coverage over a few popular OSS repos to see if there were any regressions, it was a follow up to:

> The potential performance improvement of this will be heavily dependent on details of the codebase.

yangfan-yf-yf:
You're right — I should have phrased it as a suggestion, not a request.

A concrete test idea: in `tests/test_html.py`, use `HtmlTestHelpers` to create and save ordinary disk-backed coverage data, then instantiate a fresh `Coverage` object for the report. Wrap `coverage.sqlitedb.sqlite3.connect` (or the equivalent connection-opening boundary) only while calling `html_report()`. The test could assert that the report uses one physical connection, then assert that the report data object's `_dbs` cache is cleared afterward. That would cover both the intended reuse and post-report cleanup without depending on the specific `keep_db_open` implementation.

I meant this only as the focused example I should have included in my earlier review. Thanks for the guidance.

yangfan-yf-yf (commented):
Thanks for adding both the normal and exception-path lifecycle tests. I reran the updated head with `COVERAGE_CORE=pytrace`: both new tests pass, along with `tests/test_html.py` (68 passed) and `tests/test_data.py` (99 passed).

The remaining failing check is strict mypy at `tests/test_html.py:622`. Because `connect_and_remember()` forwards `*args: Any, **kwargs: Any` to the overloaded `sqlite3.connect`, mypy infers `Any` and reports `[no-any-return]`. An explicit `cast(sqlite3.Connection, connect(*args, **kwargs))` (with `cast` imported from `typing`), or an equivalently typed wrapper, should keep the helper narrow without changing the production patch.

The added lifecycle coverage addresses the gap I raised. After the type check is green, I'd be happy to take another look at the final head.

nedbat:
I'm still trying to understand the real-world implications of this change. What kinds of projects or HTML reports re-open the database many times? Can you show specific examples?

PR #2243, opened 2026-07-26
perf: reuse parser results during HTML reporting

KRRT7:
## Summary

Reuse parser-derived data during HTML reporting to avoid redundant AST parsing and reduce parser memory retention.

This branch is stacked on `perf/sqlite-connection-reuse`.

## Changes

- Reuse the parsed AST to discover function and class code regions.
- Cache discovered regions per Python file reporter.
- Release each parser AST once region extraction is complete.
- Reuse parser-derived soft-keyword line information during HTML tokenization.
- Preserve standalone tokenization and plugin reporter behavior.

## Performance

On the local coverage.py HTML workload:

- Before: 864.3 ms
- After: 820.4 ms
- Improvement: approximately 5.1%

The profile also reduced cumulative `source_token_lines()` time from approximately 0.476 s to 0.353 s.

### DIFF
--- a/coverage/control.py
+++ b/coverage/control.py
@@ -1252,7 +1252,9 @@ class Coverage(TConfigurable):
             precision=precision,
         ):
             reporter = HtmlReporter(self)
-            return reporter.report(morfs)
+            assert self._data is not None
+            with self._data.keep_db_open():
+                return reporter.report(morfs)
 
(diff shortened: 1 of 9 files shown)


KRRT7:
2242 should be reviewed / merged first In order to make this easier to understand

PR #2253, opened 2026-08-06
Fix: local plugins are importable for erase and combine

### LINKED ISSUE #1726: `Erase` fails with local plugins due to incorrect path
**Describe the bug**
When executing `run`, `PyRunner` monkey-patches the path so that the directory of the file being run is the first entry in the path instead of the directory of the `coverage` script.  This change allows for plugins to be imported from the local directory structure.  When executing `erase` this monkey-patching doesn't happen and any local plugins fail to load and `erase` exits due to a `ModuleNotFoundError` when trying to import the plugin module.

**To Reproduce**
Python/Coverage versions 3.8/6.3.2 & 3.10/7.4.0.
Empty virtual environment (I happen to be using Pipenv, but see nothing specific to Pipenv related to this issue) with only `coverage` installed.

1. `mkdir sandbox && cd sandbox`
1. `pipenv install coverage`
1. Create a Coverage Plugin in the local directory structure:
    ```
    # ./plugins/ExamplePlugin.py

    import coverage

    class ExamplePlugin(coverage.CoveragePlugin):
        def configure(self, config):
            print('Do Nothing')

    def coverage_init(reg, options):
        reg.add_configurer(ExamplePlugin())
    ```
1. Create a file to run:
    ```
    # ./app.py

    print('The App')
    ```
1. Create coverage config file:
    ```
    # ./setup.cfg

    [coverage:run]
    omit = plugins/*
    source = .
    plugins = plugins.ExamplePlugin
    ```
1. `pipenv run coverage run app.py`
1. `pipenv run coverage erase`

Results in:

```
  File "/home/kdickerson/.local/share/virtualenvs/sandbox-ErqkbyCg/bin/coverage", line 8, in <module>
    sys.exit(main())
  File "/home/kdickerson/.local/share/virtualenvs/sandbox-ErqkbyCg/lib/python3.10/site-packages/coverage/cmdline.py", line 970, in main
    status = CoverageScript().command_line(argv)
  File "/home/kdickerson/.local/share/virtualenvs/sandbox-ErqkbyCg/lib/python3.10/site-packages/coverage/cmdline.py", line 677, in command_line
    self.coverage.erase()
  File "/home/kdickerson/.local/share/virtualenvs/sandbox-ErqkbyCg/lib/python3.10/site-packages/coverage/control.py", line 701, in erase
    self._init()
  File "/home/kdickerson/.local/share/virtualenvs/sandbox-ErqkbyCg/lib/python3.10/site-packages/coverage/control.py", line 341, in _init
    self._plugins = Plugins.load_plugins(self.config.plugins, self.config, self._debug)
  File "/home/kdickerson/.local/share/virtualenvs/sandbox-ErqkbyCg/lib/python3.10/site-packages/coverage/plugin_support.py", line 55, in load_plugins
    __import__(module)
ModuleNotFoundError: No module named 'plugins'
```

**Expected behavior**
`erase` should run as expected and remove the data file(s).

**Additional context**
It works if you perform the same monkey-patching before `erase` runs by changing `CoverageScript.command_line` in `cmdline.py:L676-678`  from:
```
        elif options.action == "erase":
            self.coverage.erase()
            return OK
```
to
```
        elif options.action == "erase":
            runner = PyRunner(['dummy.txt'], as_module=bool(options.module))
            runner.prepare()
            self.coverage.erase()
            return OK
```

But I'm highly doubtful that's the _right_ solution.

Eljees:
Closes #1726.

cmdline.py inserts "" into sys.path with the comment that plugins may need to import from the current directory, but the insert sits below the erase and combine branches, both of which return before reaching it. Both commands construct Coverage, which loads configured plugins, so a plugin living in the project directory raises ModuleNotFoundError.

The report in issue 1726 is about erase; combine is broken by the same defect, which matters more - it is the core of parallel mode.

Only the console script is affected: python -m coverage puts the CWD on sys.path itself.

Move the block above the action dispatch, keeping run excluded: there sys.path[0] is owned by PyRunner.prepare(), and run never reached the old line anyway, so its behaviour is unchanged.

### DIFF
--- a/coverage/cmdline.py
+++ b/coverage/cmdline.py
@@ -852,6 +852,12 @@ class CoverageScript:
             messages=not options.quiet,
         )
 
+        # We need to be able to import from the current directory, because
+        # plugins may try to, for example, to read Django settings.  The "run"
+        # command is excluded: PyRunner.prepare() owns sys.path[0] there.
+        if options.action != "run":
+            sys.path.insert(0, "")
+
         if options.action == "debug":
             return self.do_debug(args)
 
@@ -878,10 +884,6 @@ class CoverageScript:
             include=include,
             contexts=contexts,
         )
-
-        # We need to be able to import from the current directory, because
-        # plugins may try to, for example, to read Django settings.
-        sys.path.insert(0, "")
 
         self.coverage.load()
         self.coverage.combine(strict=False, keep=bool(options.keep_combined))
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -314,6 +314,37 @@ class PluginTest(CoverageTest):
         out = self.run_command("coverage html -q")  # sneak in a test of -q
         assert out == ""
 
+    def test_local_files_are_importable_for_erase_and_combine(self) -> None:
+        # https://github.com/coveragepy/coveragepy/issues/1726
+        self.make_file(
+            "importing_plugin.py",
+            """\
+            from coverage import CoveragePlugin
+            import local_module
+            class MyPlugin(CoveragePlugin):
+                pass
+            def coverage_init(reg, options):
+                reg.add_noop(MyPlugin())
+            """,
+        )
+        self.make_file("local_module.py", "CONST = 1")
+        self.make_file(
+            ".coveragerc",
+            """\
+            [run]
+            plugins = importing_plugin
+            parallel = True
+            """,
+        )
+        self.make_file("main_file.py", "print('MAIN')")
+
+        out = self.run_command("coverage run main_file.py")
+        assert out == "MAIN\n"
+        out = self.run_command("coverage combine")
+        assert "ModuleNotFoundError" not in out
+        out = self.run_command("coverage erase")
+        assert out == ""
+
     def test_coverage_init_plugins(self) -> None:
         called = False
 


Eljees:
Closes #1726.

Repro on `master`, straight from the issue — a plugin that is only importable from the current directory works under `run` but not under `combine` or `erase`:

```
$ python -m coverage combine
...
ModuleNotFoundError: No module named 'importing_plugin'
```

`cmdline.py` already had a `sys.path.insert(0, "")`, but it sat *below* the branch that dispatches `erase` and `combine`, so those two commands returned before ever reaching it — the insert was effectively dead code for them. This moves it above the dispatch and skips it for `run`, where `PyRunner.prepare()` owns `sys.path[0]` and inserting again would be wrong.

Test: `test_local_files_are_importable_for_erase_and_combine` in `tests/test_plugins.py`. It fails on master with exactly the `ModuleNotFoundError` above and passes with the fix; the rest of `tests/test_plugins.py` is green either way.

AI-assisted (LLM used for drafting the patch and the test); the analysis and the runs are mine.

Eljees:
Ping — this has been open since 6 August with no review yet.

`cmdline.py` inserts `""` into `sys.path` with a comment saying plugins may need to import from the current directory, but the insert sits below the `erase` and `combine` branches, both of which return before reaching it. Both construct `Coverage`, which loads configured plugins, so a plugin living in the project directory raises `ModuleNotFoundError`. #1726 reports this for `erase`; `combine` is broken by the same defect, which matters more since it is the core of parallel mode. Only the console script is affected — `python -m coverage` puts the CWD on `sys.path` itself.

The fix moves the block above the action dispatch and keeps `run` excluded, where `sys.path[0]` is owned by `PyRunner.prepare()` and the old line was never reached anyway.

Branch is mergeable with no conflicts. Happy to close it if you would rather handle #1726 another way.

Eljees:
This PR had **no CI at all** for its whole life, which I only noticed while auditing my open work. For the head commit `8383cc8` there were zero workflow runs and zero check suites — not a pending approval, simply nothing. The only status on it was Read the Docs.

It was specific to this PR, not the repository. Checking every open PR here:

```
#2252  created 2026-08-06  fork  5 workflow runs      <- opened the same day as mine
#2257  created 2026-08-11  fork  5
#2258  created 2026-08-12  fork  5
#2259  created 2026-08-14  fork  5
#2253  created 2026-08-06  fork  0                    <- this one
```

Every PR opened since Aug 2025 has 4–6 runs; the only other zeros are a handful from 2019–2025 whose last push predates the current workflows. `testsuite.yml`, `quality.yml` and `coverage.yml` all trigger on a bare `pull_request:` with no path filters, and this PR touches `coverage/cmdline.py` and `tests/test_plugins.py`, so nothing in the configuration explains it.

I have merged current `main` into the branch, which gave the PR a new head and the events it never got: **CI is running now.**

While I was there I re-verified the change against `main` at `f108ca89`, since the branch was 55 commits behind:

- on pristine `main` with only my test file applied, the new test fails with `ModuleNotFoundError: No module named 'importing_plugin'` — the bug this fixes;
- with the change, that test and the existing `test_local_files_are_importable` both pass;
- the whole `tests/test_plugins.py` is green on the merged branch (41 passed, 1 skipped) against 40 passed, 1 skipped on `main` — the difference being the added test.

The change itself is unchanged: move the `sys.path.insert(0, "")` so it also covers `erase` and `combine`, and exclude `run`, where `PyRunner.prepare()` owns `sys.path[0]`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015CsUTRwsaFCwavDJTPwEpW

PR #2257, opened 2026-08-11
HTML index: footer Total must agree with the header percentage (fixes #2256)

### LINKED ISSUE #2256: HTML index: footer Total is recomputed in JS with toFixed(), can show 100% when coverage isn't 100%
### Describe the bug

```javascript
cell.textContent = denom
```

event is dispatched on load:

```javascript
```

typed and no checkbox ticked.

### Expected behaviour

> "100".

### Actual behaviour

`precision = 0`:

|---|---|---|---|

beneath it.

### Additionally

to the rounding issue above.)

### Note

### Versions

siaklin99-gif:
Fixes #2256. Thanks for the quick reply on that issue — this is the PR you asked for.

### The change

Two parts, and the first is the one that actually guarantees the stats match:

**1. When nothing is hidden, restore the server's own value instead of recomputing.**
The default page view now does no recomputation at all, so the footer cannot disagree with the `<h1>` — there is no second implementation of the rule running. This also covers `--skip-covered`, where rows are omitted server-side but their statements are still counted in the header total.

**2. When a filter really is hiding rows, clamp the subset total.**
The recompute is still needed there (the header has no value for a subset), so it goes through `display_covered()`, a small mirror of the Python function, and a filtered subset missing one statement out of 250 can no longer render as `100%`.

### Verified in a browser, before and after

I drove the real `coverage_html.js` in Chrome against a fixture table, awaiting the debounced handler (the overwrite is async, which is why this is easy to miss by reading the DOM immediately after load).

| case | before | after |
|---|---|---|
| unfiltered, `<h1>` = `99%` (249/250) | footer **`100%`** ❌ | footer `99%` ✅ |
| filtered to a file that is truly 125/125 | `100%` | `100%` ✅ |
| filtered to 124/125 | `99%` | `99%` ✅ |
| filtered to 249/250 (99.6%) | **`100%`** ❌ | `99%` ✅ |

The genuinely-complete case still reads `100%` — the clamp only applies strictly between the bounds, same as the Python function.

### Two things worth flagging

**The JS test suite does not currently run.** `tests/js/index.html` loads `../../coverage/htmlfiles/jquery.min.js`, which no longer exists in the tree, so every test fails at `beforeEach` with `ReferenceError: $ is not defined` on `main`, before my change. That is presumably why this path has no coverage. I did not touch it here — happy to fix it in a separate PR if that would help, and then add a proper test for this behaviour.

**One deliberate non-change.** I did not reimplement Python's round-half-even in JS. In the unfiltered case nothing is recomputed, so there is nothing to diverge; in the filtered case there is no server-rendered value to disagree with. The remaining difference is `toFixed()`'s tie-breaking on an exact `.5` within a filtered subset (e.g. 173/200 shows `87` where Python's `round()` would give `86`). I thought that was better left visible than papered over with a second rounding implementation, but I am happy to add one if you would rather the two agree exactly.

Also glad to add a changelog fragment if you want one — I could not see a `changes.d/` in the tree to follow.

### DIFF
--- a/coverage/htmlfiles/coverage_html.js
+++ b/coverage/htmlfiles/coverage_html.js
@@ -142,6 +142,7 @@ coverage.wire_up_filter = function () {
 
     const footer = table.tFoot.rows[0];
     const ratio_columns = Array.from(footer.cells).map(cell => Boolean(cell.dataset.ratio));
+    const footer_cells = Array.from(footer.cells).filter(cell => Boolean(cell.dataset.ratio));
 
     // Observe filter keyevents.
     const filter_handler = (event => {
@@ -149,6 +150,17 @@ coverage.wire_up_filter = function () {
         const totals = ratio_columns.map(
             is_ratio => is_ratio ? {"numer": 0, "denom": 0} : 0
         );
+
+        // What the server rendered, captured once. When nothing is filtered out we
+        // put these back verbatim instead of recomputing, so the footer cannot
+        // disagree with the <h1> above it (which is rendered from the same
+        // Python total and is never touched by JS).
+        footer_cells.forEach(cell => {
+            if (cell.dataset.serverText === undefined) {
+                cell.dataset.serverText = cell.textContent;
+                cell.dataset.serverRatio = cell.dataset.ratio || "";
+            }
+        });
 
         var text = document.getElementById("filter").value;
         // Store filter value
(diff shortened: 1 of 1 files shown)


nedbat:
I haven't looked at this yet, but I was wondering about testing it, so I'm glad you mentioned the JS tests. If you wouldn't mind, it would be amazing to get a PR that gets those working again.

nedbat:
See also #2259

PR #2259, opened 2026-08-14
html: clamp JS-recomputed footer percentage to mirror display_covered

### LINKED ISSUE #2256: HTML index: footer Total is recomputed in JS with toFixed(), can show 100% when coverage isn't 100%

MsfPablo:
Fixes #2256.

The HTML index's `<tfoot>` Total percentage is recomputed in
`coverage_html.js` with `toFixed()`, which can round to `0` or `100%`
even when actual coverage is not exactly 0 or 100. The `<h1>` heading
on the same page is server-rendered via `coverage.results.display_covered`,
which documents the invariant that "0" is only returned when the value
is truly zero, and "100" is only returned when the value is truly 100.
Rounding can never result in either "0" or "100".

Since the JS handler also runs on page load (it dispatches an `input`
event unconditionally to reapply stored filters; see the comment at
the bottom of the same function), the footer is overwritten on every
page view, so the bug shows up even without typing anything into the
filter.

This patch mirrors `display_covered`'s clamp in JS:

1. If `pc > 0` and `pc < 1/10**places`, nudge up to `1/10**places`.
2. Else if `pc < 100` and `pc > 100 - 1/10**places`, nudge down to
   `100 - 1/10**places`.
3. Else round to the configured precision.

This keeps the JS-recomputed footer from disagreeing with the
server-rendered heading or the text-mode report.

---

Disclosed: this contribution was prepared with the help of Claude
(a large language model). The human in the loop reviewed the diff and
verified the change matches the invariant documented in
`coverage/results.py::display_covered`.

### DIFF
--- a/coverage/htmlfiles/coverage_html.js
+++ b/coverage/htmlfiles/coverage_html.js
@@ -237,10 +237,24 @@ coverage.wire_up_filter = function () {
                 const places = match ? match[1].length : 0;
                 const { numer, denom } = totals[column];  // nosemgrep: eslint.detect-object-injection
                 cell.dataset.ratio = `${numer} ${denom}`;
-                // Check denom to prevent NaN if filtered files contain no statements
-                cell.textContent = denom
-                    ? `${(numer * 100 / denom).toFixed(places)}%`
-                    : `${(100).toFixed(places)}%`;
+                // Check denom to prevent NaN if filtered files contain no statements.
+                // Also mirror coverage.results.display_covered's clamp so the
+                // JS-recomputed footer can never round to "0" or "100" — keeping
+                // the heading and footer from disagreeing. See issue #2256.
+                let pc = denom
+                    ? (numer * 100 / denom)
+                    : 100;
+                if (places >= 0) {
+                    const near0 = 1 / Math.pow(10, places);
+                    if (pc > 0 && pc < near0) {
+                        pc = near0;
+                    } else if (pc < 100 && pc > (100 - near0)) {
+                        pc = 100 - near0;
+                    } else {
+                        pc = Math.round(pc * Math.pow(10, places)) / Math.pow(10, places);
+                    }
+                }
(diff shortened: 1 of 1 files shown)


nedbat:
#2257 is already working on this.

nedbat:
#2257 also claims to fix this, with simpler code. Can you compare?

DSeaStar:
Not my PR, but I built the harness for this while comparing my own #2273, so here is the measurement behind @nedbat's question.

I ran all three implementations against `coverage.results.display_covered` over numer 0..denom, denom 1..250, precision 0/1/2 (94,875 cases):

| implementation | mismatches vs Python |
| --- | --- |
| #2257 (clamp + `toFixed`) | 345 (149 / 108 / 88 at precision 0 / 1 / 2) |
| #2259 (clamp + `Math.round` + `toFixed`) | 345 (the same cases) |
| #2273 (clamp + half-even) | 0 |

Your patch and #2257 agree on the reported 0% / 100% symptom — both clamp. The 345 are all tie cases: Python ends in `round(pc, precision)`, which is half-even, while `toFixed()` and `Math.round()` both round ties away from zero. For example `1/8` → Python `12%`, both of ours `13%`.

The real difference between this PR and #2257 is elsewhere: #2257 caches the server-rendered footer text and restores it whenever no row is hidden, so on a plain page view it never recomputes — which also takes care of `--skip-covered`. The consequence is that its 345 can only appear while a filter is active, whereas here they can appear on any page view.

The full write-up, including the per-precision breakdown, is on #2273.

PR #2267, opened 2026-08-17
Use pathlib to open Python files

### LINKED ISSUE #2266: Coverage combine chokes on coverage databases from different platforms
**Describe the bug**
I want to combine coverage data from pytest runs performed on Linux and on Windows.
I execute them with the `pytest-cov` plugin and can report the coverage for each platform individually, but if I use `coverage combine` on Linux to combine the coverage databases to get the common code coverage, it reports:

```
No source for code: 'src\<module>\__init__.py'
```

If you look closely you can see that the path contains backslashes, which are from the Windows run and this path obviously doesn't exist on the Linux system, where the code coverage data should be combined and a report generated.

It seems coverage.py stores paths in the native way of the platform, which I confirmed by analyzing the database files:

`.coverage.windows` the table `file` contains:
```
src\<module>\__init__.py
```

`.coverage.linux` the table `file` contains:
```
src/<module>/__init__.py
```

**To Reproduce**

- Execute a code coverage run on Linux and Windows platforms from the root of a cross-platform Python module
- Try to merge the coverage using `coverage combine ...`

Answer the questions below:
  -  CPython 3.11.14
1. What version of coverage.py shows the problem? The output of `coverage debug sys` is helpful.
  - `coverage==7.15.2`
1. What code shows the problem?  Give us a *specific commit* of a *specific repo* that we can check out. If you've already worked around the problem, please provide a commit before that fix.
1. What commands should we run to reproduce the problem? *Be specific*. Include everything, even `git clone`, `pip install`, and so on. Explain like we're five!

**Expected behavior**

I expect `coverage combine` to be able to combine code coverage data from different platforms. It should be able to normalize the paths to the current platform, so that `coverage html` and `coverage xml` can work on the combined data.

**Additional context**

It might make sense to use `pathlib` to store coverage data in Posix format in the database and convert on Windows if needed.

creiter64:
This fixes https://github.com/coveragepy/coveragepy/issues/2266

Signed-off-by: Christoph Karch

### DIFF
--- a/coverage/python.py
+++ b/coverage/python.py
@@ -43,14 +43,20 @@ def read_python_source(path: Path) -> bytes:
 
 def get_python_source(filename: str) -> str:
     """Return the source code, as unicode."""
+
+    print(f"Trying to read file: {filename}")
     path = Path(filename)
 
     if not path.parent.exists():
         # Path not found try other encoding
         if env.WINDOWS:
+            print("On Windows but got Posix path")
             path = Path(PurePosixPath(filename))
         else:
+            print("On Posix but got Windows path")
             path = Path(PureWindowsPath(filename))
+
+    print(f"Converted to path: {path}")
 
     ext = path.suffix
     if ext == ".py" and env.WINDOWS:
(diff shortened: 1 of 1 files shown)


creiter64:
Closing in favor of the approach from @nedbat

PR #2268, opened 2026-08-17
Map only the matched prefix in PathAliases

ckarnell:
`PathAliases.map` substitutes with `str.replace`, so once a rule matches it rewrites every occurrence of the matched text, not just the part that matched:

```python
new = path.replace(m[0], result)
```

```python
aliases.add("/ci/src", "./mysrc")
aliases.map("/ci/src/vendor/ci/src/a.py", exists=lambda _: True)
# ./mysrc/vendor./mysrc/a.py
# want ./mysrc/vendor/ci/src/a.py
```

The second `/ci/src` is part of the file's real location, and it gets rewritten too. The separator normalisation on the next line then runs over the damage, which is where the `vendor./mysrc` comes from. Combining across machines is where this shows up, since the mapped path either fails the `exists` check and falls through to the next rule, or lands on the wrong file.

The regex is anchored with `match`, so slicing at the match bounds is the same thing the code already means.

Two open PRs touch this file, #2120 and #2233, and both change which rule matches, not the substitution. Different line, different failure. Happy to rebase behind either if you would rather land those first.

Tests are 425 passing across `test_files.py`, `test_api.py` and `test_data.py`. Reverting just the `files.py` line fails the new test in both its relative and absolute forms and nothing else, with the skip count at 7 either way.

### DIFF
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -506,7 +506,7 @@ class PathAliases:
 
         for original_pattern, regex, result in self.aliases:
             if m := regex.match(path):
-                new = path.replace(m[0], result)
+                new = path[:m.start()] + result + path[m.end():]
                 new = new.replace(sep(path), sep(result))
                 if not self.relative:
                     new = canonical_filename(new)
--- a/coverage/files.py
+++ b/coverage/files.py
@@ -506,7 +506,7 @@ class PathAliases:
 
         for original_pattern, regex, result in self.aliases:
             if m := regex.match(path):
-                new = path[:m.start()] + result + path[m.end():]
+                new = path.replace(m[0], result, 1)
                 new = new.replace(sep(path), sep(result))
                 if not self.relative:
                     new = canonical_filename(new)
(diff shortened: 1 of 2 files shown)


coverage/files.py
```diff
@@ -506,7 +506,7 @@ def map(self, path: str, exists: Callable[[str], bool] = source_exists) -> str:
 
         for original_pattern, regex, result in self.aliases:
             if m := regex.match(path):
-                new = path.replace(m[0], result)
+                new = path[:m.start()] + result + path[m.end():]
```
nedbat: This seems like a good fix, but this will be simpler:
```
new = path.replace(m[0], result, 1)
```

ckarnell:
The Windows failure was my test, not the change. In relative mode map() keeps the result's separator, and ./mysrc is forward slashed, so the output stays forward slashed on every platform. My expected value wrapped it in os_sep(), which flips to backslashes on Windows, so the [True] case failed there while [False] passed through canonical_filename first. I dropped the os_sep to match test_wildcard right above it. Pushed, CI should go green.

nedbat:

PR #2269, opened 2026-08-18
Exclude the `case` line when an irrefutable case body is excluded

### LINKED ISSUE #1563: Exclusion of body doesn't propagate to `case` block
**Describe the bug**

If the body of a `case` block inside a `match` statement is excluded from coverage, the `case` line is still marked as not covered.

**To Reproduce**

Code under test - `testcase.py`:
```py
if 1 != 1:
    assert False

match 1:
    case 1:
        print("all is well")
    case _:  # reported as uncovered
        assert False
```

Configuration - `.coveragerc`:
```ini
[report]
exclude_lines =
    assert False$
```

Commands:
```
$ coverage run --source=. -m testcase
all is well
$ coverage report -m
Name          Stmts   Miss  Cover   Missing
-------------------------------------------
testcase.py       5      1    80%   7
-------------------------------------------
TOTAL             5      1    80%
```

I'm using `coverage` version 7.2.0 with Python 3.10 on Ubuntu Linux 22.04.

**Expected behavior**

I would expect the exclusion to be propagated to the surrounding block for `case` blocks similarly to how it does for `if` blocks.

Note that when enabling branch coverage measurement, the `case 1:` line is reported as partially covered, but I expect that is a direct consequence of line 8 being considered uncovered, so not a separate issue.

yhuikzdtguioaert:
## The problem

Excluding the body of a `case` clause doesn't propagate to the `case` line, so
the `case` line is still reported as missing (or as a partial branch under
`--branch`). This is inconsistent with `if`/`else`, where excluding the `else:`
body makes the `else:` line disappear from the report.

Reproducer from the issue, with `exclude_lines = assert False$`:

```py
match 1:
    case 1:
        print("all is well")
    case _:            # still reported as uncovered
        assert False   # excluded
```

## Root cause

An irrefutable `case` (a wildcard `case _:`, or a bare capture like
`case other:`, with no guard) is the `match` equivalent of an `else` clause.
Unlike `else:`, though, the `case` line is a real statement with its own arc,
so excluding only its body left the `case` line behind as a missing statement /
partial branch.

## The fix

In `PythonParser._raw_parse`, while walking the AST, if a `case` is irrefutable
and its entire body is excluded, the `case` line is added to the excluded set
too. This mirrors the existing handling that carries exclusions from a
decorator/signature onto a function or class body.

The wildcard detection that `AstArcAnalyzer._handle__Match` already performed is
factored out into a small `_case_is_irrefutable()` helper that both places now
share.

Refutable cases such as `case 2:` are deliberately left alone: their `case`
line stays reported when the body is excluded, the same way an `elif` line does.

## Tests

- `tests/test_parser.py`: three unit tests covering an irrefutable `case _:`, a
  bare-capture `case other:`, and a case whose body is only *partly* excluded
  (which must not be excluded).
- `tests/test_coverage.py`: an end-to-end test in branch mode checking that the
  wildcard is excluded while a refutable `case 2:` with an excluded body is
  still reported as missing.

```
$ python -m pytest tests/test_parser.py tests/test_coverage.py tests/test_arcs.py
287 passed, 1 skipped
```

`ruff format --check`, `ruff check`, and `mypy --strict` are clean on the
changed files; `pylint` reports only a pre-existing unrelated TODO.

Fixes #1563.


nedbat:
This is now released as part of [coverage 7.16.1](https://pypi.org/project/coverage/7.16.1).

PR #2270, opened 2026-08-18
fix: check for multiprocessing in a --concurrency list

VXNCXNX:
`coverage run` refuses run-affecting command-line options alongside `--concurrency=multiprocessing`, since they cannot reach the subprocesses. The check compares the whole option value with `==`, so it never fires for a comma list, and the run proceeds with the parent measuring branch coverage and the subprocesses not:

```
$ coverage run --concurrency=multiprocessing,thread --branch prog.py
(exit 0, no message)
$ coverage combine
Can't combine statement coverage data with branch data

after:
$ coverage run --concurrency=multiprocessing,thread --branch prog.py
Options affecting multiprocessing must only be specified in a configuration file.
Remove --branch from the command line.
(exit 1)
```

The comma form is not exotic, `doc/migrating.rst` tells people to write it: "Now that must be explicitly enabled with ``--concurrency=multiprocessing,thread``". The single value form was refused correctly the whole time.

The fix splits on commas, which is what `cmdline.py` already does with the same string a few hundred lines up when it builds the `Coverage` object, so the two now agree on what the option means.

I checked that `--concurrency=thread --branch` still exits 0, that `multiprocessing,thread` with no extra flags still exits 0, and that `thread,multiprocessing --branch` is refused as well.

AI-assisted: written with Claude Code, reviewed and tested by me before sending.


nedbat:
Thanks again. Would you like me to credit you in the Contributors list?

nedbat:

Nitwizztor (approved):

PR #2272, opened 2026-08-26
test: restore HTML JavaScript test suite

fly1d:
In [#2257](https://github.com/coveragepy/coveragepy/pull/2257), Ned asked for a separate PR to get coverage.py's JavaScript tests working again. The suite was stopping before it could run because `tests/js/index.html` still loaded local jQuery and template files that have been removed from the repository.

This PR:

- updates the fixture to match the current generated `pyfile` markup (`p > span.n > a`);
- replaces the jQuery/template helpers with native DOM APIs;
- removes the stale local asset references; and
- fixes the existing `to_prev_chunk()` boundary case where the previous highlighted chunk starts on line 1. Restoring the existing tests exposed that early return; the normal selection behavior is unchanged.

This is intentionally separate from the percentage calculation work in [#2257](https://github.com/coveragepy/coveragepy/pull/2257) and [#2259](https://github.com/coveragepy/coveragepy/pull/2259).

### Verification

- Browser QUnit suite: 96 tests, 253 assertions, 0 failures, with no console errors or warnings.
- `python -m pytest -n 0 tests/test_html.py -q` (67 passed).
- `node --check tests/js/tests.js`.
- `node --check coverage/htmlfiles/coverage_html.js`.
- `pre-commit run --all-files`.
- `git diff --check`.

OpenAI Codex assisted with the implementation and verification of this change.

### DIFF
--- a/coverage/htmlfiles/coverage_html.js
+++ b/coverage/htmlfiles/coverage_html.js
@@ -503,20 +503,23 @@ coverage.to_prev_chunk = function () {
         chunk_indicator = c.chunk_indicator(probe_line);
     }
 
+    // There is no previous highlighted chunk.
+    if (!chunk_indicator) {
+        return;
+    }
+
     // There's a prev chunk, `probe` points to its last line.
     var end = probe+1;
 
     // Find the beginning of this chunk.
-    var prev_indicator = chunk_indicator;
-    while (prev_indicator === chunk_indicator) {
-        probe--;
-        if (probe <= 0) {
-            return;
+    while (probe > 1) {
+        probe_line = c.line_elt(probe-1);
+        if (c.chunk_indicator(probe_line) !== chunk_indicator) {
+            break;
         }
-        probe_line = c.line_elt(probe);
-        prev_indicator = c.chunk_indicator(probe_line);
+        probe--;
     }
-    c.set_sel(probe+1, end);
+    c.set_sel(probe, end);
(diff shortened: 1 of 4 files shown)


yangfan-yf-yf (approved):
I rechecked d1740510. The fixture matches the current `p > span.n > a` markup, the native DOM helpers preserve the existing assertions, and the previous-chunk scan correctly retains a chunk beginning on line 1.

In a local Chromium run, the QUnit page completed 96 tests / 253 assertions with no failures or console warnings. On Windows with Python 3.12, `tests/test_html.py` completed with 66 passed and one platform-specific skip. Both JavaScript syntax checks and `git diff --check` passed.

The repository checks currently do not execute `tests/js`; adding a headless entry point would be a useful follow-up, as would removing the now-unreferenced `tests/qunit/jquery.tmpl.min.js`. Neither affects the requested suite restoration or boundary fix.

nedbat:
Thanks so much! Can we also update the contributing.rst file with directions for running the tests?  I opened `tests/js/index.html`, and saw 96 passing tests, but also 1 through 6 at the top. Is that how you expect it to look?

image

nedbat:
Hmm, they are the 6 r/b lines:
image

nedbat:
I've added a clean-up to the fixture, and updated the docs. Thanks for doing the hard part!

PR #2281, opened 2026-09-19
Fix loop completion arcs on Python 3.14

### LINKED ISSUE #2168: 3.14: uncovered branch for loop completion with break

HanpuLi:
Fixes #2168.

On Python 3.14, both settrace-based cores can receive a `return` event after the frame's current line has advanced without a preceding line event for that final source line. They were therefore recording the function-exit arc from the previous line, which can make normal loop exhaustion immediately before return appear as a missing branch.

For real returns on Python 3.14+, ctrace and pytrace now record that final line transition first when needed, then record the function-exit arc. Sysmon already reports the expected loop-completion arc.

The regression covers empty, exhausted, and broken-loop paths. On Python 3.14.7 the exact reproducer reports 100% branch coverage with both ctrace and pytrace after the fix; the complete `tests/test_arcs.py` suite also passes under both cores. A broader return/loop/try-finally probe produced identical raw arc sets for ctrace and pytrace.

PyTracer's trace callback cannot itself be measured by meta-coverage, so the new callback branch uses the repository's `pragma: not covered` marker. A direct PyTracer return-frame test exercises the same transition outside the trace callback, while the end-to-end arc regression verifies the real tracing behavior.


HanpuLi:
I missed #2204 when checking for existing work on #2168. It already addresses the same CTracer/PyTracer return-arc issue, so this PR is redundant. Closing mine to avoid competing duplicate work. Sorry for the noise.

nedbat:
Thanks for the work, and for seeing the duplication, but I prefer this change. Re-opening, and making a few tweaks.

